PR #19, opened 2015-06-29
Setting ANDROID_HOME

yaronyg:


obastemur:
Shouldn't that needs to be `~/.bash_profile` ?

yaronyg:
Sigh... I forgot the magic OS/X behavior for the terminal app. :( Will go fix. Thanks for catching this!

PR #20, opened 2015-06-29
OS/X wants .bash_profile not .bashrc

yaronyg:
Terminal on OS/X always does a full login for some reason so you want to use .base_profile because .bashrc will never be run. Thanks to @obastemur for pointing this out!


obastemur:
Thanks

PR #26, opened 2015-07-14
Story0 juksilve

DrJukka:


yaronyg:
Per our call today please delete all the sample directory files.

src/android/java/io/jxcore/node/BtConnectorHelper.java
```diff
@@ -57,20 +56,35 @@ public BtConnectorHelper() {
...
 
+    public boolean isRunning(){
+        if(mBTConnector != null){
+            return true;
+        }else{
+            return false;
+        }
+    }
     public void Stop(){
         if(mBTConnector != null){
```
yaronyg: Can we guarantee that this call will only be called from a single thread? Otherwise the call to stop on line 66 will result in a null exception. We should probably change this to store mBTConnector in a local variable before calling stop. This assumes, of course, that BTConnector itself is thread safe.

src/android/java/io/jxcore/node/BtConnectorHelper.java
```diff
@@ -57,20 +56,35 @@ public BtConnectorHelper() {
...
             mBTConnector.Stop();
             mBTConnector = null;
         }
 
-        if (mBtToServerSocket != null) {
-            print_debug("Disconnect:::Stop : mBtToServerSocket");
-            mBtToServerSocket.Stop();
-            mBtToServerSocket = null;
+        if (mServerSocketList != null) {
+            for(int i = 0; i < mServerSocketList.size(); i++) {
```
yaronyg: List isn't safe for concurrent access so if two threads both call stop then bad things are going to happen.

src/android/java/io/jxcore/node/BtConnectorHelper.java
```diff
@@ -57,20 +56,35 @@ public BtConnectorHelper() {
...
     }
 
+    public boolean isRunning(){
+        if(mBTConnector != null){
+            return true;
+        }else{
+            return false;
+        }
+    }
     public void Stop(){
```
yaronyg: Are we sure that the calls to stop below can never throw? I'm asking because if there is any problem then we will have an uncaught exception that will leave us in an undefined state and necessitate crashing the app that is using us. That's fine if we are in some horrible unrecoverable mess but otherwise we should do our best to keep running.

src/android/java/io/jxcore/node/BtConnectorHelper.java
```diff
@@ -100,6 +113,28 @@ public boolean Disconnect(String peerId){
 
         return ret;
     }
+
+    //test time implementation to simulate 'peer disappearing'
+    public boolean  DisconnectIncomingConnections(){
```
yaronyg: A big part of this function's purpose was testing how the native layer handles connections just dying. So I was expecting us to do something really nasty like jump down to the bluetooth server socket object and call close on it and see how the rest of the native code handles that. The kill below is way too clean to really test anything at the native layer.

src/android/java/io/jxcore/node/BtConnectorHelper.java
```diff
@@ -213,58 +237,85 @@ public void Connected(BluetoothSocket bluetoothSocket, boolean incoming,String p
...
+                }
 
-                    if(mBtToRequestSocket != null){
-                        mBtToRequestSocket.Stop();
-                        mBtToRequestSocket = null;
+                mBtToRequestSocket = new BtToRequestSocket(bluetoothSocket, new BtSocketDisconnectedCallBack() {
+                    //Called when disconnect event happens, so we can stop & clean everything now.
+                    @Override
+                    public void Disconnected(Thread who, String Error) {
+                        if (mBtToRequestSocket != null) {
```
yaronyg: Usual thread safety issues on mBtToRequestSocket.

yaronyg:
@DrJukka Please see my comments but otherwise I'm +1

PR #38, opened 2015-07-20
No T!

yaronyg:
We weren't actually grabbing the tape object in the function call back! OOOPs!


yaronyg:
EEEk! This is old code that should really be gone.

PR #40, opened 2015-07-21
Story 0 sree

sreesharp:
Adding cordova hook to setup the pre-requisites for thali_corodva plugin


yaronyg:
This PR is committing to the wrong branch. It should be to story_0 not master.

PR #163, opened 2015-10-14
Don't exit the Cordova process in a hook.

### LINKED ISSUE #143: Cordova Run doesn't work on Android
We need to figure out why cordova run won't work with us on Android

vjrantal:
The Android hook had an async step after which the process was exited. This
caused the Cordova process to exit too soon and was seen, for example, when
`cordova run android` was tried to be executed.

The `appRoot` variable is taken from the Cordova context instead of from
`__dirname` so that the path gets right even if the plugin is added via a
symbolic link.

Also removed unneeded dependency to the `lie` module.

Fixes #143.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/163)

### DIFF
--- a/scripts/androidAfterPrepare.js
+++ b/scripts/androidAfterPrepare.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var Promise = require('lie');
 var fs = require('fs-extra-promise');
 var path = require('path');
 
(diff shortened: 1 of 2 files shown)


yaronyg:
Reviewed 2 of 2 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/163)

PR #191, opened 2015-10-19
Make it easy to review testServer.md

yaronyg:
I'm submitting this PR to make it easy for @DrJukka and @vjrantal to provide in context feedback to my proposal for handling counting devices both on the desktop and in continuous integration.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/191)

### DIFF
--- a/doc/api/testServer.md
+++ b/doc/api/testServer.md
@@ -63,7 +63,6 @@ TestServer -> Device3 : send 'too late'
 
 
 ![Example of counting devices](http://g.gravizo.com/g?
- {
   @startuml
   participant TestServer
   participant Device1
@@ -77,7 +76,7 @@ TestServer -> Device3 : send 'too late'
   TestServer -> Device2 : send 'start tests' + { "android": 1, "ios": 1 }
   Device3 -> TestServer : send 'present' + { "os": "iOS" }
   TestServer -> Device3 : send 'too late'
-  }
+  @enduml
 )
 ### Test Server Logic
 
--- a/doc/api/testServer.md
+++ b/doc/api/testServer.md
@@ -69,7 +69,7 @@ TestServer -> Device3 : send 'too late'
   participant Device2;
   participant Device3;
   TestServer -> TestServer : Start timer;
-  Device1 -> TestServer : send 'present' +  "os": "android" ;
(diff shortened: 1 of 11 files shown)


vjrantal:
Overall, looks good.

One thing that was talked before and I think could be added to the specification is the "pinging mechanism" to avoid present devices to shut down the connection prematurely. Currently, it seems like Android devices start disconnecting somewhere like ~130 seconds from the beginning, which can be seen from lines like:
```
Android  DEV3497 added : 81.26 sec., device count 18
Android DEV2008 id now disconnected 131.425 sec.
Android DEV6002 id now disconnected 131.654 sec.
Android DEV3159 id now disconnected 133.76 sec.
```
(above from https://gist.github.com/ThaliTester/a6459b103f8ef1960b1c)

The "pinging mechanism" would be essentially a process of sending a ping-message during the time period of when a device states it is present to until the server starts the tests.

yaronyg:
@vjrantal There is actually a timeout setting for socket.io that is passed into the constructor. If you set it to 0 then the connection should never timeout. But this is only a short term hack. Once we start running tests that involve turning off the local radio stack (think chaos monkey) then we either have to re-write the code so that the test server (and the device) stop assuming that the socket.io connection will stay up forever or we need to switch to vanilla HTTP which doesn't care about dropped connections.

DrJukka:
@yaronyg Actually the value is defined as "timeout connection timeout before a connect_error
and connect_timeout events are emitted", and indeed value 0 makes the connections to fail strait away.

@vjrantal appears that there is 'native pinging' implemented with the API, and where the server should get ping reply within 60 seconds, and client default value is 25 seconds.
Investigating currently depracated 'heartbeat interval' & 'heartbeat timeout' which according to docs are still implemented there, and thus might be a way on changing the default values.

Also will check what 'close timeout' would do, supposedly it might delay disconnects, and if there is reconnect before the timeout expires, it should not declare disconnect..

DrJukka:
BTW, never got to check 'close timeout'. I set really big values to  'heartbeat interval' & 'heartbeat timeout' and checked with 3 android devices that they were happy waiting for a full hour without any disconnections. 

Currently waiting CI to be back online in order to verify that it works on there as well.

Issue #192, opened 2015-10-20
Test setup script sometimes fails on OS X in thali postinstall script


vjrantal:
It might be just on my machine, but creating this as a reminder to figure out the root cause and as reference in case someone else sees this as well.

The error is coming when running:
```
./setUpTests.sh UnitTest_app.js 192.168.1.150
```
The error details below:
```
node.js:813
throw ee;
      ^
Error: getaddrinfo ENOTFOUND
    at errnoException (dns.js:17:7)
    at Object.onanswer [as oncomplete] (dns.js:102:10)
npm ERR! thali@2.0.1 postinstall: `jx installCordovaPlugin.js`
npm ERR! Exit status 8
npm ERR! 
npm ERR! Failed at the thali@2.0.1 postinstall script.
npm ERR! This is most likely a problem with the thali package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     jx installCordovaPlugin.js
npm ERR! You can get their info via:
npm ERR!     npm owner ls thali
npm ERR! There is likely additional logging output above.
```

vjrantal:
@obastemur: To debug this, I'd like to see which address is not found. My issue was that I wasn't sure which `node.js` file is referenced here (to set breakpoints or console.logs). In the bash script, we are prefixing all commands with `jx`, i.e., not calling `node` or `npm` directly.

vjrantal:
After a quick investigation, seems like it is coming when we try to reach `https://github.com/jxcore/jxcore-cordova-release/raw/master/0.0.7/io.jxcore.node.jx`. Since I am sure there is nothing wrong with that server, perhaps this is some local DNS issue or something like that.

@obastemur: When Googling `Error: getaddrinfo ENOTFOUND` seems like people have had this kind of issues with some older versions of `node`. What was the version the latest `jxcore` is based on?

vjrantal:
@yaronyg and @tobybrad: I noticed you discussed in Slack about "bad download of io.jxcore.node.jx". Do you think the issue you see is what I reported here?

vjrantal:
Continuing the monologue by saying that I also now saw this error which is what you might have been talking about:
```
Downloaded io.jxcore.node.jx
Failed with - Error: Command failed: 
module.js:1917
throw err;
      ^
Error: Package is either corrupted or not compatible with this JX binary.
```
[...]

PR #245, opened 2015-10-27
Use OS temporary folder as the download cache.

vjrantal:
A part of the installation is the download of the required version of the
JXCore Cordova plugin (currently ~90MB download) and unpacking that with `jx`
before adding it as Cordova plugin using Cordova commands.

This commit implements caching of the unpacked plugin folder so that the
overall installation process becomes faster. The OS temporary folder is used
for caching instead of any Thali-app folder so that full benefits of the cache
are received even if the app-folder is re-generated (this is the case with CI
scripts, for example).

This commit also adds some catches to the promise-chains so that in case of
errors, it is easier to see where the issue was.

Relates to #192.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/245)

### DIFF
--- a/thali/install/install.js
+++ b/thali/install/install.js
@@ -1,6 +1,7 @@
 'use strict';
 var exec = require('child_process').exec;
 var path = require('path');
+var os = require('os');
 var https = require('https');
 var unzip = require('unzip');
 var Promise = require('lie');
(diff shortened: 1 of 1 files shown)


vjrantal:
Closing this PR in favor of https://github.com/thaliproject/Thali_CordovaPlugin/pull/246 because the CI doesn't kick in for PRs outside of this repository.

Issue #164, opened 2015-10-14
Test if we have connection issues with iOS


yaronyg:
There is a possibility that if two iOS devices connect simultaneously to each other over both advertising and listening then there is a potential bug that will cause one of the connections to fail after around 2 minutes. We have to write a perf test to validate if this issue exists.

yaronyg:
Note that we have never seen this.

tobybrad:
This should naturally be covered by a long running perf test I believe.

yaronyg:
Have you looked at the perf tests and convinced yourself that any of them will do the job?

DrJukka:
We can add the test once we get the multiple out-going connections working. Basically then I can start connection out, and connect back to any connection that comes in and wait. Thus I would either have just two devices communication as a pair, or two pairs doing the same. Is there any other requirements, for example, do we need to exchange data while waiting ?

yaronyg:
We will need to exchange data. So the set up would be: Device A connects to Device B, Device B then reciprocates by connecting to Device A and then they start moving as much data as humanly possible for as long as possible.

PR #255, opened 2015-10-28
Ios p2p fixes

tobybrad:
Simplify state management, especially around reasons for disconnect happening, seems to do the trick.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/255)

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -70,7 +70,7 @@
           compiler-flags="-Werror"/>
         <source-file src="src/ios/MultipeerConnectivity/THEMultipeerSession.m" 
           compiler-flags="-Werror"/>
-        <header-file src="src/ios/MultipeerConnectivity/THEMultipeerSessionDelegate.h" 
+        <header-file src="src/ios/MultipeerConnectivity/THEMultipeerDiscoveryDelegate.h" 
           compiler-flags="-Werror"/>
         <header-file src="src/ios/MultipeerConnectivity/THEMultipeerClient.h" 
           compiler-flags="-Werror"/>
--- a/src/ios/MultipeerConnectivity/THEMultipeerClient.h
+++ b/src/ios/MultipeerConnectivity/THEMultipeerClient.h
@@ -37,7 +37,7 @@
 - (id)initWithPeerId:(MCPeerID *)peerId
   withPeerIdentifier:(NSString *)peerIdentifier
      withServiceType:(NSString *)serviceType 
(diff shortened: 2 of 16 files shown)


yaronyg:
I was expecting this to include a fix for #164 but I don't see it. That isn't a show stopper but I just wanted to make sure that this was supposed to fix #164.


---

Reviewed 15 of 15 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/ios/MultipeerConnectivity/THEMultipeerClientSession.m, line 132 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/255#-K1jG8f5FUfY6myivZjL)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/ded03d997e2a0fa1f5e617d803a0d85c0655a91b/src/ios/MultipeerConnectivity/THEMultipeerClientSession.m#L132)):</sup>
NIT - I realize you have the assert above to check a bad state but if the state machine changes or what have you wouldn't it be more robust to help us find errors if there was a final else here that threw? In fact have you considered just using an explicit state machine library? I did a search for "objective-c state machine" and found at least three on GitHub. Using a state machine first approach you would hang logic off the state machine transitions so 'impossible' situations are, well, impossible.


---

<sup>**[src/ios/MultipeerConnectivity/THEMultipeerPeerSession.m, line 92 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/255#-K1jRLKQYsPEPyCs6bCj)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/ded03d997e2a0fa1f5e617d803a0d85c0655a91b/src/ios/MultipeerConnectivity/THEMultipeerPeerSession.m#L92)):</sup>
One suspects this shouldn't just be an assert but should be live code. If something does go wrong we want to make sure we capture it no matter what, not only if we are in debug.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/255)

yaronyg:
I'm waiting for responses.

yaronyg:
Toby will check if this is superseded by another of his checkins and if so he will close.

tobybrad:
Indeed these commits are all in  PR 272. Closing.

PR #251, opened 2015-10-27
Use OS temporary folder as the download cache.

vjrantal:
A part of the installation is the download of the required version of the
JXCore Cordova plugin (currently ~90MB download) and unpacking that with jx
before adding it as Cordova plugin using Cordova commands.

This commit implements caching of the unpacked plugin folder so that the
overall installation process becomes faster. The OS temporary folder is used
for caching instead of any Thali-app folder so that full benefits of the cache
are received even if the app-folder is re-generated (this is the case with CI
scripts, for example).

This commit also adds some catches to the promise-chains so that in case of
errors, it is easier to see where the issue was.

Relates to #192.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/251)

PR #259, opened 2015-10-28
Proposal for thaliMobile.js

yaronyg:
This is the proposal for the replacement for our emitter layer. This would be the formal definition of the Mobile APIs that Thali directly exposes.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/259)


ThaliTester:
PR is added to the queue for testing as 4. task. (34376db)

ThaliTester:
Test (Fail) 49083571 build is completed (34376db)

See https://github.com/ThaliTester/TestResults/tree/4908357134376db_Proposal_for_thaliMobile_js_yaronyg/ for the logs

yaronyg:
If you look at the [build link](https://github.com/ThaliTester/TestResults/blob/4908357134376db_Proposal_for_thaliMobile_js_yaronyg/build.md) you will find our old friend:

```
Failed to process the downloaded io.jxcore.node.jx file
```

Will your checkin fix this or do we need to send this to Oguz?

vjrantal:
> Will your checkin fix this or do we need to send this to Oguz?

Yes, #251 will fix this.

DrJukka:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 76 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1msecC_oN1R9O5K2kO)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L76)):</sup>
Turning radios off will of course turn advertisement off. This will then generate event. 
Currently when Radios are turned back on the advertising etc. were turned back on, I suppose with this spec we don't do this, but  we just generate the event for the change.


---

<sup>**[thali/thaliMobile.js, line 183 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1my0T3IUkQapNkkBVX)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L183)):</sup>
Do we just disconnect all incoming connections, or do we also disconnect all out-going connections when this function is called ?


---

<sup>**[thali/thaliMobile.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1n4tA-ptX0Pk7FXyBf)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L227)):</sup>
Do we generate event for the closing of the non-TCP connection ? 
we did not originally have event defined, but as the current version of the replication manager was not detecting the issue, we added an event there.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

vjrantal:
Review status: 0 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 20 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1n8VmCZMXOffMkw3hW)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L20)):</sup>
If I understood above open issue correctly, my gut feeling is that "illegal" calls should be prevented already at the Node.js layer. This way, it is easier to implement the error behavior similarly cross-platform. Like stated in http://www.catb.org/esr/writings/taoup/html/ch01s06.html "When you must fail, fail noisily and as soon as possible" :)


---

<sup>**[thali/thaliMobile.js, line 104 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1nBS16ykf6CoNJO-IV)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L104)):</sup>
In one way, I guess one could say that this function has three separate functions. Two defined on these lines and the third being signaling an update to the advertisement.

Would it make at all sense to take the update part away and make it a separate function? Then, you would have, for example:

`StartAdvertisingAndListenForIncomingConnections`
and
`UpdateAdvertising`

In this spec (below), it specified that the `StartAdvertisingAndListenForIncomingConnections` function should not be called with different port, but if the update part would be separate, you wouldn't have that issue since the `UpdateAdvertising` would not take a port as argument.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

yaronyg:
Review status: 0 of 8 files reviewed at latest revision, 5 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 20 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1n8VmCZMXOffMkw3hW)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L20)):</sup>
I would tend to agree.


---

<sup>**[thali/thaliMobile.js, line 76 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1msecC_oN1R9O5K2kO)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L76)):</sup>
Yup. Per our conversation today if the radios go off then an event fires and everything is disabled. It's up to the app to call back with start to recover, presumably something it would do when it gets another event telling it that the radios are back on (something, in the case of Android, it might have done itself programmatically)


---

<sup>**[thali/thaliMobile.js, line 104 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1nBS16ykf6CoNJO-IV)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L104)):</sup>
As per the call I agree with the basic idea that we should separate these. I didn't do that because I want to keep the native layer as simple as possible and few people will ever call this API. But we can also add in updateAdvertising as a node.js API that would then call StartAdvertisingAndListenForIncomingConnections with the same portNumber.


---

<sup>**[thali/thaliMobile.js, line 183 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1my0T3IUkQapNkkBVX)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L183)):</sup>
I will clarify this in the spec, but it's just incoming connections. Now the question is - if we stop listening for other node's advertisements should we stop outgoing connections? I don't believe this is required on iOS and it certainly isn't a requirement on Android. So I would tend to say no.


---

<sup>**[thali/thaliMobile.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1n4tA-ptX0Pk7FXyBf)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L227)):</sup>
Per the call our plan is not to fire an event then. Rather the fact that the TCP/IP connection dies is what tells the app that there is a problem.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

ThaliTester:
PR is added to the queue for testing as 1. task. (8cd7fff)

ThaliTester:
Test (Success) 49083571 build is completed (8cd7fff)

See https://github.com/ThaliTester/TestResults/tree/490835718cd7fff_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 490835718cd7fff(8cd7fff) has failed

See https://github.com/ThaliTester/TestResults/tree/490835718cd7fff_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (9ba48d2)

ThaliTester:
Test (Success) 49083571 build is completed (9ba48d2)

See https://github.com/ThaliTester/TestResults/tree/490835719ba48d2_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 490835719ba48d2(9ba48d2) has failed

See https://github.com/ThaliTester/TestResults/tree/490835719ba48d2_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (c1c7a3a)

ThaliTester:
Test (Success) 49083571 build is completed (c1c7a3a)

See https://github.com/ThaliTester/TestResults/tree/49083571c1c7a3a_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 49083571c1c7a3a(c1c7a3a) has failed

See https://github.com/ThaliTester/TestResults/tree/49083571c1c7a3a_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

vjrantal:
Review status: 0 of 11 files reviewed at latest revision, 7 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 16 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K26K-Lu05fp89eWK5tc)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobile.js#L16)):</sup>
I was myself thinking that would it be possible for the layers higher up to deal with the app server rather than passing the responsibility onto the thaliWifiInfrastructure layer? IP and BSSID changes could be subscribed to in the layer above the thaliMobileNative and thaliWifiInfrastructure layers.

On mobile platforms, the BSSID changes are coming from the thaliMobileNative layer and I believe on desktop platforms, we would need to write some library (unless we find one) that uses some OS-specific APIs to determine when BSSID changes.

In my opinion, managing the BSSID changes and the express app server would fit better into the layer above thaliWifiInfrastructure, because otherwise, there would need to be a dependency from thaliWifiInfrastructure to thaliMobileNative, which would be a bit more like a "peer-dependency" (in terms of the layers there exists). Also, it feels "unnatural" that the code in thaliWifiInfrastructure would have to know about the express web application framework, which is a part of a more higher level "app logic".


---

<sup>**[thali/thaliWifiInfrastructure.js, line 17 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K26Y7x5Iqhy96RVDbDj)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliWifiInfrastructure.js#L17)):</sup>
Does above mean that a queuing mechanism should be implemented inside of the ThaliWifiInfrastructure implementation? For example, let's imagine that I'd have an implementation like:
```
ThaliWifiInfrastructure.prototype.startListeningForAdvertisements = function () {
  return new Promise(function(resolve, reject) {
    this._server.start('0.0.0.0', function () {
      resolve();
    });
  });
};
```
Above would not be according to the specification, but a queue should be implemented so that subsequent calls would be queued until previous calls resolved?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

yaronyg:
Review status: 0 of 11 files reviewed at latest revision, 7 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 16 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K26K-Lu05fp89eWK5tc)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobile.js#L16)):</sup>
Let me split this into two issues. One is about BSSID and the other is about the app server.

### BSSID
On the issue of BSSID there is currently no defined way in node.js to get BSSID information. So I had left that to be resolved by thaliMobileNative. But I admit that is a cop out because that will only work on Android and iOS. If we want it to work on the desktop then the only solution I know of is to make a system call out to the native platform. But that isn't as easy as it sounds since it typically requires system calls outs. 

I think the right resolution is that we should create a stand alone library that is responsible for publishing radio changes. When we are on Android or iOS we will get the info from the native layer and when we are on desktop we will have to make OS call outs. This information will be published from that library and most normal users will grab it as an event on thaliMobile.js. This means that networkChangedCallback will be taken out of thaliMobileNative and moved to its own file. Instead there will be a "mobileStateChange" event that will fire anytime advertising or listening is turned on or off. BTW, without writing native code I don't know anyway to find out about BSSID changes on desktops other than polling with something like https://github.com/stevenmiller888/scanner/blob/master/lib/index.js.

I opened https://github.com/thaliproject/Thali_CordovaPlugin/issues/277 to track this refactor.

Thanks for the point btw! I was just assuming that thaliMobileNative would handle this and obviously that isn't workable on the desktop.

### App server
In regards to the app server the situation is a bit more complex.

In theory someone can create an Express Route (e.g. an app) and just let it sit there no available to anyone.

In practice, if only so that things can be managed from the local webview, the developer is likely to host the app server on a 127.0.0.1 address and port. By definition this address cannot be routed to from off device.

So to make the wifi functionality available we have to create a HTTP server that hosts the user's APP which has an externally visible address. And to stop wifi functionality we have to be able to turn off that server.

So the simplest solution to this problem was just to have the user pass in their app and then we can create a server with an external address and turn it off when necessary.

There are other solutions that wouldn't require passing in the app but I think they bring more complexity than they are worth.

* We could try to use iptables to route an external address and port to the user's 127.0.0.1 address. What's nice about this is that all the user would have to tell us is their local port so this would mirror the thaliMobileNativeWrapper API. What's not so nice about this approach is that at least in the case of Android it isn't clear if this is possible without a rooted device and it appears to require C level code. And as far as I can tell iOS just plain doesn't support it.
* Another option is a proxy. We could create effectively a proxy that listens on an external address and when it receives an incoming connection it just automatically opens an outgoing connection to the user's localhost server port. The code would be pretty trivial (very similar to what we do today in the mux layer) and it would let the API mirror that of the thaliMobileNativeWrapper. BUT (you knew there was a but coming, right?) it introduces a whole new layer for things to go wrong. It also adds an extra hop for every single request. I also am not sure if this won't screw up anything in regards to TCP/IP headers and such because. It's basically just something else to go wrong that we are introducing not because we have no choice (we had no choice, for example, with the mux layer) but because it makes the API prettier.

So my gut was to just make the API a bit uglier by requiring the user to pass in the app object so that we could control the external access directly and not have to introduce any intervening logic.

What do you think?


---

<sup>**[thali/thaliWifiInfrastructure.js, line 17 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K26Y7x5Iqhy96RVDbDj)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliWifiInfrastructure.js#L17)):</sup>
That is exactly correct. The above wouldn't be right. The easiest solution is probably what we did [here](https://github.com/thaliproject/Thali_CordovaPlugin/blob/story_001_yarong/thali/identityExchange/identityexchangeendpoint.js#L42) where we use a promise to create a single global queue that ensures that commands get run serially.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

vjrantal:
Review status: 0 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 16 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K26K-Lu05fp89eWK5tc)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobile.js#L16)):</sup>
BSSID: Agreed.

App server: I'll sync-up with you offline since it might be that I wasn't clear enough what I wanted to point out.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

DrJukka:
Review status: 0 of 11 files reviewed at latest revision, 6 unresolved discussions.

---

<sup>**[thali/thaliMobileNative.js, line 128 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2HJg9fVPhsmBAsZpsw)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobileNative.js#L128)):</sup>
What are Clientport & Serverport. I assume these are used for outgoing connection, and currently as far as I can see there is only one port used. That is originally null, thus it gets assigned when we start doing local connection binding.


---

<sup>**[thali/thaliMobileNative.js, line 212 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2HIw8lG-rM6z2Q9cG8)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobileNative.js#L212)):</sup>
Does this Kill all outgoing & incoming connections, or something else ?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

yaronyg:
Review status: 0 of 11 files reviewed at latest revision, 6 unresolved discussions.

---

<sup>**[thali/thaliMobileNative.js, line 128 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2HJg9fVPhsmBAsZpsw)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobileNative.js#L128)):</sup>
Talked to Jukka. He now understands the answer to this question. He didn't before because he just read the paragraph above but not the paragraph below.


---

<sup>**[thali/thaliMobileNative.js, line 212 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2HIw8lG-rM6z2Q9cG8)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliMobileNative.js#L212)):</sup>
If you read the text that is linked to above (not that the link works here, you will have to manually go to the thaliMobileNativeWrapper.killConnections definition) you will see that this only applies to iOS and kills everything in both directions.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

tobybrad:
Reviewed 6 of 8 files at r1, 1 of 3 files at r2, 1 of 2 files at r3, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

<sup>**[thali/thaliMobile.js, line 183 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1my0T3IUkQapNkkBVX)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L183)):</sup>
There's a huge difference between disconnecting existing connections and not allowing new (incoming ones). I'm not clear which way you're proposing. Practically.. on iOS if you're not listening to advertisements you wont be able to make outgoing connections.


---

<sup>**[thali/thaliWifiInfrastructure.js, line 124 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2ql-42dnaWB4I0vo4a)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliWifiInfrastructure.js#L124)):</sup>
I find this confusing when taken alongside the similar API in thaliNativeLayer. In an iOS world MPCF will perform this exact same function when connected to an AP. I understand this is probably intended to be the separate x-platform SSDP-based mechanism but I'm concerned this is just sowing the seeds of confusion for application developers. Perhaps just a name change is required to make it clear this is a separate, standalone discovery method ??


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

yaronyg:

---

<sup>**[thali/thaliMobile.js, line 183 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1my0T3IUkQapNkkBVX)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L183)):</sup>
This file is unfortunately not quite done yet (yes, I should have closed the review, this PR has been a bit of a mess). But the general idea is that if you aren't listening for advertisements then you CANNOT make outgoing connections on either Android or iOS.


---

<sup>**[thali/thaliWifiInfrastructure.js, line 124 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2ql-42dnaWB4I0vo4a)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliWifiInfrastructure.js#L124)):</sup>
In the case of iOS we would always use MPCF but if the peer happens to be on WiFi then we would also use SSDP. The reason for this is that depending on how hardcore MPCF is it should be able to run simultaneously on Apple's flavor of wi-fi direct and thus find iOS phones that aren't on the same AP while also using the Wi-Fi flavor of MPCF that can find iOS phones that are on the AP as well as using SSDP which can find non-iOS devices.

There is an obvious problem here that we can and will find the same iOS phone in this scenario at least twice. Once over MPCF and again over SSDP if they are on the same WiFi AP. I do have an open issue to resolve that but I'm not sure how much we really care at this point. It's a perf improvement to resolve this since identity exchange WILL detect this situation, in other words, we will know that we have found the same peer twice across two different mechanisms. So it's strictly a perf issue, not a correctness issue.

With that as background your statement is exactly right, thaliWiFiInfrastructure is meant as a parallel discovery mechanism to thaliMobileNativeWrapper. They are theoretically independent of each other. If you think the names are confusing I'm open to alternative names.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

ThaliTester:
PR is added to the queue for testing as 2. task. (8da7ee9)

ThaliTester:
Test (Success) 49083571 build is completed (8da7ee9)

See https://github.com/ThaliTester/TestResults/tree/490835718da7ee9_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 490835718da7ee9(8da7ee9) has failed

See https://github.com/ThaliTester/TestResults/tree/490835718da7ee9_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

tompaana:
Reviewed 1 of 9 files at r5.
Review status: 8 of 12 files reviewed at latest revision, 7 unresolved discussions.

---

<sup>**[thali/NextGeneration/thaliMobileNativeWrapper.js, line 291 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2wMPI9Kw3g9yXDx9v5)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/8da7ee95e3a4916f04690c158f837e0634b86688/thali/NextGeneration/thaliMobileNativeWrapper.js#L291)):</sup>
Since I can't find actually issues to bitch about, I have to bring this up: TCP vs. Tcp. In proper camel case, I'd prefer this way i.e. "Tcp", but majority of instances is written all upper case.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

ThaliTester:
PR is added to the queue for testing as 1. task. (2943f5a)

ThaliTester:
Test (Success) 49083571 build is completed (2943f5a)

See https://github.com/ThaliTester/TestResults/tree/490835712943f5a_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (5f33d22)

ThaliTester:
Test (Success) 49083571 build is completed (5f33d22)

See https://github.com/ThaliTester/TestResults/tree/490835715f33d22_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 490835712943f5a(2943f5a) has failed

See https://github.com/ThaliTester/TestResults/tree/490835712943f5a_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

ThaliTester:
Test 490835715f33d22(5f33d22) has failed

See https://github.com/ThaliTester/TestResults/tree/490835715f33d22_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (c9a8fea)

ThaliTester:
PR is added to the queue for testing as 2. task. (4574cc1)

yaronyg:
Review status: 6 of 14 files reviewed at latest revision, 7 unresolved discussions.

---

<sup>**[thali/NextGeneration/thaliMobileNativeWrapper.js, line 291 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2wMPI9Kw3g9yXDx9v5)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/8da7ee95e3a4916f04690c158f837e0634b86688/thali/NextGeneration/thaliMobileNativeWrapper.js#L291)):</sup>
So TCP is an acronym and acronyms are properly given in all upper case so I have changed my use of TCP to be consistently upper case everywhere. Sorry but in this case I think English grammar has to win. :)


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

ThaliTester:
Test (Success) 49083571 build is completed (c9a8fea)

See https://github.com/ThaliTester/TestResults/tree/49083571c9a8fea_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test (Success) 49083571 build is completed (4574cc1)

See https://github.com/ThaliTester/TestResults/tree/490835714574cc1_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 49083571c9a8fea(c9a8fea) has failed

See https://github.com/ThaliTester/TestResults/tree/49083571c9a8fea_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

ThaliTester:
Test 490835714574cc1(4574cc1) has failed

See https://github.com/ThaliTester/TestResults/tree/490835714574cc1_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

vjrantal:
Review status: 6 of 14 files reviewed at latest revision, 12 unresolved discussions.

---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 24 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K3-wmKmiP6NdwAvxebU)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L24)):</sup>
What does `cause us to do nothing` mean in this context? I'd assume there is at least some logic within `start` related to the passed `router` object.


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 31 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K3-xQXlCeDcLMmVaNNW)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L31)):</sup>
Do you meant that `ThaliWifiInfrastructure` should be implemented as a singleton and you would not use it with the `new` keyword like you currently use the ThaliEmitter:
```
var emitter1 = new ThaliEmitter();
var emitter2 = new ThaliEmitter();
```


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 100 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K30-LxQLTHUOQFN-cmV)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L100)):</sup>
Call `start`or `startUpdateAdvertisingAndListenForIncomingConnections`?


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 130 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K30-frvYpa_yeQhYCey)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L130)):</sup>
Is it `start`or `startUpdateAdvertisingAndListenForIncomingConnections` under which the HTTP server is started?


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 169 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K301E5vni-XiWd2KWrI)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L169)):</sup>
"we MUST fire this event ourselves whenever we call"  <- this wasn't perfectly clear to me. Does "we" in this context refer to a case where a call is made within this module?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

ThaliTester:
PR is added to the queue for testing as 1. task. (a5e1382)

yaronyg:
Review status: 6 of 14 files reviewed at latest revision, 12 unresolved discussions.

---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 24 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K3-wmKmiP6NdwAvxebU)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L24)):</sup>
That is exactly right. I've updated the text to reflect that.


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 31 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K3-xQXlCeDcLMmVaNNW)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L31)):</sup>
That is correct. That is the intent of the text and is the original design based on mirroring the thaliMobileNative code which has to be a singleton because we kept the native code simple by letting it assume that there is just one single global owner in node.js. In other words, the native code doesn't have to worry about different programs calling start or stop advertising at different points and figuring out how to do the right thing.

But last night I realized (and then forgot so THANK YOU for this issue!!!) that this isn't the right approach for WiFi because we want to have multiple WiFi instances for testing purposes and that WiFi, unlike thaliMobileNative, can easily run multiple independent instances without running into itself. So long as SSDP uses different ports for each instance (something I'll point out in the text) and keeps using port 0 then we can run lots of parallel wifi instances. Which is great for testing.

So I have updated the spec and code to make this into a class rather than a singleton.


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 100 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K30-LxQLTHUOQFN-cmV)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L100)):</sup>
Yup, I clarified this in the text.


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 130 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K30-frvYpa_yeQhYCey)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L130)):</sup>
startUpdateAdvertisingAndListenForIncomingConnections. I have clarified the text.


---

<sup>**[thali/NextGeneration/thaliWifiInfrastructure.js, line 169 \[r9\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K301E5vni-XiWd2KWrI)** ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4574cc153c8335eb7a900a95288bc9ac349ff53f/thali/NextGeneration/thaliWifiInfrastructure.js#L169)):</sup>
Clarified in the text


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)

ThaliTester:
Test (Success) 49083571 build is completed (a5e1382)

See https://github.com/ThaliTester/TestResults/tree/49083571a5e1382_Proposal_for_thaliMobile_js_yaronyg/ for the logs

ThaliTester:
Test 49083571a5e1382(a5e1382) has failed

See https://github.com/ThaliTester/TestResults/tree/49083571a5e1382_Proposal_for_thaliMobile_js_yaronyg/ for the fail logs

yaronyg:
Review status: 6 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 291 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2wMPI9Kw3g9yXDx9v5:-K4UUiaZQgneVhRF7uCj:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/8da7ee95e3a4916f04690c158f837e0634b86688/thali/NextGeneration/thaliMobileNativeWrapper.js#L291)):*
Done.

---

*[thali/thaliMobile.js, line 183 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K1my0T3IUkQapNkkBVX:-K4UUjhlsJP_UdNcA8Yn:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/34376db27dc60a31d34e514b681ef9cf9ed949bd/thali/thaliMobile.js#L183)):*
Done.

---

*[thali/thaliWifiInfrastructure.js, line 124 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259#-K2ql-42dnaWB4I0vo4a:-K4UUkeOCu2IqyutpgWx:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c1c7a3a2c39b2323c1cb5477413062a37f80e43c/thali/thaliWifiInfrastructure.js#L124)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/259)*

PR #458, opened 2016-01-15
Changes that @tobybrad and I agreed to

yaronyg:
Specified that if connect is called a second time on the same peer ID and the first call either has resulted in a connection or hasn't completed yet then we will return an error. That way we don't have to manage a queue for connection calls at the native layer.

Fixed the link to the binding spec.

Specified that when we get too many peer Availability Changed updates that overloads our storage queue between sending batch updates to
node.js that we should drop the oldest announcements first in order to keep the queue at or below it's maximum size. Also clarified that for
incomingConnectionToPortNumberFailed only one notification in regard to any particular port's failure needs to be sent in each 100 ms window.

Specified that when the first subscription is received to networkChanged that we have to fire a networkChanged event with the current status. That
will let the thaliMobileNativeWrapper initialize it's method that tracks this information.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/458)

### DIFF
--- a/thali/NextGeneration/thaliMobileNative.js
+++ b/thali/NextGeneration/thaliMobileNative.js
@@ -164,7 +164,7 @@
  * other words we are pushing an iOS bug up from the iOS native code into
  * Node.js and requiring us to solve it up in Node.js land. For all the gory
  * details on how this works see the [binding
- * spec](http://www.thaliproject.org/presenceprotocolBindings.md).
+ * spec](http://thaliproject.org/PresenceProtocolBindings/).
  *
  * We use the `clientPort` value below so that the mux layer can figure out
  * which of its connections is the one it needs to use to talk to the desired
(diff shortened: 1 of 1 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (c98eeae)

ThaliTester:
Test (Success) 56204092 build is completed (c98eeae)

See https://github.com/ThaliTester/TestResults/tree/56204092c98eeae_Changes_that__tobybrad_and_I_agreed_to_yaronyg/ for the logs

ThaliTester:
Test 56204092c98eeae(c98eeae) has failed

See https://github.com/ThaliTester/TestResults/tree/56204092c98eeae_Changes_that__tobybrad_and_I_agreed_to_yaronyg/ for the fail logs

Issue #317, opened 2015-11-12
Create spec for replication layer


yaronyg:
This spec will consume information from the notification layer about what peers are available and have data for the current peer and will then initiate a replication with that peer based on a prioritized list and limits of the platform.

This layer will also specify how we hook into PouchDB databases to find out when there have been changes and then communicates with the notification layer to advertise that we have data for folks.

Finally this layer will also specify how we figure out when someone has actually replicated with us so we can remove them from our notification list.

PR #466, opened 2016-01-19
First bits of the Wifi infrastructure work

vjrantal:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/466)


ThaliTester:
PR is added to the queue for testing as 1. task. (7226f84)

ThaliTester:
Test (Success) 56449660 build is completed (7226f84)

See https://github.com/ThaliTester/TestResults/tree/564496607226f84_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 564496607226f84(7226f84) has failed

See https://github.com/ThaliTester/TestResults/tree/564496607226f84_Wifi_infrastructure_work_vjrantal/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (31ce140)

vjrantal:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 53 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OguQaqIjiFpEvEWbh:-K8OguQaqIjiFpEvEWbi:435389902) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L53)):*
I wasn't 100% sure which one was the right name for this function and hoping my change made it correct.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 35 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OWXWhNGA7VKNwaswP:-K8OWXWi0eIn9r2_0nO6:678255326) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/ca30f267d243f788e63b1a924f8f777f51cf65de/thali/NextGeneration/thaliWifiInfrastructure.js#L35)):*
This one I have confirmed in the desktop testing scenarios.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 46 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OXLM6ALNnSb48hXpO:-K8OXLM7cbFu6pHRojf4:-196111815) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7226f8437597bf36b59e0e6bb2c8b513805f655b/thali/NextGeneration/thaliWifiInfrastructure.js#L46)):*
From the documentation perspective, how should this kind of arguments be tackled?

They aren't supposed to be a part of the public API, but something optional that is used in the automated tests.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 251 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OhNn6TrewKzPssEcQ:-K8OhNn6TrewKzPssEcR:1002865219) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L251)):*
I think this chapter in the specification might need attention.

The way I implemented this was that in SSDP messages, we advertise the port on which we run the express router object that was passed to us and not the port used in multicast messaging.

In fact, the node-ssdp library at the moment uses a fixed port for the multicast traffic and if I remember right, I was not able to see messages from peers running on same machine if I picked up a random port for each one of them. Things worked correctly when all peers were using the "standard" port 1900 https://github.com/diversario/node-ssdp/blob/33eac5e0561d638fe781ef94e077faace016fef7/lib/index.js#L80

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 312 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Oh63LoCnj1GOu9LF-:-K8Oh63LoCnj1GOu9LF0:-1326356507) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L312)):*
Was this the right way to start the passed router?

In normal scenarios, self.port is 0, which means OS will pick a random port that is available.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
Test (Success) 56449660 build is completed (31ce140)

See https://github.com/ThaliTester/TestResults/tree/5644966031ce140_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 5644966031ce140(31ce140) has failed

See https://github.com/ThaliTester/TestResults/tree/5644966031ce140_Wifi_infrastructure_work_vjrantal/ for the fail logs

yaronyg:
There are some missing features including:

* discoveryAdvertisingStateUpdateWifiEvent support
* Turning discovery on and off as appropriate as WiFi goes away (we'll have to mock the native layer events for now)
* Advertising frequency isn't changed when we go into the background on Android (we could use onPause/onResume for this from JXcore but we need to check what those bind to in Android)
* It doesn't appear to have the throttling to make sure that we only respond to discovery requests at a fixed rate.
* We need to throttle the incoming UDP traffic so if there is too much we stop listening for some fixed period of time and then try listening again to see if the problem has gone away.

I would appreciate it if you could review [here](https://github.com/thaliproject/thali/blob/gh-pages/pages/documentation/PresenceProtocolBindings.md#local-wi-fi-binding) and [here](https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext/thali/NextGeneration/thaliWifiInfrastructure.js) to see if there is anything else missing.

---

Reviewed 1 of 4 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 34 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 1 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qza3HWaNcsnzkw8MM:-K8Qza3HWaNcsnzkw8MN:258497904) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L1)):*
I have to admit that I was expecting a bit more testing. I would thought that we would have tests to:

* Make sure that we are actually checking that the wifi radio is available and turned on. When we are on mobiles we would get this via the networkChanged event. This applies to both startListeningForAdvertisements as well as startUpdateAdvertisingAndListening. We can use a native mock layer for now.
* Make sure we respond to the wifi radio events properly, e.g. if state changes that we handle this either by re-advertising (as required on BSSID changes) or turning off advertising and listening when wifi goes away. Again, this can all be mocked for now.
* Test all the idempotent methods to make sure they really are idempotent (you do this for some methods but not all)
* Make sure that stop methods actually stop, we can cheat by examining the state of server objects if possible but live tests are better.
* Tests for all the events we are supposed to fire (once we implement them of course)
* Tests that when we stop advertising we actually issue a byebye
* Make sure when we change our USN that we both issue a byebye for the old USN and an alive for the new one.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 22 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qz7zqs16HPoNlyXB_:-K8Qz7zqs16HPoNlyXBa:-2009056394) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L22)):*
How do we test if the infrastructure has stopped cleanly? Note that eventually you'll need to use something like [this](https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext_yarong_317/thali/NextGeneration/makeIntoCloseAllServer.js) but when I check in #317 you will get the details on that.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 33 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R0VLjCpLSLBCeu9Ca:-K8R0VLkYADmgV9BQ_UY:-2073956675) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L33)):*
Might want to use a shorter ad interval just so the test ends sooner.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 43 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R0cpZ6MXxTLWzZ9pD:-K8R0cpZ6MXxTLWzZ9pE:798308925) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L43)):*
Shouldn't we also test that stop works? In other words that once we call stopListeningForADvertisements that we can fire an advertisement and we don't get a notification?

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 50 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R1WOaZ3h-LXxguzvj:-K8R1I7R-Jdt8FXTrDsm:-265832052) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L50)):*
This is a really fragile test. Any change to the code could break it. Anyway we can check this the old fashioned way by setting up a node-ssdp listener? That will also verify that what we send 'on the wire' is right.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 53 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OguQaqIjiFpEvEWbh:-K8QxZpJ799WqJFizNxN:-646346827) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L53)):*
vNext is the official branch so whatever is there is correct. That is what you have above.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 77 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R1pPpsXAQhVSJdb4G:-K8R1pPpsXAQhVSJdb4H:1504185033) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L77)):*
This is a good test but shouldn't it be augmented with a "live" test where we get node-ssdp to send an irrelevant message and see that we successfully ignore it? End to end testing is good isn't it?

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 101 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R3PmGQShVjzv1OwrY:-K8R3PmHig4JUpKiPqbT:110856070) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L101)):*
I probably would have snuck in a live test here where you get the location from the location header of the SSDP advertisement. This proves that everything works end to end. 

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 46 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OXLM6ALNnSb48hXpO:-K8QD2zd0lh_p2D1m3Dg:-545389255) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7226f8437597bf36b59e0e6bb2c8b513805f655b/thali/NextGeneration/thaliWifiInfrastructure.js#L46)):*
[Insert grumble about real languages supporting overloading]. We want to be super careful about our public APIs. So I would probably suggest removing this from the constructor and instead sticking it in as a private property setting. E.g. since it's testing it's fine to have a "setDeviceNameForTest(deviceName)" function that is marked `@private` and lets you force a new name. Another, probably safer, approach would be to introduce a new 'startTest(deviceName)' method that would be defined on this object as `@private` and it would set the deviceName and call the 'real' start. But if at all possible we should get test values off public APIs.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 6 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qsju9tsJNFq0oxVAS:-K8QrxjwVejg3WgljTMr:-1639717149) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L6)):*
This library uses some odd language that makes a mess of things. It has an option in its configuration called "UDN". But this is what SSDP actually calls USN. If you looks [here](https://github.com/diversario/node-ssdp/blob/master/lib/server.js#L29) you will see that they initialize _usns using the udn value. Then if you look at [addUSN}(https://github.com/diversario/node-ssdp/blob/master/lib/index.js#L370) you will see that they take whatever value you submit and slap it onto the end of the UDN. 

Now that might make sense in some limited UPnP scenarios but it isn't the right behavior for SSDP in a generic sense. The right behavior is that the USN is set by what node-ssdp calls the UDN (which, in fact, node-ssdp does [here](https://github.com/diversario/node-ssdp/blob/master/lib/server.js#L111) but the NT value (set [here](https://github.com/diversario/node-ssdp/blob/master/lib/server.js#L109)) should just be the flat NT value that we submitted. Not the prefixed value they have created. Ideally we would fork and fix this but otherwise we can just add an object to fix it and file a bug.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 10 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QF5p7f-l_bcwlb4LU:-K8QF5p7f-l_bcwlb4LV:-1611473167) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L10)):*
I'm confused, how can this be a constant? The USN is the Unique Service Name and as specified [here](https://github.com/thaliproject/thali/blob/gh-pages/pages/documentation/PresenceProtocolBindings.md#ssdpalive) is supposed to be a UUID that is changed every time the beacons change.  

The constant should be the service type URI which is recorded in the NT. So this value should be for the NT and as given in the spec it's supposed to be `http://www.thaliproject.org/ssdp`.

Now none of this is holy writ. If it's wrong, that's cool, but the place to handle that is by submitting a PR to the spec and then catching the code up. Otherwise we won't ever be on the same page.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 48 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QDpMis0mjJ2zuyLDA:-K8QDpMjWr0Gx_Kienhv:-1636643269) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L48)):*
Since the USN is a constant why have a local variable for it at all? Why not just always refer to the constant?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 49 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QE-AG4HnQJRypEnt5:-K8QE-AG4HnQJRypEnt6:891355677) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L49)):*
This is actually supposed to be a UUID URL. Something like [node-uuid](https://github.com/broofa/node-uuid) generates them automatically. I don't have any experience with that library so I don't know how good it is. I'd also touch base with @mpodwysocki to see what library he is using for URL safe UUIDs.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 68 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QHiK1CXJOgpriSWxK:-K8QHiK1CXJOgpriSWxL:-1623408176) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L68)):*
Out of curiosity, have you tried to test the SSDP functionality with two different devices? I'm curious if the default 0.0.0.0 unicastHost address gives an Internet route-able address for the server or not.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 69 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QJkaNq0EXtHtijw75:-K8QJkaNq0EXtHtijw76:1370879074) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L69)):*
I notice that there is no option in node-ssdp to advertise a cache-control header. That's fine and certainly not worth fixing but it would be nice to have a bug about this in ice box so we can eventually fix it.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 73 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QISE1F39RNHkWiq-b:-K8QISE1F39RNHkWiq-c:-455838883) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L73)):*
I think using deviceName is misleading here because this value really isn't the deviceName. Instead it is just the hard coded string "http://www.thaliproject.org/ssdp"

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 93 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QKvju1t3WrCguVXYR:-K8QKvjvLbGYyYcNRUdV:1162380625) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L93)):*
This is a style issue but generally having methods whose only purpose is to set a value on an internal variable can lead to confusion. Also the SSDP constructor already supports setting this value as an option with location. So why not just add location to serverOptions and set it to the output of this function which would be the URL?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 104 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QMke3KYxThn7KCh7x:-K8QMke3KYxThn7KCh7y:-1849668480) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L104)):*
This is not the format specified for wifiPeerAvailabilityChanged, please see [here](https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext/thali/NextGeneration/thaliWifiInfrastructure.js#L226).

Passing in the whole location URL is problematic because it could contain a path and then what do we do? We are only supposed to be communicating the IP and host. So I think we should parse the LOCATION header.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 119 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QLP8jQB4jDXWkq-Io:-K8QLP8jQB4jDXWkq-Ip:1217870647) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L119)):*
The USN should be some random UUID, shouldn't we be filtering on the NT? Isn't it the NT that tells us if this is a Thali device?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 121 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QLeB9xjJ5lyWWBU5h:-K8QLeB9xjJ5lyWWBU5i:2057993288) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L121)):*
The above sentence is not grammatically correct but I would just delete it and edit the sentence below to read "filtering out messages from ourselves"

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 153 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QOqHsQJcORjbNgM-h:-K8QOqHsQJcORjbNgM-i:109797603) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L153)):*
Why bother with self when you don't have any contexts to worry about that could get confused? Same with stop and other places where this comment is relevant.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 155 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QO5hGbTNkn3MmWZ41:-K8QO5hH57xJRx3v09lW:1463980096) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L155)):*
When the spec says "Error" we explicitly mean the Error object so this should technically be 

```javascript
return Promise.reject(new Error('Call Stop!'));
```

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 205 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QPoXLMDxMTI-tSCCF:-K8QPoXLMDxMTI-tSCCG:1260559803) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L205)):*
In this case you really do need self. But I believe it's generally considered good practice that once you have defined self you use it everywhere. E.g. it's considered bad style to mix this and self in the same function.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 208 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QQ2wagw_Vnl_80O44:-K8QQ2wb8w-_yillI1jR:1745735143) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L208)):*
Imagine the following situation:
Time 0 - We get our first call to this method and return Promise P1
Time 1 - P1 hasn't resolved but we get a second call to this method and immediately return a resolved promise (because we mark listening as true immediately without waiting for P1 to resolve)
Time 2 - P1 is rejected for some reason

So the second caller thinks that we have started listening for advertisements when, in fact, we have not.

The way to avoid this problem is that we need to record P1 and return it to everyone who calls until a call to stop is made and its value is cleared. That way everyone is in the same state.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 210 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QRiLhET0UnOsAS0Ms:-K8QRiLhET0UnOsAS0Mt:69549228) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L210)):*
I dug around the node-ssdp code and the CB works in a seriously screwed up way. Essentially the CB is called when the binding occurs. But if there is a problem that causes the binding to not occur then there is no callback and this promise will hang forever.

This is because of what I can only think of as a bug in node-ssdp. If you look [here](https://github.com/diversario/node-ssdp/blob/master/lib/index.js#L159) you will see that it is just /dev/nulling (effectively) any errors it receives. It's even worse, if you check out _createSocket it will just blindly accept the result.

Now I have to admit that the node documentation is a bit vague on exactly how bind behaves if an illegal address is provided. But I believe it will throw an exception. Shouldn't we catch that exception so we can resolve this promise as reject? 

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 235 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QUD8FptCF1egERHxo:-K8QUD8FptCF1egERHxp:-1187147612) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L235)):*
Same race condition with the same fix as above.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 251 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OhNn6TrewKzPssEcQ:-K8QYQui-HiBoK4brBkx:-1630696795) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L251)):*
You are right. My text is a mess. I got confused and somehow decided that the response address had to be in the host head which makes no sense since the host header is the address you are writing to. In any case, I just sent you a [PR](https://github.com/thaliproject/Thali_CordovaPlugin/pull/470) with new proposed text.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 302 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrO4TtzaaS0956KB8:-K8QrO4ULpjLwGpHz5YW:1225714326) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L302)):*
It would probably be a good idea to test if this is actually a router object (as opposed to just something that isn't null). I think there is a way to do this with Express.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 304 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrSYaKb-_VXeZC9Dd:-K8QrSYaKb-_VXeZC9De:-468644588) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L304)):*
Please see previous comment about generating UUID URLs.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 305 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QuhZhE4q2mwXPsK50:-K8QuhZhE4q2mwXPsK51:-1638213421) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L305)):*
Yup, that's a show stopper. But as I discuss at the start of this file, node-ssdp's addUSN function is broken in other dimensions as well.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 306 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QreP9QxHA36zmFT08:-K8QrXLUZI5zYlKdLEoc:-36021758) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L306)):*
Again, this should be a UUID URL.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 311 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QvZVitz-0iAthvXN_:-K8QvZVitz-0iAthvXNa:1301373492) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L311)):*
 If someone calls this method multiple times we'll try to start up a different server each time! But if you use the same design I previously discussed where we store the promise and return it until we get a stop then this problem resolves itself nicely.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 312 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Oh63LoCnj1GOu9LF-:-K8QuyM2uZQ6Sq9HlJL0:1615180916) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L312)):*
No. Because we need to use the router in other places. I would have expected something like:

```javascript
var express = require('express');
var app = express();
app.use('/', router);
app.on('error', function(err) ...);
app.listen(self.port, function() {
   do more stuff;
});
```

 

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 341 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QxzvyhLRaZi_e4uCk:-K8QxzvyhLRaZi_e4uCl:1245712750) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L341)):*
Yup, same comment as all the other ones about the race condition. :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-:-K8QyNtY78G5ozQglNUA:1639663135)*

yaronyg:
Review status: all files reviewed at latest revision, 34 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 1 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qza3HWaNcsnzkw8MM:-K8R9OJQhdD7hGdKne9u:246598214) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L1)):*
* We should test two instances on the same machine of the infrastructure successfully working with each other. We rely on this in so many places it's good to have a clean test for it so if it has failed we will immediately know why all the other desktop tests have failed.
* Need to test that all functions that can do start/stop/start actually work properly.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

vjrantal:
There might have been a misunderstanding since I opened the PR on purpose for the work I have done thus far and not because I would be claiming all aspects of the specification are completed :)

There are some aspects where mine and @tobybrad work interact (like some native API tests and the Wifi based mock for them) so I'd like to merge my work little by little to vNext where these come together. Otherwise, we will be working isolated in our personal branches for a long time and will end up with big PRs.

---

Review status: all files reviewed at latest revision, 23 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 1 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qza3HWaNcsnzkw8MM:-K8SgXZbmVztSUSRItHR:1946769846) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L1)):*
You can expect more testing, but that is work to be done and not competed yet.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 22 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qz7zqs16HPoNlyXB_:-K8SgwK3LKIVXvHnkTH6:1006341379) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L22)):*
A test case will be added.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 33 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R0VLjCpLSLBCeu9Ca:-K8Sh2lS-UtEG_q3T0WR:-299077157) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L33)):*
Actually, I could remove the interval since it doesn't matter in a success case. This is because the first advertisement is sent right after starting and only for the next one, the interval matters.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 43 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R0cpZ6MXxTLWzZ9pD:-K8ShPxrIlUXehIRa-ds:677251276) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L43)):*
Yes.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 50 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R1WOaZ3h-LXxguzvj:-K8Sinl7W5vlh5KjZuSG:23341877) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L50)):*
There was reasons why I did it this way and am planning to change it to a "live test" with node-ssdp once the feature works correctly.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 53 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OguQaqIjiFpEvEWbh:-K8SkO8Ll4tTKgryDIrL:-1044378725) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L53)):*
I wanted to double-check since there can always be mistakes. In another branch, I noticed the older name was used https://github.com/thaliproject/Thali_CordovaPlugin/blob/6db8d2b7f4842a0ddd91d61a36bb234b3e6dd234/test/www/jxcore/bv_tests/testThaliMobileNative.js#L64 so I wanted to get it clarified.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 77 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R1pPpsXAQhVSJdb4G:-K8SkyKSz7kmxU-l76ei:707983173) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L77)):*
I'd say this feature could be augmented with an e2e test.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 101 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8R3PmGQShVjzv1OwrY:-K8SlsOg_sJdHkVLhSiG:36799958) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L101)):*
I prefer keeping this test related to the router object and have a separate test to verify location in SSDP advertisement is correct.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 6 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qsju9tsJNFq0oxVAS:-K8SmEjkiOW_HSlaI6im:939910188) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L6)):*
I noticed this as well and got a bit confused by the terms and variable names used. I am not planning to fix the library from this regard at the moment, but just use the confusing names for the time being. This is to get faster to a working vNext implementation.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 10 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QF5p7f-l_bcwlb4LU:-K8SmhzCQVsV3iLmK3Zp:-1172117469) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L10)):*
The constant is a prefix and the full value is changed on every beacon change as specified. I'll change the name to reflect this.

I didn't yet use the NT, because node-ssdp had a bit weird terms here as well https://github.com/diversario/node-ssdp/blob/33eac5e0561d638fe781ef94e077faace016fef7/lib/server.js#L109 . Will try to start using it as specified.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 48 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QDpMis0mjJ2zuyLDA:-K8SqUcg5-2z6SL1EIEM:-204157600) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L48)):*
This related to how I wrote one of the tests, but can be improved.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 49 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QE-AG4HnQJRypEnt5:-K8Sq_bSEQAxDDcaZm8T:1754033451) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L49)):*
Will start using node-uuid.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 68 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QHiK1CXJOgpriSWxK:-K8Su7aGbjmstXLJZWxP:-710641736) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L68)):*
I have tried between two physical laptops (one Windows and one Mac) and the peers can discover each other other while I have them connected to the same Wifi AP.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 69 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QJkaNq0EXtHtijw75:-K8SrB_s6O79w64gzT80:655115937) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L69)):*
Created https://github.com/thaliproject/Thali_CordovaPlugin/issues/471

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 73 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QISE1F39RNHkWiq-b:-K8SutBSMvmPOfydgeSa:987149780) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L73)):*
See above where I mentioned I am not yet using the NT value as specified.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 93 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QKvju1t3WrCguVXYR:-K8Sv9EXPNCVTvJoPkBe:188842032) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L93)):*
I am currently constructing the SSDP objects only once, but the location needs to be set many times since I am also using this function from the mock of the older Thali protocol.

I don't think there is too much confusion to have a function like this.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 104 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QMke3KYxThn7KCh7x:-K8SvwMPPTuQoUB15zUO:1330724466) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L104)):*
I was actually going to propose a change in the specification related to this. Will add that change to the PR. I think splitting the location to IP and host is not required anywhere.

I thought it is intended that the location has a path which is /NotificationBeacons in the specification at https://github.com/thaliproject/thali/blob/gh-pages/pages/documentation/PresenceProtocolBindings.md#transferring-discovery-beacon-values-over-http .

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 119 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QLP8jQB4jDXWkq-Io:-K8SxAaL1ePCvte8XOQV:-1582418045) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L119)):*
See above.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 121 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QLeB9xjJ5lyWWBU5h:-K8SxLFT8Pv9jmPNsNwY:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L121)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 208 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QQ2wagw_Vnl_80O44:-K8Sxnm_y5NHAGSrJqkk:618948691) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L208)):*
Yep - this is one of the not-yet-implemented features of the specification.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 210 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QRiLhET0UnOsAS0Ms:-K8Sy03AJEXCPux6Tp4z:1208296593) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L210)):*
Yeah - I had mentioned this earlier that the node-ssdp library had essentially done synchronous APIs on top of asynchronous APIs. Part of my thali-changes was to add these callbacks to the relevant places.

I am 100% sure there are still things broken in the library, but my approach has been that I don't try to fix all of them now, but rather try to implement the Thali-logic on top and only fix the issue blocking me.

If we see too much issues with the library, we need to have separate work items to fix them or change the library we use.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 251 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OhNn6TrewKzPssEcQ:-K8Syauk4aGtqWPerWZN:-1258464983) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L251)):*
PR merged.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 302 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrO4TtzaaS0956KB8:-K8SyhghmQG8sdOChZXq:1163860493) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L302)):*
I didn't want to go too deep with all the tests, because I wanted to proceed with the basic implementation. Things can always be improved later.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 305 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QuhZhE4q2mwXPsK50:-K8SysuF0NsAWUo-icdU:-725612925) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L305)):*
Yep - it is.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 312 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Oh63LoCnj1GOu9LF-:-K8Sz9KwlU0hoMPsF5pZ:2098356615) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L312)):*
Should we unbound the error handler in `do more stuff;`, because there, we will resolve or reject the promise. Or, must we keep that error handler in place and emit some errors if later on, something goes wrong with the app?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-:-K8Sf3x_0QNn3NOSbv9W:1665183205)*

ThaliTester:
PR is added to the queue for testing as 1. task. (6be5677)

ThaliTester:
Test (Success) 56449660 build is completed (6be5677)

See https://github.com/ThaliTester/TestResults/tree/564496606be5677_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 564496606be5677(6be5677) has failed

See https://github.com/ThaliTester/TestResults/tree/564496606be5677_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the fail logs

yaronyg:
Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 53 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OguQaqIjiFpEvEWbh:-K8V-5TCmRv_OmvCUkTx:-1715936853) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L53)):*
Please file a bug with Toby. BTW, I believe that Toby didn't realize that some of the vNext specs had been updated.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 6 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qsju9tsJNFq0oxVAS:-K8V0IrV880uj2iofTNb:1498924664) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L6)):*
Please just file a bug that we have to figure out how to work with node-ssdp since they don't seem to acknowledge our issues or PRs. Stick it in backlog, don't mark it with "new infra" milestone and we'll deal with it later.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 104 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QMke3KYxThn7KCh7x:-K8V3rJEWmkBXauVURJm:36321917) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L104)):*
Per our conversation we agreed to stick with the existing spec and split this into host and port. Just remember that the URL we were sent could be evil and designed to make us produce garbage so please validate that the host and port it produces are reasonable.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 210 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QRiLhET0UnOsAS0Ms:-K8V8HHAMDd1_x8KeL8A:1321284429) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L210)):*
Let's file a bug in icebox that we need to make sure that if two apps on the same iOS or Android phones are trying to listen to the same UDP port, is it shared? What if they are in the same instance of JXcore? We should test this.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 302 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrO4TtzaaS0956KB8:-K8V8aj1VRZmaqAxIFAc:474993570) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L302)):*
Please add to the checklist

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 312 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Oh63LoCnj1GOu9LF-:-K8V9aJJKdYq9cVBgWfR:698908882) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L312)):*
Per our conversation please use winston to log the errors but we won't put an error emitter on this object right now because we don't have a good example of what the error would be. When we see it then we can change this.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 118 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8V47MmjjAvcI_he-UO:-K8V47MmjjAvcI_he-UP:1367318206) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6be56771e22e6cfb719f88f65bef05e180f6312b/thali/NextGeneration/thaliWifiInfrastructure.js#L118)):*
This need to us the Winston logging framework. Please see https://github.com/thaliproject/Thali_CordovaPlugin/blob/989ec1b5cce9557f823afb87f463754eae939e4e/thali/thalilogger.js

Please move this into vNext

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
PR is added to the queue for testing as 1. task. (311ec66)

ThaliTester:
Test (Success) 56449660 build is completed (311ec66)

See https://github.com/ThaliTester/TestResults/tree/56449660311ec66_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (0f5d794)

vjrantal:
Review status: 3 of 7 files reviewed at latest revision, 7 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 53 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8OguQaqIjiFpEvEWbh:-K8Xyfzpyw6cjTjlvH4c:1647179058) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L53)):*
Done.

https://github.com/thaliproject/Thali_CordovaPlugin/issues/478

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 6 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Qsju9tsJNFq0oxVAS:-K8XykQ6C2XtzaG-4qCL:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L6)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 104 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QMke3KYxThn7KCh7x:-K8Y3LVhXrM4-lbYjm1_:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L104)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 210 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QRiLhET0UnOsAS0Ms:-K8Y46XP0d82TrtvWXPb:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L210)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 302 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrO4TtzaaS0956KB8:-K8Y4976Ssfhu_uQ-DXW:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L302)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 312 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8Oh63LoCnj1GOu9LF-:-K8YEkaMQwLiCa0dVJih:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L312)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 118 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8V47MmjjAvcI_he-UO:-K8Y3QesB--u9F95pgKs:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6be56771e22e6cfb719f88f65bef05e180f6312b/thali/NextGeneration/thaliWifiInfrastructure.js#L118)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
Test (Success) 56449660 build is completed (0f5d794)

See https://github.com/ThaliTester/TestResults/tree/564496600f5d794_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 56449660311ec66(311ec66) has failed

See https://github.com/ThaliTester/TestResults/tree/56449660311ec66_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the fail logs

ThaliTester:
Test 564496600f5d794(0f5d794) has failed

See https://github.com/ThaliTester/TestResults/tree/564496600f5d794_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the fail logs

vjrantal:
Reviewed 1 of 2 files at r4.
Review status: 4 of 7 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
PR is added to the queue for testing as 1. task. (bb41d23)

ThaliTester:
Test (Success) 56449660 build is completed (bb41d23)

See https://github.com/ThaliTester/TestResults/tree/56449660bb41d23_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 56449660bb41d23(bb41d23) has failed

See https://github.com/ThaliTester/TestResults/tree/56449660bb41d23_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the fail logs

yaronyg:
Reviewed 2 of 2 files at r5, 4 of 4 files at r6.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 114 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_5SetXNmmOFxOUb1U:-K8_5SetXNmmOFxOUb1V:-1977466279) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L114)):*
The previous then clause should cause an Error to be thrown so this then clause should be skipped. So shouldn't it either be removed or to be really thorough shouldn't there be a t.fail() here?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 71 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6C_ykyCvveAY3Mp5:-K8_6C_zNxtWazxZfXyx:-1499475384) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6be56771e22e6cfb719f88f65bef05e180f6312b/thali/NextGeneration/thaliWifiInfrastructure.js#L71)):*
Did you change node-ssdp to use our logging? If not then shouldn't we turn off their logging? Have you investigated what they do in regards to logging?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 106 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6RZQQHQRx47aOZ3c:-K8_6RZQQHQRx47aOZ3d:1507912374) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L106)):*
Sorry to be a pain but it would also be good to check that it's a legal port. Technically ports can be from 0 to 65536 but in general the first 1024 ports are reserved and we really shouldn't be using them for a Thali scenario. So for now I'd check that the port was > 1024 and <= 65536.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 111 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6mSMCQvdbLpilJzE:-K8_6mSMCQvdbLpilJzF:536642982) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L111)):*
Are we confident that node-ssdp can't be fooled when setting the USN? For example, could they be tricked into producing a null or a number or a boolean? We need to check that the USN is a non-empty string.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 112 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6udFyOrW9lIm0ULZ:-K8_6udFyOrW9lIm0UL_:337523932) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L112)):*
Validating hostname is genuinely hard. We don't want people able to use us to attack other people. This actually happens with UPnP all the time (although to be fair only when the UPnP device is on the open Internet).

I randomly found https://www.npmjs.com/package/ip-address and https://www.npmjs.com/package/ip for validating IP addresses. In fact for now I think we should put in a requirement that we ONLY accept IP addresses. We won't accept a DNS address at all. Once we validate the IP address we have to make sure it is not 127.0.0.1 or equivalent for IPv6 (::1). We also have to make sure it is not our own public address.

I'm very tempted to also require that we make sure that the IP address is from one of the reserved local only ranges. But that may be going too far since at least in theory it's possible that someone has a wifi network configured to give out publicly routable addresses (especially on IPV6). But what I do think we should do is make sure that the IP address is in the local netmask range (https://www.npmjs.com/package/range_check might be useful for that). I just don't know if there is a reliable way to get the netmask on JXcore. They are based on 0.10 and I don't think it gives a standard way to get the netmask.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 344 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_AAd_gq-dEfP4xcOi:-K8_AAd_gq-dEfP4xcOj:1475483192) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L344)):*
I'm pretty sure this is a bug. Imagine that the router starts up fine and you call resolve(). Now later on there is an error. Now you are going to call reject() on a resolve() promise. I'm pretty sure that is bad.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
PR is added to the queue for testing as 1. task. (fbe3bb9)

ThaliTester:
PR is added to the queue for testing as 2. task. (5d11e75)

vjrantal:
Review status: 5 of 9 files reviewed at latest revision, 14 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 114 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_5SetXNmmOFxOUb1U:-K8x6e6VWmupPzVCHG2t:-608411843) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L114)):*
In the specification, it was specified that the call to `startUpdateAdvertisingAndListening` inspects the router object and `start` is just a fairly "dummy function". That is why I think the current test code is appropriate and reflects what was specified.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 155 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QO5hGbTNkn3MmWZ41:-K8x_9HR3kk9XtaUO1f9:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L155)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 205 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QPoXLMDxMTI-tSCCF:-K8x_cFXtvYWg7kXifsH:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L205)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 235 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QUD8FptCF1egERHxo:-K8x_PzgULlZKf5ww3ZF:290239292) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L235)):*
Will be fixed with promise queue.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 304 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrSYaKb-_VXeZC9Dd:-K8x_T_MOdXlj-P1B46M:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L304)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 306 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QreP9QxHA36zmFT08:-K8x_V-bCqn3pkPMhVXW:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L306)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 311 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QvZVitz-0iAthvXN_:-K8x_VubqWaeXfvDxUR9:290239292) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L311)):*
Will be fixed with promise queue.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 341 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QxzvyhLRaZi_e4uCk:-K8xb05kxyYYU97eQ0vR:290239292) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L341)):*
Will be fixed with promise queue.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 71 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6C_ykyCvveAY3Mp5:-K8x7ByxmsQ906pmHlPp:-307823950) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6be56771e22e6cfb719f88f65bef05e180f6312b/thali/NextGeneration/thaliWifiInfrastructure.js#L71)):*
Yes, I have checked.

When I made the fork, they were using logging that was incompatible with jxcore https://github.com/jxcore/jxcore/issues/654 .

However, they have recently moved to use a logging framework that would probably work. I'd have to rebase our fork onto the latest node-ssdp and then see if the logging works. I am not planning to do this in the context of this PR, but created a separate issue https://github.com/thaliproject/Thali_CordovaPlugin/issues/489.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 106 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6RZQQHQRx47aOZ3c:-K8xQgPKSdwN3JBi4mAp:-975982400) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L106)):*
Done.

I found a function `ensureValidPort` that validates the port and started using it. It seems to validate that the port is above 0, but I'd suggest modifying that in case we need to check for above 1024. The only thing I'd check is can ports lower than 1024 be ever returned when we pass 0 to https://nodejs.org/api/net.html#net_server_listen_port_hostname_backlog_callback (which is something we do quite a lot).

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 111 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6mSMCQvdbLpilJzE:-K8xYjPIhNJp03U68fN0:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L111)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 112 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_6udFyOrW9lIm0ULZ:-K8xZPuBe7AIfOLQrACG:1588248700) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L112)):*
I am already using the ip package so that could be a good choice. Since the specification above is not 100% accurate, I created this as a separate issue https://github.com/thaliproject/Thali_CordovaPlugin/issues/490.

Once we know exactly how to validate, I can implement. It might also make sense to have the implementation in `validations.js`or other centralized "security-module" so that it is easier to review and adjust the security-critical code.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 344 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_AAd_gq-dEfP4xcOi:-K8xabDpmzBNGCVht5Ea:-1275558454) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L344)):*
Here is the test code I used to try to determine if this is an issue https://gist.githubusercontent.com/vjrantal/e653bf2cdd8781b7b50f/raw/cb9ca5ea7d285e9104e9132a700adb52e7ed1b61/gistfile1.txt . Based on what that outputs, there isn't a problem calling reject on a handled promise.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
Test (Fail) 56449660 build is completed (fbe3bb9)

See https://github.com/ThaliTester/TestResults/tree/56449660fbe3bb9_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test (Success) 56449660 build is completed (5d11e75)

See https://github.com/ThaliTester/TestResults/tree/564496605d11e75_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 564496605d11e75(5d11e75) has failed

See https://github.com/ThaliTester/TestResults/tree/564496605d11e75_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the fail logs

yaronyg:
Reviewed 4 of 4 files at r7.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 205 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QPoXLMDxMTI-tSCCF:-K8z2dOJNd7tvOZfPFJc:1989551653) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L205)):*
I still see this.listening twice inside of this function.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 304 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8QrSYaKb-_VXeZC9Dd:-K8z3SVZ5X0O_Kau98lz:1744904372) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/31ce1407891ccaaf6ebbc185839d37a84431bb82/thali/NextGeneration/thaliWifiInfrastructure.js#L304)):*
According to RFC 4122 the format should be "urn:uuid:...."

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

yaronyg:

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 344 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466#-K8_AAd_gq-dEfP4xcOi:-K8zPUVkJqHHxoeKbY32:1327886900) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/bb41d2300bebdcf54a3f129b201bdd9b74d02bc3/thali/NextGeneration/thaliWifiInfrastructure.js#L344)):*
I tested it both in ES6 and with LIE and they both seem to allow a call to reject after resolve but keep in mind that in both cases the call to reject is a NOOP. That is not good coding practice. A developer is going to see that reject above and think it is meaningful but it may not be. That is not easy to understand code. You are essentially relying on a race condition.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

ThaliTester:
PR is added to the queue for testing as 1. task. (4206eac)

ThaliTester:
Test (Success) 56449660 build is completed (4206eac)

See https://github.com/ThaliTester/TestResults/tree/564496604206eac_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the logs

ThaliTester:
Test 564496604206eac(4206eac) has failed

See https://github.com/ThaliTester/TestResults/tree/564496604206eac_First_bits_of_the_Wifi_infrastructure_work_vjrantal/ for the fail logs

vjrantal:
Reviewed 1 of 4 files at r1, 2 of 2 files at r5, 2 of 4 files at r6, 2 of 4 files at r7, 2 of 2 files at r8.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/466)*

PR #268, opened 2015-10-30
Localhost GetDocumentsPath method update in mockmobile.js - fixes #233

deadlyfingers:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/268)

PR #494, opened 2016-01-27
Revert "Localhost GetDocumentsPath method update in mockmobile.js - fixes #233"

### LINKED ISSUE #233: Fix localhost version of Mobile.GetMobileDocumentsPath
Postcard app is now generating string error with Mobile.GetMobileDocumentsPath in 'mockmobile.js'

```
TypeError: Arguments to path.join must be strings
```

vjrantal:
Reverts thaliproject/Thali_CordovaPlugin#268

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/494)

### DIFF
--- a/test/www/jxcore/bv_tests/mockmobile.js
+++ b/test/www/jxcore/bv_tests/mockmobile.js
@@ -1,5 +1,4 @@
 "use strict";
-var os = require('os');
 
 var keys = {};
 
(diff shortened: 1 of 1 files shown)


vjrantal:
The change in the mock broke the test in file `testThaliCryptoManager.js` so reverting it. There is a better solution for the original problem.

PR #514, opened 2016-02-05
Completes Issue 209 - Implement Thali Notification Beacons

mpodwysocki:
Implements the spec as per the [Presence Protocol for Opportunistic Syncing](https://github.com/thaliproject/thali/blob/gh-pages/pages/documentation/PresenceProtocolForOpportunisticSynching.md)

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/514)


ThaliTester:
PR is added to the queue for testing as 1. task. (bd4e809)

ThaliTester:
Test (Fail) 58380500 build is completed (bd4e809)

See https://github.com/ThaliTester/TestResults/tree/58380500bd4e809_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (f6bada7)

ThaliTester:
Test (Fail) 58380500 build is completed (f6bada7)

See https://github.com/ThaliTester/TestResults/tree/58380500f6bada7_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (af797b8)

ThaliTester:
Test (Fail) 58380500 build is completed (af797b8)

See https://github.com/ThaliTester/TestResults/tree/58380500af797b8_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (69f842e)

ThaliTester:
Test (Success) 58380500 build is completed (69f842e)

See https://github.com/ThaliTester/TestResults/tree/5838050069f842e_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 5838050069f842e(69f842e) has failed

See https://github.com/ThaliTester/TestResults/tree/5838050069f842e_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

yaronyg:
Reviewed 4 of 6 files at r1, 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 23 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 1 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nbU2prSQUhdEdtwJN:-K9nbU2prSQUhdEdtwJO:-1426100278) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L1)):*
Please test the expiration logic to make sure we don't take tokens from the past or too far into the future.

Please also add a test that ensures that we actually pass the right unencryptedKeyId values in the right order to the addressBookCallback. You could easily adapt addressBookCallback returns public key to do this.

Please stick in a test with two beacons both for the target device just to make sure we actually stop after we find the first result.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 24 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nbxWgQ-KG779xcquK:-K9nbxWhEJXJXFjJN0dJ:-3257749) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L24)):*
Shouldn't we check what happens if we give it an empty array of keys? Or null keys?

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 40 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncAcgQm8xKH0Fz2SX:-K9ncAcgQm8xKH0Fz2SY:-1021795283) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L40)):*
Shouldn't we check that when we translate the expiration value back we get a plausible result? E.g. 9000 ms from now +/- say 100 ms?

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 82 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncgPsgrnkFo84Wk2s:-K9ncgPt9hcr3UbfyVlz:554158089) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L82)):*
I'd actually put in a t.fail() here since if the address book gets called at all something very bad has happened.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 85 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncYP-hD6iaRUhpl_J:-K9ncYP-hD6iaRUhpl_K:-791637976) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L85)):*
We need to validate that we threw the right exception. Eventually thought this should be a check that the promise returned the right Error object.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 109 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncz29jx828Pgn7s4O:-K9ncunYCTq3GTEiyDOk:-96485542) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L109)):*
I'd actually say that no beacons at all is an error. Nobody should be bothering to advertise an ephemeral key and an expiration if there are no beacons to go along with it.

---

*[test/www/jxcore/package.json, line 21 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n0VwFeAmP-GlzrcWc:-K9n0VwGaumPTaN2ozer:1973020233) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/package.json#L21)):*
Are we still using bn.js?

---

*[test/www/jxcore/package.json, line 43 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n0kkXyQkmcR9pYr4d:-K9n0kkYG0u7swhDlPQg:2059340729) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/package.json#L43)):*
Are we still using supertest? Also, since we are already using supertest-as-promised couldn't we just use that instead since I believe it includes all of supertest?

---

*[thali/NextGeneration/hkdf.js, line 68 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n3Vf5NAKaIYvS9yr1:-K9n3Vf5NAKaIYvS9yr2:-1540147018) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/hkdf.js#L68)):*
I'm really worried about this code making us non-responsive. Every call to update will require a native round trip and I wouldn't be surprised if createHmac doesn't as well. That means we are non-responsive while this is happening. Now maybe for the 4K sized values we think we will create it doesn't matter? I can see two approaches. One approach is to run perf tests on both iOS and Android and see what numbers come back and that way we will know if there is an issue. But I think there is an easier solution which is just to use setImmediate. This would be a really straight forward recursive function. It shouldn't take more than 20 minutes to write.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 1 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n6aQhGlVw7lpSdnlq:-K9n6aQhGlVw7lpSdnlr:1087718543) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L1)):*
This doesn't pass JSCS/JShint validation. It only took a second so I fixed it myself. I'll submit a PR. But in the future please check this before checkin.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 19 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n5yYkTiDXkGt3yZHh:-K9n5yYlTkSjpcrDgMWd:1859146165) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L19)):*
Do you think this needs to be a constructor? Why not just export generatePreambleAndBeacons and the other function directly?

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 35 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nIGLaOH96HVyxJH1X:-K9nIGLaOH96HVyxJH1Y:1275812652) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L35)):*
Why go through the work of adding the beacons to the array just to pull them out again and stick them into a buffer? Why not just append the beacons to the output buffer as you create them?

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 49 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nJUoxi0b61JojAfet:-K9nJUoyHqRKeG6znOQK:-585978309) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L49)):*
This would be very useful as a generic function that can take a public key and create the hash short hand. I would like to use this in the PouchDB replication code to record the document where a user stores replication information. Also anyone implementing addressBookCallback will need it too.

Actually I implemented this in the PR I'll send you as I needed it to write a test.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 54 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nJo9UTsMptEFbI6sf:-K9nJo9UTsMptEFbI6sg:-613550194) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L54)):*
Wouldn't it be less error prone and easier to read if we just used publicKeysToNotify.forEach? E.g. publicKeysToNotify.forEach(function(pubKy) {... })?

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 58 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nKKfpkc5IwwTSRafF:-K9nKKfqBesATEdPeGiX:135020422) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L58)):*
Do we really need to call createECDH and setPrivateKey on each iteration? Couldn't we create it once with ecdhForLocalDevice.getPrivateKey and then call computeSecret over and over again inside the loop?

I'm assuming, btw, that it's possible to call computeSecret multiple times. I haven't tested that theory.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 118 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nLrxb2gW3o3OkoonI:-K9nLrxb2gW3o3OkoonJ:-1739810473) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L118)):*
I would suggest a slightly more informative name for chunkSize such as beaconSize.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 121 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nMUXNZosRh79kk8f3:-K9nMUXNZosRh79kk8f4:-986538493) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L121)):*
The use of slice worries me because it makes the code hard to read. It's just so easy to get lost. Have you considered something like https://github.com/gagle/node-streamifier? It's a tiny bit of code that turns a buffer into a stream. Then you can use stream.read to pull off the bytes you need and the next section of code pulls off the bytes it needs and so on. That way we don't need to keep a cursor pointing to where we are in the buffer. The stream handles that for us. It also means we can get rid of the for loop to parse the individual beacons and instead just use a while that checks if the read is null. The beauty of this is that it will handle cases where we have malformed (e.g. too short) beacons since read(size) returns null if it can't pull of size bytes. What do you think?

BTW, just to be clear, I am suggesting using streams in non-flowing mode. I think writing it in flowing mode with event handlers would cause serious headaches. :)

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 123 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nM1KMp56D_JTtspsv:-K9nM1KMp56D_JTtspsw:-695402121) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L123)):*
Once we make HKDF async this function will have to return a promise anyway so why not just have this resolve the promise as an Error? Same for the other exceptions below.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 128 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nPGFB7_kmKNrHjqN8:-K9nPGFB7_kmKNrHjqN9:776408719) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L128)):*
Shouldn't we also check that the expiration encodes a time in the future? Also the spec states we shouldn't accept beacons that are too far into the future. So if the expiration is for a time more than say 24 hours (make that a const) into the future then we should reject.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 156 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nQkFzww35vd6IYdMU:-K9nQkFzww35vd6IYdMV:-728913247) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L156)):*
Why are we using the same variable name for two different purposes?

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 168 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nRCVpmN0WgNi9dbtW:-K9nRCVqfMXLpkS01VhB:1152456376) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L168)):*
This is unfortunately a major problem with our inability to use GCM. With GCM when we get a beacon we know for sure if it is intended to be addressed to us because the GCM tag check confirms it. But now we don't have that check because JXcore doesn't support GCM. Which means that if we get a beacon that was not intended for us then our decrypt will still return a value but the value should be nonsense. So unfortunately this create an ambiguity where the fact that the address book doesn't have a matching value could be because there is no matching value or it could be because it's gibberish.  So now the best we can do is just continue on to the next beacon if we don't get a match since we can't be sure what caused the issue.

I already put the fix for this into my PR.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 188 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nabukaBbdz8ew_4bX:-K9nabukaBbdz8ew_4bY:421447667) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L188)):*
You are right, they don't support equals. But they do support compare. Please see [here](http://jxcore.com/docs/buffer.html#bufcompareotherbuffer).

---

*[thali/package.json, line 11 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n19iC5Zkg8sAH9aYa:-K9n19iDzvPUwN43AjUu:-1886433230) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/package.json#L11)):*
Why would be go from 3.0.2 to 3.0.1?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

ThaliTester:
PR is added to the queue for testing as 1. task. (fccea7e)

ThaliTester:
Test (Success) 58380500 build is completed (fccea7e)

See https://github.com/ThaliTester/TestResults/tree/58380500fccea7e_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 58380500fccea7e(fccea7e) has failed

See https://github.com/ThaliTester/TestResults/tree/58380500fccea7e_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

vjrantal:
Review status: 3 of 6 files reviewed at latest revision, 24 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 3 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9pWssQeR6AAKZ7IzM1:-K9pWssQeR6AAKZ7IzM2:1817782594) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/fccea7e88e56be399f07ec44c0d717ba4f438ee9/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L3)):*
If possible, would be good to use thali-tape, because only that works correctly with the test coordination server. It should have logic to default to wrapping-tape if not environment where we run tests in a coordinated manner (https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext/test/www/jxcore/lib/thali-tape.js#L249).

---

*[test/www/jxcore/package.json, line 43 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n0kkXyQkmcR9pYr4d:-K9pVnxGoiC-nUjtA2bp:1278564292) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/package.json#L43)):*
(off-topic: one of the features I dislike in reviewable is that in the default review flow, it effectively hides which commit introduces which change)

This was my change that came via merge from `vNext`. I added it, because supertest is a peer dependency (https://github.com/WhoopInc/supertest-as-promised/blob/master/package.json#L32) and at least latest npm version doesn't pull those automatically.

I think we currently use supertest only in the disabled identity exchange tests so in reality, this would not really matter much at the moment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 24 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nbxWgQ-KG779xcquK:-KA0SG0Y9z-35X3pQa4d:1693443325) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L24)):*
Is null an error?  Empty will be just fine in that it won't contain any beacons to parse

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 1 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nbU2prSQUhdEdtwJN:-KA0TVH554q9n6jT9O8t:-774980530) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L1)):*
What is the +- for too far in the future?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 3 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9pWssQeR6AAKZ7IzM1:-KA0TdbDPeDlQy2k-u0R:-1453037224) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/fccea7e88e56be399f07ec44c0d717ba4f438ee9/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L3)):*
That doesn't seem to work on the desktop so someone needs to fix thali-tape because running it via jx does nothing.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 3 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9pWssQeR6AAKZ7IzM1:-KA0To1hO3AVlqAY9eVU:-509013559) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/fccea7e88e56be399f07ec44c0d717ba4f438ee9/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L3)):*
And no, I don't want to run the whole test suite, just my tests which runTests do

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 85 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncYP-hD6iaRUhpl_J:-KA0Wwa2GJqurFbzmZC4:-132150771) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L85)):*
Do you want to have separate subclasses for each test then?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 109 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncz29jx828Pgn7s4O:-KA0XL62YLQG0ZBmt2Uv:-761671151) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L109)):*
It's not an error, it's just returning null. Not aware of me saying this was an error more than it was not going through my loop logic

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[test/www/jxcore/package.json, line 21 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n0VwFeAmP-GlzrcWc:-KA0XTosdW4QKNodthd5:1634638716) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/package.json#L21)):*
I saw that in the jx install, so I would say so, but I'm not sure

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/hkdf.js, line 68 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n3Vf5NAKaIYvS9yr1:-KA0Xilv9OWyb04Nw9Ef:825028789) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/hkdf.js#L68)):*
I really don't think for the small number that we're actually using that its a sizable perf hit.  Do we want to add the benchmarks for this?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 19 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n5yYkTiDXkGt3yZHh:-KA0Y8t4sc6zNOpw7hAi:1040089851) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L19)):*
No, I don't think this needs to be a class, just two exported functions. Just rolled with what you had here.  Yeah, there is no need for any instance variables, so yeah, two static functions work just fine.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 35 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nIGLaOH96HVyxJH1X:-KA0YTS74g21tF9t_zk3:1449269530) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L35)):*
Buffer.concat might be a costly operation if done often, hence why doing it at the end was much better since Array.prototype.push is really really cheap

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 54 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nJo9UTsMptEFbI6sf:-KA0fNmOxsDVZFfdSplF:1910602718) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L54)):*
Somewhat, yes, as it will deal with dense arrays instead of sparse arrays, so point taken, therefore no dealing with empty spaces/undefineds in our arrays

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

vjrantal:
Review status: 3 of 6 files reviewed at latest revision, 24 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 3 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9pWssQeR6AAKZ7IzM1:-KA0f-bN8yeYTqXC-9ic:815139326) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/fccea7e88e56be399f07ec44c0d717ba4f438ee9/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L3)):*
Ok - maybe something wrong somewhere then..

On my machine (and with your branch), I was able to use thali-tape and do:
```
$ jx bv_tests/testThaliNotificationBeacons.js
```
And all tests in that file passed.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 58 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nKKfpkc5IwwTSRafF:-KA0ffQaKCBZ-8C5YZ_V:129848260) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L58)):*
Yes, I had that originally and was going back and forth while trying to get the code to work.  Let me revert

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 121 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nMUXNZosRh79kk8f3:-KA0mCWbNumbtYPblkHY:-1713066855) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L121)):*
Is it worth it to increase the bloat of our framework by adding a multi-purpose library that does just one thing that we need to do which is just reading a byte buffer? I mean both work I'm just skeptical that adding one library is worth it given the simplicity otherwise.

Either way works it's just a tradeoff

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 123 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nM1KMp56D_JTtspsv:-KA0s5tBcSAPuQbeNgzB:59268348) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L123)):*
I say we back it up with numbers before we add the extra overhead of Promises (and yes, it is overhead of a forced async tick)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 128 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nPGFB7_kmKNrHjqN8:-KA0sRsPKKQQ6AtLq2Vp:1276628806) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L128)):*
Ok, so it should be a constant of one day that is too far in advance?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 156 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nQkFzww35vd6IYdMU:-KA0sYcJL3OZ59aEyr6Z:-447993538) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L156)):*
Variable reuse because the end name is really what I wanted and wasn't any need to rename the decipher to something else. But I can change if that's a big issue

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/NextGeneration/thaliNotificationBeacons.js, line 188 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nabukaBbdz8ew_4bX:-KA0smKB93trIwTN7LEs:1924548584) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L188)):*
I'd rather use something that is Node neutral that 0.10.x nor 0.12.x have versus something that only JXcore has so once we move off the merge is painless from this code standpoint.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

mpodwysocki:
*[thali/package.json, line 11 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n19iC5Zkg8sAH9aYa:-KA0t21SUrQGD3MGuTu1:2059743252) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/package.json#L11)):*
I don't see any difference between from and to here, but I can certainly up the version...  Wasn't caught in the merge from vNext

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

yaronyg:
Review status: 3 of 6 files reviewed at latest revision, 16 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 1 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nbU2prSQUhdEdtwJN:-KA1c04Xb3hMzqkLG-L4:827918401) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L1)):*
See the spec, it suggested 24 hours but in any case this should just be some constant we can change later.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 24 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nbxWgQ-KG779xcquK:-KA1c9FJOYNsSCfjs6TI:-1495274023) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L24)):*
We shouldn't be outputting a beacon string that contains a pre-amble and no values. So if we get null or empty we should return a null value.

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 85 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncYP-hD6iaRUhpl_J:-KA1cJHhpCOAu6oXmW-Y:-63320384) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L85)):*
See skype conversation. :)

---

*[test/www/jxcore/bv_tests/testThaliNotificationBeacons.js, line 109 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9ncz29jx828Pgn7s4O:-KA1cPaVgvlO2u3xiyd_:-2042401153) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js#L109)):*
I just meant that if someone is advertising a pre-amble and no beacons then something bad has happened.

---

*[test/www/jxcore/package.json, line 21 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n0VwFeAmP-GlzrcWc:-KA1cZL2Y6UIeotvfhN0:934395643) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/package.json#L21)):*
Is this another @vjrantal add in from vNext? Reviewable made it look like you were the one who made this change so I was hoping you would know what to do with it.

---

*[thali/NextGeneration/hkdf.js, line 68 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n3Vf5NAKaIYvS9yr1:-KA1czIXwFKtZdCMcv5E:1966340057) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/hkdf.js#L68)):*
It's your call. Either benchmark or async. But please make sure the benchmark, if you go that way, is run on actual mobile devices.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 35 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nIGLaOH96HVyxJH1X:-KA1d4r6N8_Ie2FKDZic:2127641648) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L35)):*
But we don't support much more than 10 or 20 beacons so we wouldn't be doing it often. No premature perf optimizations (hence why I'm happy to do the benchmark vs the async for HKDF).

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 121 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nMUXNZosRh79kk8f3:-KA1dHnOLAvcGeu5dbgO:-1890462179) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L121)):*
The dependency added by node-streamifier is one tiny file and in return we take complicated error prone code and make it easier to read. I'd say it's worth the trade off but this is your code and my rule is that unless there is a killer reason to do something it's the dev's choice, so it's your choice.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 123 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nM1KMp56D_JTtspsv:-KA1dVu0kw_S-AhXze4k:-454366630) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L123)):*
Fair enough. As I said above, this one is your call. We can either benchmark or make it async. Just make sure you run that benchmark on a mobile device.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 128 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nPGFB7_kmKNrHjqN8:-KA1dcF4FyopgLY1Ru0L:-1629531620) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L128)):*
Yeah I picked that number out of the air. We can make it a hard coded figure for now.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 156 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nQkFzww35vd6IYdMU:-KA1dhl_XyiVLvgaa8ur:1053867467) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L156)):*
It's not so much a big issue as something that looks like a bug.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 188 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9nabukaBbdz8ew_4bX:-KA1dnBp3JOGh9pfooXK:220898296) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/NextGeneration/thaliNotificationBeacons.js#L188)):*
That boat sailed a long time ago. We are tied to JXcore. But if it makes you feel better can you can create a buffer.equals using compare and make it a polyfill.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

ThaliTester:
PR is added to the queue for testing as 1. task. (055030c)

ThaliTester:
Test (Success) 58380500 build is completed (055030c)

See https://github.com/ThaliTester/TestResults/tree/58380500055030c_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 58380500055030c(055030c) has failed

See https://github.com/ThaliTester/TestResults/tree/58380500055030c_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (962e22b)

ThaliTester:
Test (Success) 58380500 build is completed (962e22b)

See https://github.com/ThaliTester/TestResults/tree/58380500962e22b_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 58380500962e22b(962e22b) has failed

See https://github.com/ThaliTester/TestResults/tree/58380500962e22b_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (c26a9ef)

ThaliTester:
Test (Success) 58380500 build is completed (c26a9ef)

See https://github.com/ThaliTester/TestResults/tree/58380500c26a9ef_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 58380500c26a9ef(c26a9ef) has failed

See https://github.com/ThaliTester/TestResults/tree/58380500c26a9ef_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (ee4964c)

ThaliTester:
Test (Fail) 58380500 build is completed (ee4964c)

See https://github.com/ThaliTester/TestResults/tree/58380500ee4964c_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (4251330)

ThaliTester:
Test (Success) 58380500 build is completed (4251330)

See https://github.com/ThaliTester/TestResults/tree/583805004251330_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (4f2b042)

ThaliTester:
Test 583805004251330(4251330) has failed

See https://github.com/ThaliTester/TestResults/tree/583805004251330_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

ThaliTester:
Test (Success) 58380500 build is completed (4f2b042)

See https://github.com/ThaliTester/TestResults/tree/583805004f2b042_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 583805004f2b042(4f2b042) has failed

See https://github.com/ThaliTester/TestResults/tree/583805004f2b042_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

yaronyg:
Reviewed 1 of 3 files at r5, 1 of 2 files at r6, 1 of 1 files at r10, 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[test/www/jxcore/package.json, line 43 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n0kkXyQkmcR9pYr4d:-KA6qExpZ87Lxk2G0kiE:-340816844) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/test/www/jxcore/package.json#L43)):*
Given the new npm behavior for peer dependencies we need to include supertest.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 168 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-KA6smmdZp0A_HwvMmJ1:-KA6smmdZp0A_HwvMmJ2:599564465) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4f2b0425b32e92eed53fe2ef55fec0e2534554fc/thali/NextGeneration/thaliNotificationBeacons.js#L168)):*
this is actually supposed to be an unsigned integer. Does Long support unsigned values?

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 175 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-KA6rPu1F_vDqlXbPKEE:-KA6rPu1F_vDqlXbPKEF:515671543) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/4f2b0425b32e92eed53fe2ef55fec0e2534554fc/thali/NextGeneration/thaliNotificationBeacons.js#L175)):*
I don't understand this code. If you receive an expiration of 1, what does that mean? 1 what? If someone generated an expiration two months ago and two seconds ago, how would you tell the difference?

---

*[thali/package.json, line 11 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514#-K9n19iC5Zkg8sAH9aYa:-KA6qTsPzJmZ3aS5TCeb:252781342) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/69f842e940ae9d0ef6d65bd4bd1edc37f4fd7f44/thali/package.json#L11)):*
Please set back to 3.0.2

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/514)*

ThaliTester:
PR is added to the queue for testing as 1. task. (3a0697c)

ThaliTester:
PR is added to the queue for testing as 2. task. (306a2c5)

ThaliTester:
Test 583805003a0697c(3a0697c) has failed

See https://github.com/ThaliTester/TestResults/tree/583805003a0697c_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

ThaliTester:
Test (Success) 58380500 build is completed (306a2c5)

See https://github.com/ThaliTester/TestResults/tree/58380500306a2c5_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 58380500306a2c5(306a2c5) has failed

See https://github.com/ThaliTester/TestResults/tree/58380500306a2c5_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (a584679)

ThaliTester:
Test (Success) 58380500 build is completed (a584679)

See https://github.com/ThaliTester/TestResults/tree/58380500a584679_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the logs

ThaliTester:
Test 58380500a584679(a584679) has failed

See https://github.com/ThaliTester/TestResults/tree/58380500a584679_Completes_Issue_209_-_Implement_Thali_Notification_Beacons_mpodwysocki/ for the fail logs

Issue #318, opened 2015-11-12
Create spec for ACL layer


yaronyg:
This spec will define how we control access to all of Express-PouchDB's endpoints as well as our own endpoints and decide who can make requests and who cannot.

We also have to protect _local so that remote requesters can only see entries they created. We almost certainly have to put in some kind of key protection to make sure that IDs can't collide as this could be used to cause malicious behavior.

Note that with the exception of _local we only support GET and sometimes POST on a few endpoints needed for pull replication. For _local we have to allow write both to the magic ID that PouchDB will create on its own (and somehow prevent collisions so bad guys can't screw things up for other folks) but we also need to allow write for our own magic _local value.

We also need to make sure that we can associate an authenticated ID via PSK with a particular native level connection. Trust is still at the TCP layer but this does help the peer pool implementation when it has to pick a connection to kill.

Issue #320, opened 2015-11-12
Create spec to secure communication with the WebView


yaronyg:
We need to define for iOS and Android (and hopefully soon UWP) how we secure communications between the WebView and the JXCore instance so random apps on the device can't just connect to localhost and send random commands to our node code.

mohlsen:
@dakr0 has been working on this for [@project_stanton](https://twitter.com/project_stanton), using [salti-admin](https://github.com/thaliproject/salti-admin) as part of the process.  We would be happy to write something up for this.  @yaronyg how would you like us to post the spec?

yaronyg:
@mohlsen I see two choices. One is to update https://github.com/thaliproject/thali/blob/gh-pages/pages/TechnologyInvestigations/SecuringCordovaAndNodeJs.md but perhaps better is to submit a PR to create a new doc in https://github.com/thaliproject/thali/tree/gh-pages/pages/documentation with an actual spec.

yaronyg:
Where is the spec on this?

larryonoff:
@yaronyg please see below the document that captures the investigation results available so far and detail the next steps necessary to find a suitable solution to the issue.

### Suggested approach

As of today, we've evaluated the potential approaches outlined in [Thali documentation](http://thaliproject.org/SecuringCordovaAndNodeJs/):

1. Secret in the URL
2. Secret in a cookie
3. Secret in a HTTP header
4. TLS mutual authentication

The solutions 1-3 all have potential issues (including, but not limited to those described in Thali documentation), therefore we believe that proceeding with solution _4 TLS mutual authentication_ - is the most promising. This approach should work as follows:

1. App generates client and server certificates when it starts and saves them into a shared storage.
2. When node.js server starts, it uses server certificates generated on step 1.
3. WebView uses client certificates generated on step 1 by sending them with any request to node.js server.

Each of these steps has a number of open issues, which are described below.

### Further steps necessary

In order to provide a definite recommendation on a solution, we need to continue investigation of the applicability of TLS mutual authentication approach to JXcore to WebView paring process and create an actual proof of concept. This will involve covering the following points:

1. Investigate certificate generation to allocate this to a correct system module. Decide when exactly should the certificates be generated (on first launch only or on each launch).
2. Verify that it is possible to organize a shared local storage for WebView and node.js server.
3. Configure JXcore server to use the generated certificate at startup.
[...]

PR #520, opened 2016-02-09
#318 ACL Spec

yaronyg:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/520)


ThaliTester:
PR is added to the queue for testing as 1. task. (3da948e)

ThaliTester:
Test (Success) 58698349 build is completed (3da948e)

See https://github.com/ThaliTester/TestResults/tree/586983493da948e__318_ACL_Spec_yaronyg/ for the logs

ThaliTester:
Test 586983493da948e(3da948e) has failed

See https://github.com/ThaliTester/TestResults/tree/586983493da948e__318_ACL_Spec_yaronyg/ for the fail logs

cicorias:
I would say that w/ out a working skeletal implementation, isolated as a POC it's hard to fully vet a design here.  The concepts all are sound, but reality/practicality is needed.

Is there a working example somewhere?

Also, there's many "capabilities' within this overall "feature" - so, I'd suggest prioritizing each capability within this BIG feature to ensure we build only what's really needed by 'New Infra' timeframe.

---

Review status: 0 of 2 files reviewed at latest revision, 13 unresolved discussions.

---

*[thali/NextGeneration/thaliACLLayer.js, line 56 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGX3-FI32eKlKPq8tV:-KAGX3-FI32eKlKPq8tW:-736280290) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L56)):*
wouldn't this just be middle-ware in nodejs/express terms?  Does it need to be per path?
```
var apiRoutes = express.Router(); 
...
app.route('/foo').all( function(req, res, next))..
```

---

*[thali/NextGeneration/thaliACLLayer.js, line 62 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGg1UmiMr_XZ2zyQGm:-KAGg1UmiMr_XZ2zyQGn:1755981054) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L62)):*
Can these "routes" be separated or a distinct server listening on localhost for admin traffic? 

---

*[thali/NextGeneration/thaliACLLayer.js, line 68 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGaTWzdUGQM0rTi2kP:-KAGaTWzdUGQM0rTi2kQ:-2119499730) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L68)):*
I'm assuming this is NOT TLS - and really port 80/clear... 

why not use basic w/ some well known username!pwd(b64) or NO header even temporary.  It's really not any more secure, but putting in CLEAR is a total hack. it's already CLEAR to start with. If using existing basic authn then libraries already can deal with this new type of AuthN header.

To pile on, why not just use a PSK of "foorbarblah" or something else?  We end up with the same result in terms of exposure, but then we also get a secure channel.

---

*[thali/NextGeneration/thaliACLLayer.js, line 85 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGbjiLpK1cIzdUu5Sk:-KAGbjiLpK1cIzdUu5Sl:1872129581) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L85)):*
How "public" is this Un-authn traffic? ~~~can it just be "guest" as the identity - regardless, we need an identity that the server can then find a PSK to use.  I don't believe an empty identity works on the PSK handshake.~~~  Ahh, I see 'beacons' as below and a array of zeros..

---

*[thali/NextGeneration/thaliACLLayer.js, line 109 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGcpngZXoChnWqgi8b:-KAGcpngZXoChnWqgi8c:87303848) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L109)):*
any reason to not just use `/:db/` and allow that?  are there other paths/methods that are not part of this white-listing? 

eg. I don't see `OPTIONS` in the mix, will there ever be a need for pre-flight/options calls?  

I also see a  `app.get('/_session')` in the express-pouchdb [code](https://github.com/pouchdb/pouchdb-express-router/blob/master/index.js#L28) 

---

*[thali/NextGeneration/thaliACLLayer.js, line 143 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGeejuChIk4In9WvFg:-KAGeejuChIk4In9WvFh:-2031542959) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L143)):*
this can be done in express by adding a middleware function early on - eg: the following allows anybody to call CORS, but clearly any fn will do.
```
// allow on CORS
app.use(function(req, res, next) {
  /**
   * TODO: caller verification is part of the mix....
   * 1. caller needs to be validated against whitelist; all other's are denied.
   * */
  //just get the caller and allow it...
  var caller = req.get('origin');
  /// TODO: this needs to be validated - the caller - and only allow whitelist ones
  res.header("Access-Control-Allow-Origin", caller);
  res.header("Access-Control-Allow-Credentials", true);
  res.header("Access-Control-Allow-Headers", "Origin, X-Requested-With, Content-Type, Accept");
  next();
});
```
ex [here](https://github.com/cicorias/central-authentication/blob/master/server.js#L34)

---

*[thali/NextGeneration/thaliACLLayer.js, line 146 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGf07jCd7P983tQSVC:-KAGf07jCd7P983tQSVD:-888454230) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L146)):*
is ipv6 at all a need or can happen?  Also, the server can also bind to only the loopback address as well, instead of 0.0.0.0 or '::'

So in other-words, both should probably be in place. 

---

*[thali/NextGeneration/thaliACLLayer.js, line 148 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGgPUHrkhx2QcFJsAd:-KAGgPUHrkhx2QcFJsAe:-68883038) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L148)):*
again, if we can leverage basic authn and existing libs, then extract that concern from our primary code path.

---

*[thali/NextGeneration/thaliACLLayer.js, line 199 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGhnSQNg5Apo46Xs5r:-KAGhnSQNg5Apo46Xs5s:1553220769) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L199)):*
if we can map a path + verb to something tabular, then something like MustBe or express-authz can work.

beyond functional authorization, there's "object" authz. So, while I can do something , what about specific objects to which I have levels of entitlement?  So, each object has to have some affinity relationship to the identity in play in order to triangulate for entitlement rules.

---

*[thali/NextGeneration/thaliACLLayer.js, line 209 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGkpOJbN6dTA0nACpg:-KAGkpOJbN6dTA0nACph:-733572526) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L209)):*
setting the header on response tells the client to disconnect (res.set('connection', close'). res.end will not, 

server side you'd have to destroy the socket for each connection. 

here's a good tip on SO...  
[http://stackoverflow.com/questions/14626636/how-do-i-shutdown-a-node-js-https-server-immediately/14636625#14636625](http://stackoverflow.com/questions/14626636/how-do-i-shutdown-a-node-js-https-server-immediately/14636625#14636625)

---

*[thali/NextGeneration/thaliACLLayer.js, line 230 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGm2w-8LDqhNvUHsS5:-KAGm2w-8LDqhNvUHsS6:-510434740) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L230)):*
face?  'race'

---

*[thali/NextGeneration/thaliACLLayer.js, line 271 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGmUQP9Gn1dtE9hEDM:-KAGmUQP9Gn1dtE9hEDN:-1927185976) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L271)):*
this whole area just seems to be regular "authentication" needs in that it shoves into context some "identity" that shows up via the REST call, then we supplement that context with other "bits/attributes" that downstream can use.

---

*[thali/NextGeneration/thaliACLLayer.js, line 290 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGmkC2cwlpb9Q00N83:-KAGmkC2cwlpb9Q00N84:-1355371541) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L290)):*
ditto on this - why is it called 'validatePskIdentity' - why not 'validateIdentity' - or is this some layer below the regular authentication boundary middle ware? which that middleware should supplement the context(request)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-:-KAGoGorWSbja4NvlztL:1729298229)*

cicorias:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520)*

cicorias:
@yaronyg - not sure who else needs to review..

yaronyg:
I intentionally wrote requirements into the spec to explain what we think we need for New Infra. So if you see a feature in here that isn't necessary to meet the requirements then we don't need it.

That having been said, the purpose of this review is for you to ask any questions you have and get clarifications before you start coding up what is defined here. Obviously there will be more issues down the road.

For now you are the only person who needs to review this because you are the one implementing it. :)

---

Review status: 1 of 2 files reviewed at latest revision, 13 unresolved discussions.

---

*[thali/NextGeneration/thaliACLLayer.js, line 56 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGX3-FI32eKlKPq8tV:-KAGr-Ir1VHE4f5KV_0s:1599637803) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L56)):*
Hum... I'm not sure what in the language above made you think this was required on each path. The only thing that matters is that in Express routers are applied in the order they are registered. So if someone were to apply Express-PouchDB before our ACL router then their code would get to run before our code and unless they call next() (and they wouldn't) we wouldn't ever get to enforce ACLs. So my point was that when the original router object is created we need to be able to add our router to it before anyone else. 

So I would expect us to create our own express.Router() instance and put app.use("/",)... in it to catch every request for every path and method. I would then expect our use to be registered before everyone else's so we can guarantee we go first.

---

*[thali/NextGeneration/thaliACLLayer.js, line 62 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGg1UmiMr_XZ2zyQGm:-KAGs3k5O3ffwyvIq0oi:189059185) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L62)):*
We are always listening on localhost for everything. The reason has to do with how P2P works on non-TCP stacks (e.g. over Bluetooth or Multi-peer connectivity framework (MPCF)). When we get an incoming request over Bluetooth from another device we bridge that to a local TCP client who will then connect to the localhost version of our Express server. So by the inherent nature of bridging TCP over Bluetooth and MPCF we are forced to run a server instance on localhost.

One could argue that we could choose to run that internal instance on a publicly routable address but this assumes that WiFi is on and that we want to be exposed via WiFi. By using localhost we are able to enforce if we only want connections over Bluetooth/MPCF by having an address (a localhost address) that can only be reached by the Bluetooth/MPCF stack.

Also keep in mind that for admin operations we obviously only want to send requests over localhost.

Note, btw, that when WiFi is activated we actually host a second separate server that is on a public IP. Both server instances however run the exact same router! I tested this and it worked fine with Express-PouchDB.

So the bottom line is that we will have a single router object that we will have registered our ACL server on, Express-PouchDB on and our notification beacon endpoint. We will host that router object twice, once on localhost for non-TCP transports and a second time on a public address.

---

*[thali/NextGeneration/thaliACLLayer.js, line 68 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGaTWzdUGQM0rTi2kP:-KAGtNnf52pDqKKW4oQp:1778847160) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L68)):*
So let's step back for a second and think about the security model. The CLEAR mechanism is only intended to be used between the webview and the local node.js instance. That communication will be over localhost and thus cannot be intercepted by anyone. In fact the only real attack is if someone manages to hijack the port that the node.js server is running on. To do that the attacker would have to be on the local machine. Nevertheless that is still a real threat and we have an issue, https://github.com/thaliproject/Thali_CordovaPlugin/issues/320 that covers it but that we have made Rockwell aware we won't have ready by the New Infra time frame. The eventual goal there, btw, is to provision the WebView with a custome dynamically generated CA cert (all the code is already in OpenSSL) and then have the server expose an endpoint with that cert. That way the webview knows it has connected to the right server and the server will know, via CLEAR, that it has the right client.

So even with TLS added, CLEAR is still secure.

Now, if we could use PSK that would be cool. But no WebView I'm aware of supports PSK so I don't see anyway for the Cordova code in the WebView to create a PSK connection. The best we could do would be to write a Cordova Plugin which then talked to our code to use PSK but at that point all communications would have to be over strings and we would be using a MOCK for XMLHTTP Request. I leave the perf implications of that as an exercise for the reader.

I did consider using basic or digest auth. Neither offers anything additional in terms of security, I would go farther and argue that the way they are implemented makes them less secure. And unfortunately they appear to complicate our code. The reason is that once one turns on name/password security in PouchDB then all requests have to be through the name/password database. So to make this work we would have to create a bogus name/password to submit on all of our requests for the anonymous and pull replication roles.

My guess is that it will be faster and less error prone to just do a quick header check on incoming requests then bring into play the full pouchdb-security module and deal with all the implications of that.

---

*[thali/NextGeneration/thaliACLLayer.js, line 85 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGbjiLpK1cIzdUu5Sk:-KAGvA9QUaFKdN_p1y8z:-997933524) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L85)):*
Yup. so it's as public as public can be. :)

---

*[thali/NextGeneration/thaliACLLayer.js, line 109 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGcpngZXoChnWqgi8b:-KAGvuF1g9erE0CauMKE:234357627) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L109)):*
Oh, there are tons and tons and tons. Check out http://docs.couchdb.org/en/latest/api/index.html for an almost complete list. Many of those paths do very scary things like _purge and _design. What I did was go through every single path both in the spec as well as in Express-PouchDB's code and only white list the ones that I believe are needed for pull replication. Everything else is dangerous and so banned.

I didn't include OPTIONS because when I searched through the PouchDB code I never found a reference to it. I *think* that our Cordova webviews are set up so they don't need to do pre-flighting but if I'm wrong then we can easily add in CORS support.

_session is part of pouchdb-authentication (https://github.com/nolanlawson/pouchdb-authentication), something I'm trying to avoid for reasons previously discussed in the context of CLEAR. But don't take my word for it! Go look at pouchdb-authentication and pouchdb-security yourself and decide if you think they are a better choice. I mention them in the pre-existing work section of the doc and why I choose not to use them. But I have no illusions of infallibility.

---

*[thali/NextGeneration/thaliACLLayer.js, line 143 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGeejuChIk4In9WvFg:-KAGxCjBMegjBu3pgImY:-65345131) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L143)):*
My understanding of Express (and I'm no expert) is that the way Express works is that as router objects (middleware) are registered they are put into an array in the order they were registered. When a request comes in the request is sent to the first router in the array. If there are no matching paths then the next router is called. But if there is a matching path (such as all) then the request will be handed to the router object. It is then up to the router object to either handle the request itself and return a response or do some work and then call next() so that the next router object in the array can also get a chance at the request.

Using your CORS example if I were to first register my foo bar router that handles the OPTIONS method on all() and responds to it without calling next() then the CORS router would literally never get called.

In our case this matters because if, for example, the Express-PouchDB router gets registered before the ACL router then the request will be processed and responded to without the ACL code ever getting a chance to run.

That is why I keep stressing how important it is that the ACL middleware router object is registered before everything else.

---

*[thali/NextGeneration/thaliACLLayer.js, line 146 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGf07jCd7P983tQSVC:-KAGyE4TsgKxjSBZ_K8J:-2028794767) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L146)):*
So I'm being sleezy. I'm stating that for now we will only accept localhost connections over iPv4. If someone tries to connect over IPv6 the request will fail because it isn't from the right address. Less work. Less tests. :)

---

*[thali/NextGeneration/thaliACLLayer.js, line 148 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGgPUHrkhx2QcFJsAd:-KAGyLUGWO8uvX2xdQ3s:-2120627813) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L148)):*
And introduces a bunch of others because of how basic is handled in pouchdb-authentication. See previous comments. Again, if I misunderstood the code (a real possibility, I just stared at it, I didn't try it out) then I'm open to whatever is the fastest, easiest, cleanest, most maintainable route to our destination.

For example, an obvious approach would be if we are going to have two servers (one for non-tcp and one for wifi) why not just have three? We could have a third instance (meaning two on localhost) and configure that third instance to use basic auth only using admin mode. My only concern with that approach is that when I looked through the PouchDB code I noticed that they enforced their ACLs on most (but not all) endpoints not in the Express layer but rather as a wrapper around the PouchDB object inside of the Express server. This scares me. There just isn't any way you can compare that level of security to having a front door on the entire server with a simple 'yes'/'no'. Which is why I went for CLEAR. It also reduces the number of servers we have to manage. But nobody declared me right about everything. What do you think?

---

*[thali/NextGeneration/thaliACLLayer.js, line 199 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGhnSQNg5Apo46Xs5r:-KAH7pGqDWouhm6Z-bG7:56446787) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L199)):*
For new infra we explicitly only provide access to all pull replication interfaces on a single DB. That is it. In the future we will support multiple DBs but not today.

And using existing code to enforce the rules is double plus good.

---

*[thali/NextGeneration/thaliACLLayer.js, line 209 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGkpOJbN6dTA0nACpg:-KAH8EQJQoflGHdCjZDN:-774924850) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L209)):*
That's why I said that after res.end we would have to call request.connection.destroy. And for the tip, yup, we [already implemented it](https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext/thali/NextGeneration/makeIntoCloseAllServer.js)

---

*[thali/NextGeneration/thaliACLLayer.js, line 230 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGm2w-8LDqhNvUHsS5:-KAHIQSWrxldrJC1leum:-625692079) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L230)):*
I couldn't face the race condition (fixed)

---

*[thali/NextGeneration/thaliACLLayer.js, line 271 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGmUQP9Gn1dtE9hEDM:-KAHIVKnGIVD9KVoHW6t:1438921362) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L271)):*
I'm intentionally trying not to be generic because generic means more features with more testing. I'm following YAGNI here. So we only do exactly what we need for right now and no more. Let it be that people show up with interesting scenarios that need more, it will be great to expand.

---

*[thali/NextGeneration/thaliACLLayer.js, line 290 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-KAGmkC2cwlpb9Q00N83:-KAHJ6VkAwKSBwdOvubj:1619516239) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508658b0159fe77eb7206e87baa4c35a91576e3d/thali/NextGeneration/thaliACLLayer.js#L290)):*
Because we are explicitly not boiling any oceans. If and when we need a more generic layer we can build it on top of this one. There will always need to be a validatePskIdentity, it just may be that one day it's wrapped in a higher layer API that can handle other kinds of identity. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/520#-:-KAGqJ25hOp133C7i7XU:-1437759322)*

PR #530, opened 2016-02-10
Network change behavior for thaliWifiInfrastructure

vjrantal:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/530)

### DIFF
--- a/TeamReadMe.md
+++ b/TeamReadMe.md
@@ -139,3 +139,22 @@ Once built the library should be visible in:
 
 ## Developing node specific code and tests for the Thali Cordova Plugin
 Please see the Thali_CordovaPlugin/test/README.md
+
+## Debugging CI failures locally
+
+If you get a build or test error in CI that you didn't get locally, here
+are suggested steps that might help reproducing the issue:
+
+* To make sure you are testing the same code CI does, you need to first
+checkout the source branch of your PR and then merge the target branch onto it.
+  * If there is a merge conflict, that you need to resolve first and push the
+  resolution onto your source branch, because CI can't handle the conflict
+  by itself.
+* To make sure you don't have excessive or wrong dependency packages, the easiest
+way is to nuke all `node_modules` folders from your local source tree.
+  * If you are sure you have no uncommitted changes, you can do:
+  `git clean -fxd`.
+* To install dependencies and run tests the same way CI does, you should run
+`./build.sh` from the root of the repository.
+  * The script removes the folder `../ThaliTest` (relative to your clones souces)
+  so before running it, make sure you don't have any precious changes there.
--- a/test/TestServer/UnitTestFramework.js
+++ b/test/TestServer/UnitTestFramework.js
@@ -80,20 +80,24 @@ UnitTestFramework.prototype.startTests = function(platform, tests) {
 
     function emit(device, msg) {
       var retries = 10;
+      var emitTimeout = null;
 
       var acknowledged = false;
       device.socket.once(util.format("%s_ok", msg), function() {
-        acknowledged = true;      
+        acknowledged = true;
+        if (emitTimeout !== null) {
+          clearTimeout(emitTimeout);
+        }
       });
 
       // Emit message every second until acknowledged
       function _emit() {
         if (!acknowledged) {
-          device.socket.emit(msg);          
+          device.socket.emit(msg);
           if (--retries > 0) {
-            setTimeout(_emit, 1000);
+            emitTimeout = setTimeout(_emit, 1000);
           } else {
-            logger.error("test server: Device %s", device.name);
+            logger.error("test server: Device %s", device.deviceName);
           }
         }
       }
--- a/test/www/jxcore/bv_tests/testThaliMobile.js
+++ b/test/www/jxcore/bv_tests/testThaliMobile.js
@@ -42,6 +42,10 @@ var testIdempotentFunction = function (t, functionName) {
 
 var testFunctionBeforeStart = function (t, functionName) {
   ThaliMobile[functionName]()
+  .then(function () {
+    t.fail('call should not succeed');
+    t.end();
+  })
   .catch(function (error) {
     t.equal(error.message, 'Call Start!', 'specific error should be returned');
     t.end();
--- a/test/www/jxcore/bv_tests/testThaliMobile.js
+++ b/test/www/jxcore/bv_tests/testThaliMobile.js
@@ -102,7 +102,15 @@ test('a peer should be found after #startListeningForAdvertisements is called', 
     }, 'peer should have a non-empty host address');
     t.equals(typeof peer.suggestedTCPTimeout, 'number', 'peer should have suggested timeout');
     t.equals(typeof peer.portNumber, 'number', 'peer should have port number');
-    t.ok(peer.connectionTypes.length > 0, 'peer should have more than one connection type');
+    t.ok(peer.connectionType, 'peer should have a connection type');
+    var connectionTypeKey;
+    for (var key in ThaliMobile.connectionTypes) {
+      if (peer.connectionType === ThaliMobile.connectionTypes[key]) {
+        connectionTypeKey = key;
+      }
+    }
+    t.equals(peer.connectionType, ThaliMobile.connectionTypes[connectionTypeKey],
+             'connection type should match one of the pre-defined types');
 
     ThaliMobile.emitter.removeListener('peerAvailabilityChanged', peerAvailabilityChangedHandler);
     ThaliMobile.stopListeningForAdvertisements()
--- a/test/www/jxcore/bv_tests/testThaliMobile.js
+++ b/test/www/jxcore/bv_tests/testThaliMobile.js
@@ -1,26 +1,29 @@
 'use strict';
 
 var ThaliMobile = require('thali/NextGeneration/thaliMobile');
+var ThaliMobileNativeWrapper = require('thali/NextGeneration/thaliMobileNativeWrapper');
 var tape = require('../lib/thali-tape');
+var testUtils = require('../lib/testUtils.js');
 var express = require('express');
 var validations = require('thali/validations');
-
-var test = tape({
-  setup: function(t) {
-    t.end();
-  },
-  teardown: function(t) {
-    ThaliMobile.stop()
-    .then(function (combinedResult) {
-      t.end();
-    });
-  }
-});
 
 var verifyCombinedResultSuccess = function (t, combinedResult, message) {
   t.equal(combinedResult.wifiResult, null, message || 'error should be null');
   // TODO: Add nativeResult check once it is returned
 };
+
+var test = tape({
+  setup: function (t) {
+    t.end();
+  },
+  teardown: function (t) {
+    ThaliMobile.stop()
+    .then(function (combinedResult) {
+      verifyCombinedResultSuccess(t, combinedResult);
+      t.end();
+    });
+  }
+});
 
 var testIdempotentFunction = function (t, functionName) {
   ThaliMobile.start(express.Router())
@@ -85,15 +88,42 @@ if (!(typeof Mobile !== 'undefined' && Mobile.iAmAMock)) {
   return;
 }
 
-test('a peer should be found after #startListeningForAdvertisements is called', function (t) {
-  var peerAvailabilityChangedHandler = function (peers) {
+var setupDiscoveryAndFindPeer = function (t, callback) {
+  ThaliMobile.emitter.once('peerAvailabilityChanged', function (peers) {
     // Just pick the first peer from the list. In reality, it is possible that
     // if this test is run in environment with multiple Thali apps running, the
     // peer we get here isn't exactly the one with whom we are running these
-    // tests with. However, even with any peer, this test vefifies that we do get
-    // correctly formatted advertisements.
+    // tests with. However, even with any peer, this test vefifies that we do
+    // get correctly formatted advertisements.
     var peer = peers[0];
+    callback(peer, function () {
+      ThaliMobile.stopListeningForAdvertisements()
+      .then(function (combinedResult) {
+        verifyCombinedResultSuccess(t, combinedResult);
+        // On purpose not stopping advertising within the test
+        // because another device might still be running the test
+        // and waiting for advertisements. The stop happens in the
+        // test teardown phase.
+        t.end();
+      });
+    });
+  });
+  ThaliMobile.start(express.Router())
+  .then(function (combinedResult) {
+    verifyCombinedResultSuccess(t, combinedResult);
+    return ThaliMobile.startUpdateAdvertisingAndListening();
+  })
+  .then(function (combinedResult) {
+    verifyCombinedResultSuccess(t, combinedResult);
+    return ThaliMobile.startListeningForAdvertisements();
+  })
+  .then(function (combinedResult) {
+    verifyCombinedResultSuccess(t, combinedResult);
+  });
+};
 
+test('a peer should be found after #startListeningForAdvertisements is called', function (t) {
+  setupDiscoveryAndFindPeer(t, function (peer, done) {
     t.doesNotThrow(function () {
       validations.ensureNonNullOrEmptyString(peer.peerIdentifier);
     }, 'peer should have a non-empty identifier');
@@ -111,29 +141,38 @@ test('a peer should be found after #startListeningForAdvertisements is called', 
     }
     t.equals(peer.connectionType, ThaliMobile.connectionTypes[connectionTypeKey],
              'connection type should match one of the pre-defined types');
-
-    ThaliMobile.emitter.removeListener('peerAvailabilityChanged', peerAvailabilityChangedHandler);
-    ThaliMobile.stopListeningForAdvertisements()
-    .then(function (combinedResult) {
-      verifyCombinedResultSuccess(t, combinedResult);
-      // On purpose not stopping advertising within the test
-      // because another device might still be running the test
-      // and waiting for advertisements. The stop happens in the
-      // test teardown phase.
-      t.end();
-    });
-  };
-  ThaliMobile.emitter.on('peerAvailabilityChanged', peerAvailabilityChangedHandler);
-  ThaliMobile.start(express.Router())
-  .then(function (combinedResult) {
-    verifyCombinedResultSuccess(t, combinedResult);
-    return ThaliMobile.startUpdateAdvertisingAndListening();
-  })
-  .then(function (combinedResult) {
-    verifyCombinedResultSuccess(t, combinedResult);
-    return ThaliMobile.startListeningForAdvertisements();
-  })
-  .then(function (combinedResult) {
-    verifyCombinedResultSuccess(t, combinedResult);
+    done();
   });
 });
+
+/* This is commented out since there are timing issue why this doesn't always pass
+test('when network connection is lost a peer should be marked unavailable', function (t) {
+  setupDiscoveryAndFindPeer(t, function (peer, done) {
+    var availabilityChangedHandler = function (peers) {
+      var matchingPeer = null;
+      peers.forEach(function (peerCandidate) {
+        if (peerCandidate.peerIdentifier === peer.peerIdentifier) {
+          matchingPeer = peerCandidate;
+        }
+      });
+      if (!matchingPeer) {
+        return;
+      }
+      t.equals(matchingPeer.hostAddress, null, 'host address should be null');
+      t.equals(matchingPeer.portNumber, null, 'port number should be null');
+      ThaliMobileNativeWrapper.emitter.emit('networkChangedNonTCP',
+        testUtils.getMockNetworkStatus(true)
+      );
+      setTimeout(function () {
+        done();
+      }, 500);
+    };
+    ThaliMobile.emitter.on('peerAvailabilityChanged', availabilityChangedHandler);
+    setTimeout(function () {
+      ThaliMobileNativeWrapper.emitter.emit('networkChangedNonTCP',
+        testUtils.getMockNetworkStatus(false)
+      );
+    }, 500);
+  });
+});
+*/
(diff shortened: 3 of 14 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (c0fcaf3)

ThaliTester:
Test (Success) 58918757 build is completed (c0fcaf3)

See https://github.com/ThaliTester/TestResults/tree/58918757c0fcaf3_Network_change_behavior_for_thaliWifiInfrastructure_vjrantal/ for the logs

ThaliTester:
Test 58918757c0fcaf3(c0fcaf3) has failed

See https://github.com/ThaliTester/TestResults/tree/58918757c0fcaf3_Network_change_behavior_for_thaliWifiInfrastructure_vjrantal/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (1df074b)

vjrantal:
@yaronyg: You could start the code review on this one already, but before I merge, I'll still check why one of the on-device test results in CI was not what I expected..

ThaliTester:
Test (Success) 58918757 build is completed (1df074b)

See https://github.com/ThaliTester/TestResults/tree/589187571df074b_Network_change_behavior_for_thaliWifiInfrastructure_vjrantal/ for the logs

yaronyg:
Reviewed 10 of 11 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 3 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KAB_dVUs0oXWjPKnxz6:-KAB_dVUs0oXWjPKnxz7:268865458) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L3)):*
I would recommend sticking in a test where we start and then stop and then validate that our state is clean.

I would also stick in a test where we start and then stop and then start again and make sure we are still functionality.

---

*[test/www/jxcore/lib/testUtils.js, line 140 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KAB_wk9y_EkzFTcAHgL:-KAB_wk9y_EkzFTcAHgM:-2040321529) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/test/www/jxcore/lib/testUtils.js#L140)):*
Shouldn't this be called something like getMockWifiNetworkStatus since when we are mocking iOS or Android we'll want to return results claiming that bluetooth and BLE are available?

---

*[thali/NextGeneration/thaliMobileNative.js, line 444 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABThu0cXnS4qpI-zHW:-KABThu0cXnS4qpI-zHX:-174003704) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNative.js#L444)):*
Should we specify that {string} becomes {?string} and inside of saying "if missing" we should say "If null"?

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 247 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABSLFIO0JGJbAu-va1:-KABSLFIO0JGJbAu-va2:-683121749) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L247)):*
If someone calls getNonTCPNetworkStatus before they have called start then we should return an Error.

Otherwise if we don't have nonTCPNetworkStatus it just means we have a race condition where someone managed to call start and getNonTCPNetworkStatus before the response came back from the native layer. But in that case we need to create a new promise that we hand off to the thaliMobileNativeWrapper event handler to call when it gets an event. That will then tell us when the value is available and we can return it.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 422 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABTGb06P3435ogzz3s:-KABTGb15SDyvf8Sczf-:111966938) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L422)):*
I don't believe that our mock code should be in our production code. In other words I believe we should avoid a situation where iAmAMock shows up in a production file. Instead, I think, we should have mocked Mobile using proxyquire  for testing purposes and the thaliMobileNativeWrapper shouldn't need to know or care. Please talk to me if this is proving difficult, I'm sure together we can figure out a good approach.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 153 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABXCWxC7mshTkOESEW:-KABXCWxC7mshTkOESEX:-213262230) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L153)):*
Setting self.peerAvailabilities[key] = false actually creates a memory leak. Javascript has to remember every key for as long as the code is running so it knows its value is false. Another approach is to use the delete operator to remove the key all together. That way, no memory leak.

And yes, it is safe to use delete inside of a for in loop so long as you only call delete on the current key. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in for the gory details.

If we use delete then we can get rid of the 'if (self.peerAvailabilities[key])' check above.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 177 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABXhJ17y0SenhDBZKU:-KABXhJ26GMWEiHpwGl9:489767288) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L177)):*
I think using for in is confusing with an array. It's generally recommended that one instead use results.forEeach(). Eventually we'll switch to for...of but the version of v8 and Spidermonkey that we are using doesn't support ES6 yet.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 313 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABYbER6-x5aqy5uMP_:-KABYbESnSmOFG-Rk5GQ:-1837283324) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L313)):*
Shouldn't other parts of states be cleared out such as the current and desired target states? Sounds like a good thing to test for as well.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 388 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABZJ4VdvSGQcxcU95k:-KABZJ4VdvSGQcxcU95l:-1236009852) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L388)):*
Typically the public function shouldn't have internal arguments on it. I would suggest having a public function that takes no arguments and then a _stopListeningForAdvertisements version that has whatever internal test hooks we need.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 468 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABZr0XH2y7ZnAZDiut:-KABZr0XH2y7ZnAZDiuu:-1558430062) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L468)):*
This sets our goal state to true even if nobody has called start yet and we will eventually return an error. Isn't that a problem? If nothing else it puts the states variable in a strange state.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530)*

vjrantal:
Review status: 5 of 11 files reviewed at latest revision, 10 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js, line 3 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KAB_dVUs0oXWjPKnxz6:-KADqbSNcBvTKH5JmHDl:231024420) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/test/www/jxcore/bv_tests/testThaliWifiInfrastructure.js#L3)):*
Done.

Added an explicit test for validating state after stop and assertions to setup and teardown that gets run many times to ensure state is always right. Since we start and stop many times during the tests and all tests pass, I'd say the second paragraph gets tested with current tests.

---

*[test/www/jxcore/lib/testUtils.js, line 140 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KAB_wk9y_EkzFTcAHgL:-KADrmjlAhNZtqvv6gWg:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/test/www/jxcore/lib/testUtils.js#L140)):*
Done.

---

*[thali/NextGeneration/thaliMobileNative.js, line 444 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABThu0cXnS4qpI-zHW:-KADs56HZycHuVI9f5Kq:585920298) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNative.js#L444)):*
In the description of the property, there was already a defined behavior for when it is `null`. If we want to "overload" that and also cover the case where platform can't get the value with `null` I can change the text.

I think the question is, does anyone in the upper layers need to distinguish from this property whether we aren't connected to an access point versus whether we are unable to get the BSSID? I tend to think that separation might be valuable once someone implements the BSSID-based logic.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 247 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABSLFIO0JGJbAu-va1:-KADtpmmH2y1NVmsYUJZ:546949826) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L247)):*
Done.

This is true. The reason for this is that I implement only the bare minimum from `thaliMobilNativeWrapper.js` to be able to mock network events so I didn't have any state handling nor `start` function implemented.

I now implemented the basics for this to work like you say in above comment.

While implementing, I realized one scenario, which may result in behavior that isn't necessarily expected:
* The thaliWifiInfrastructure needs to call getNonTCPNetworkStatus to get the network  status and with this new requirement, it also needs to call thaliMobilNativeWrapper.start.
* To restore the objects in their default states after all is done, I am calling thaliMobilNativeWrapper.stop in thaliWifiInfrastructure.stop (otherwise, thaliMobilNativeWrapper would be in started state when thaliWifiInfrastructure tests are done)
* Above means that if someone stops thaliWifiInfrastructure, it also stops thaliMobilNativeWrapper (for listening and advertisement), which isn't necessarily what was wanted

This makes me wonder, should there a separate function in thaliMobilNativeWrapper called something like `init` that would be used to just "wakep up" the native side? thaliWifiInfrastructure could just call that and not have to call start/stop.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 422 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABTGb06P3435ogzz3s:-KAE01GZZIb6Z0jtpCIn:-1648181163) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L422)):*
I agree - this was just a quick way for me to get going.

I tried to use proxyquire, but that did not solve any of my issues. I guess the main reason for that is that here, we don't want to mock something that this file requires, but we want to mock a global object (which is something proxyquire doesn't do).

I think one approach is how I mocked the global Mobile object in `story_001` branch, but we can figure out better ways as well.

I suggest we let it be like this in this PR and cover the follow-up under https://github.com/thaliproject/Thali_CordovaPlugin/issues/531.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 153 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABXCWxC7mshTkOESEW:-KAE78CqkT74m2pgw3s7:-928499985) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L153)):*
Done.

This was a good catch and clearly had to be fixed.

Nitpick: I wouldn't necessarily call it a memory leak, but rather it was an array of peers that could have grown indefinitely :)

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 177 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABXhJ17y0SenhDBZKU:-KAEALo6ayZbMUB9RO9h:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L177)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 313 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABYbER6-x5aqy5uMP_:-KAEFUbPSyCfbzu8UfqV:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L313)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 388 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABZJ4VdvSGQcxcU95k:-KAEHpB0Ksu9TT59sHFb:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L388)):*
Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 468 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABZr0XH2y7ZnAZDiut:-KAEKay__IifWvUmM321:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L468)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530)*

vjrantal:
> @yaronyg: You could start the code review on this one already, but before I merge, I'll still check why one of the on-device test results in CI was not what I expected..

I was able to fix this issue so from my perspective, this PR would be ready to go in.

yaronyg:
Reviewed 9 of 9 files at r3.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 247 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABSLFIO0JGJbAu-va1:-KAGHG7zrLq1MiikTuN1:-568510765) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L247)):*
One thing I'm pretty sure of is that thaliWifiInfrastructure should never be calling start and stop on thaliMobileNativeWrapper. That is an abstraction failure. At worst thaliWifiInfrastructure should just detect that thaliMobileNativeWrapper isn't started and fail with a "hey, you got to start thaliMobileNativeWrapper first!"

But there is another approach that I think makes more sense. If you notice in the definition of start on thaliMobileNativeWrapper I explicitly omitted needing to call start in order to register for event handlers. This was to avoid some nasty race conditions. I only required that all other methods were blocked until start was called.

But what happens if we explicitly omit getNonTCPNetworkStatus from the start requirement? It is, after all, just syntactic sugar and not actually part of the native layer.

Remember, the native side is always 'awake'. Start doesn't wake it up. Start is just there to tell thaliMobileNativeWrapper when to set up its state. If you look into thaliMobileNative you won't find a start or stop method.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 422 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABTGb06P3435ogzz3s:-KAGJ_TY_W71CworChQF:-1675436562) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L422)):*
But mocking global objects is easy isn't it? Just replace the object in setup and reset it in tear down. That should work recursively through all the requires. Why not just do that from the start?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 153 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABXCWxC7mshTkOESEW:-KAGJusc-yqaZDvCgfb8:646649036) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliWifiInfrastructure.js#L153)):*
Isn't that the definition of a memory leak?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530)*

vjrantal:

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 247 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABSLFIO0JGJbAu-va1:-KAGNsZIrbzFgbsWsrdk:1421910370) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L247)):*
If I am not mistaken, what you suggest is what I had originally until you wrote as a review comment "If someone calls getNonTCPNetworkStatus before they have called start then we should return an Error." and then I implemented that way and run into this issue.

I'd prefer we don't require start before calling that and thus we don't have to call start/stop from thaliWifiInfrastructure.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 422 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABTGb06P3435ogzz3s:-KAGOXiGVRgInKzftrQz:-1831586769) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L422)):*
That is similar to what I had before and what I am doing in `story_001`. The workaround I required for this was https://gist.github.com/vjrantal/24231aa0df442ca1a668, because requiring a file to replace the global.Mobile with doesn't work if in the required file (or in the "sub-requires") there are calls to Mobile. Anyways, this is now a separate issue so suggesting to close this discussion from under this PR.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530)*

yaronyg:

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 247 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABSLFIO0JGJbAu-va1:-KAGTuZEpRK4uu_pdUK3:1225504478) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L247)):*
Not my first mistake, not my last. :)

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 422 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530#-KABTGb06P3435ogzz3s:-KAGUelFwSN6O0vKHtx2:1264083197) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df074bf9e04739fa5ca833311b651284c392ba7/thali/NextGeneration/thaliMobileNativeWrapper.js#L422)):*
I haven't tested it so I reserve the right to be full of crap but I think https://github.com/thlorenz/proxyquire#globally-override-require does exactly what we need.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530)*

vjrantal:
Reviewed 3 of 3 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/530)*

PR #540, opened 2016-02-16
Better desktop testing support with improved mocking

vjrantal:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/540)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -65,7 +65,7 @@ environment.
 
 ### Desktop
 
-All of the tests are designed to be run on the desktop. Tests that require the ThaliEmitter (our mobile environment)
+All of the test files are designed to be runnable also on the desktop. Tests that require the real mobile environment
 will de-activate themselves when run on the desktop. What is especially nice about running on the desktop is that
 one can develop and debug directly in the Thali_CordovaPlugin directory. There is no need to do the kind of 
 copying and pasting that Cordova development normally requires.
@@ -83,7 +83,8 @@ $ jx npm run test-meta
 $ jx npm run test-coordinated
 ```
 
-Some tests will also happily run stand alone so you can run a test directly (e.g. `jx bv_tests/testThaliCryptoManager.js`)
+Some test files will also happily run stand-alone so you can run a test directly
+(e.g. `jx runTests.js bv_tests/testTests.js`)
 thus allowing you to easily run and debug individual tests.
 
 ### Writing Unit Tests
--- a/test/www/jxcore/PerfTest_app.js
+++ b/test/www/jxcore/PerfTest_app.js
@@ -64,5 +64,4 @@ getDeviceCharacteristics(function(deviceName, bluetoothAddress) {
   });
 });
 
-// Log that the app.js file was loaded.
-console.log('Perf Test app loaded loaded');
+console.log('Perf Test app is loaded');
--- a/test/www/jxcore/UnitTest_app.js
+++ b/test/www/jxcore/UnitTest_app.js
@@ -9,7 +9,7 @@
 var testUtils = require('./lib/testUtils');
 
 if (typeof Mobile === 'undefined') {
-  global.Mobile = require('./lib/MobileUsingWifi.js');
+  global.Mobile = require('./lib/wifiBasedNativeMock.js')('iOS', null);
 }
 
 testUtils.toggleRadios(true);
--- a/test/www/jxcore/UnitTest_app.js
+++ b/test/www/jxcore/UnitTest_app.js
@@ -6,17 +6,22 @@
 
 'use strict';
 
-var testUtils = require('./lib/testUtils');
-
 if (typeof Mobile === 'undefined') {
-  global.Mobile = require('./lib/wifiBasedNativeMock.js')('iOS', null);
+  global.Mobile = require('./lib/wifiBasedNativeMock.js')();
 }
 
-testUtils.toggleRadios(true);
+var testUtils = require('./lib/testUtils');
 
-Mobile('GetDeviceName').callNative(function (name) {
-  console.log('My device name is: %s', name);
-  testUtils.setName(name);
-  require('./runTests.js');
-  console.log('Test app app.js loaded');
+testUtils.toggleRadios(true)
+.then(function () {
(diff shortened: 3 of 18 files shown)


yaronyg:
Good work and sorry for the delay. None of my comments are blocking so please resolve them to your own satisfaction and then :shipit: 

---

Reviewed 19 of 19 files at r1.

---

*[test/www/jxcore/lib/testUtils.js, line 35 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm0lFRT5ZbAwpjGFVG:-KAm0lFSABmbceAUSMLe:-272194130) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/test/www/jxcore/lib/testUtils.js#L35)):*
Why not make this work for mocked on desktop? Should be easy, right? Just track what you are pretending the state is.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 24 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm2NPS22BdMVAEtTC_:-KAm2NPS22BdMVAEtTCa:-1641281630) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/test/www/jxcore/lib/wifiBasedNativeMock.js#L24)):*
This assert will always pass so what is the point? I think you mean to check networkStatusCalled

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 195 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm3ORzlv2aNtyax-3M:-KAm3OS-jW9NnkdwUO4Q:-1123965071) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/test/www/jxcore/lib/wifiBasedNativeMock.js#L195)):*
Couldn't you just use pipe to set up data forwarding in both directions?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 192 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm1pRHg6SK5E2vcTCe:-KAm1pRHg6SK5E2vcTCf:-1899731777) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/thali/NextGeneration/thaliWifiInfrastructure.js#L192)):*
Rather than hard coding in test code into our production environment (which I admit sometimes is necessary) why not just reach in and manually override routerServerPort from the tests?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-:-KAm6UdZN5iYRHN8wZu6:-1199358379)*

vjrantal:
Reviewed 1 of 19 files at r1, 1 of 1 files at r2.

---

*[test/www/jxcore/lib/testUtils.js, line 35 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm0lFRT5ZbAwpjGFVG:-KAn3IWttmuSTSNg9F_u:1193725695) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/test/www/jxcore/lib/testUtils.js#L35)):*
Done.

This is already mocked on desktop.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 24 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm2NPS22BdMVAEtTC_:-KAn5XraVaL6iTnR9BBJ:-143373621) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/test/www/jxcore/lib/wifiBasedNativeMock.js#L24)):*
Done.

Thanks for catching that - a clear mistake from my side.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 195 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm3ORzlv2aNtyax-3M:-KAn5dQXIOM2av8ctIPY:-1284353485) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/test/www/jxcore/lib/wifiBasedNativeMock.js#L195)):*
Pipe would have worked here as well, but I used the more verbose way of wiring things, because I wanted to be prepared for the cases where we want to simulate certain socket connection errors or data timeouts etc. in the mock layer. I admit this is a bit pre-mature preparation since we aren't doing that kind of "advanced mocking" at the moment, but I'd like to leave it like this still.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 192 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540#-KAm1pRHg6SK5E2vcTCe:-KAn6sCc4X3ThnBRHDMd:1065997178) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/e2018a68f38cc88736cc07f716ee9b2c6467f065/thali/NextGeneration/thaliWifiInfrastructure.js#L192)):*
The advertised port override is a bit different thing than the router server port. When the Wifi code is used from the mock, the router server is still started, but it isn't used for anything. In the advertisement, we don't want to advertise the router server port, because that is not matching to where the other peer's mock is hosting a server than hosts the "proxy TCP server".

This is a bit of a hack to make the mock work, but I wanted to do it this way so that due to the mocking, I don't have to change the way the production Wifi code deals with the router server.

I have a better approach in mind that would be splitting the whole Wifi code into two parts (base and extension parts) and the mock would only use the base that doesn't do the things mock doesn't need. However, this would be some amount of extra work just for mocking sake so would not like to go there at this moment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/540)*

PR #543, opened 2016-02-17
Fix closing sockets when server closed

### LINKED ISSUE #542: makeIntoCloseAllServer does not close connections
Two main issues:
* The new connections are never added to the connections array and thus never iterated through during the `closeAll`phase
* Usage of Array.prototype.findIndex which is an ES6 feature that isn't available at least with all JavaScript engined JX supports

vjrantal:
Fixes #542.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/543)

### DIFF
--- a/thali/NextGeneration/makeIntoCloseAllServer.js
+++ b/thali/NextGeneration/makeIntoCloseAllServer.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var assert = require('assert');
+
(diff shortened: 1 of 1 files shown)


tobybrad:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/543)*

PR #544, opened 2016-02-17
didRegisterToNative specification and implementation on iOS

tobybrad:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/544)

### DIFF
--- a/src/ios/JXcoreExtension.m
+++ b/src/ios/JXcoreExtension.m
@@ -43,6 +43,8 @@
 - (void)appEnteringBackground;
 - (void)appEnteredForeground;
 
+- (void)didRegisterToNative:(NSString *)name;
+
 @end
 
 // JavaScript callbacks.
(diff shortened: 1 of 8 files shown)


vjrantal:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[test/www/jxcore/UnitTest_app.js, line 16 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/544#-KAk8H0_jI_XYfakQwZK:-KAk8H0_jI_XYfakQwZL:1224604215) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b774585f53dcb9de673e968f1a00d43f3e49fd38/test/www/jxcore/UnitTest_app.js#L16)):*
Since the monkey-patching is done only in the unit test app, I guess this would not yet solve the issue if thali would be used in stanton or postcard app. Or, from the performance test app.

I wonder how the solution could be expanded to cover those cases as well?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/544)*

yaronyg:
The situation would appear to be that if someone in Node.js land calls Mobile(x).registerToNative there is no notification given to the native code? So that when the native code calls (in the iOS case) JXcore callEventCallback there is no way for the iOS code to know that anyone in Node.js land every called registerToNative on that callback?

If that is the case then I'm happy with the change but you need to also go to thaliMobileNative.js and add in a new entry for Mobile('didRegisterToNative').callNative with an argument of the name of the registerToNative that was called. Then you also need to go to thaliMobileNativeWrapper and specify there that we MUST call didRegisterToNative whenever we have called a registerToNative at least once. 

---

Reviewed 4 of 4 files at r1.

---

*[test/www/jxcore/UnitTest_app.js, line 16 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/544#-KAk8H0_jI_XYfakQwZK:-KAlOS2_eSfL0PEF-seL:-1884671860) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b774585f53dcb9de673e968f1a00d43f3e49fd38/test/www/jxcore/UnitTest_app.js#L16)):*
See my comment above, if I'm following the bouncing ball correctly then this is a change we need formally (not just monkey patched) in thaliMobileNative and thaliMobileNativeWrapper.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/544#-:-KAlLI5okRjf-K2rASW_:1983032328)*

vjrantal:
Reviewed 4 of 5 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/544)*

Issue #533, opened 2016-02-11
[Android] Implement stubs for vNext interface in JXcoreExtensions.java 


tompaana:
See #532

tompaana:
Sub-task of #356

Issue #532, opened 2016-02-11
[Android] Write stubs for vNext interface to JXcoreExtensions.java


tompaana:
See:
* https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext/thali/NextGeneration/thaliMobileNative.js
* https://github.com/thaliproject/Thali_CordovaPlugin/blob/vNext/thali/NextGeneration/thaliMobileNativeWrapper.js

tompaana:
Sub-task of #356

PR #548, opened 2016-02-18
Pull request for issues #532 and #533

tompaana:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/548)


yaronyg:
This is a massive piece of work. Thank you! It's quite awesome.

StreamCopyingThread.setBufferSize - This function is, I believe, confusing because it looks like it should support dynamically changing the buffer size but in fact it doesn't. Once run is called you can call this method to your heart's desire and nothing is going to happen. I would suggest that we just make setBufferSize into an argument on the constructor. If you really want to make it optional the we can have a constructor overload without it that uses a default value.

StreamCopyingThread - I would seriously consider having the uncaught exception handler passed in with the constructor. This object can throw and missing it would be bad so forcing the developer to set the handler by passing it into the constructor is just good defensive coding. In other words, rather than having to call setUncaughtExceptionHandler on the StreamCopyingThread object externally, just have the constructor include the handler and then set it on self inside of StreamCopyingThread.

StreamCopyingThread.closeLocalSocketAndStreams & closeBluetoothSocketAndStreams - I was quite surprised to not see the various values that were closed then set to null. Doing so would have the nice side effect of making sure that any subsequent call to startStreamCopyingThreads would fail.

ConnectivityChangeListener - Can't we delete this file? I'm pretty sure this thing is seriously out of date.

---

Reviewed 15 of 15 files at r2.
Review status: all files reviewed at latest revision, 35 unresolved discussions.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 40 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqHMD5pSCyQHXPrUeF:-KAqHMD61ML9qhyDaf3v:-618687404) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L40)):*
I thought we agreed that in the Thali code we wouldn't use a peer name since Thali_CordovaPlugin_Bluetooth doesn't need it anymore (e.g. it's optional)? Also advertising our device manufacturer and model is not ideal. It's leaking information that can potentially help an attacker.

This however may be the point where you tell me we need it for debugging and it never goes on the wire but I thought we used the Bluetooth name for that and just made sure to set it to a unique value on each phone? I don't remember.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 43 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqHljZIN8PHzlavUZt:-KAqHljZIN8PHzlavUZu:2042072247) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L43)):*
15 Seconds seems like a really long time. Why was such a long interval necessary? Just curious.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 44 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqHtYAhczI2HDNX2el:-KAqHtYAhczI2HDNX2em:919230279) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L44)):*
Can you please put the bug number tracking this issue into the comment so we can be sure not to forget to eventually do something about this?

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 71 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqINtoT3Z34mCl1n_L:-KAqINtoT3Z34mCl1n_M:-1377815915) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L71)):*
Who would actually end up catching this?

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 94 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqJ2pE6bu1JeH9Z33L:-KAqJ2pE6bu1JeH9Z33M:1711417009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L94)):*
That this option is even available seriously concerns me. We are creating code paths in our main line code that we don't support, don't test and in fact know doesn't work. I was border line o.k. with keeping this code in Thali_CordovaPlugin_Bluetooth but I don't think it should be in Thali_CordovaPlugin. Only mainline, supported code should be here. So if we don't support BLE on the device then I think we need to just error out.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 113 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqK3jIlF47xmRrYN7K:-KAqK3jIlF47xmRrYN7L:1151465925) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L113)):*
I'm going to guess that the negative value is intended for testing purposes. If that is the case can you make a note of that? If it's intended for production purposes could you put a note of when we would do that? It wasn't obvious to me.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 120 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqKSwT2B8JarNLo8uJ:-KAqKSwT2B8JarNLo8uK:530188421) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L120)):*
0 is not a legal TCP port value

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 123 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqKLqSBOidjx9tOMyK:-KAqKLqSBOidjx9tOMyL:-2012454194) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L123)):*
mServerPortNumber is initialized to 0. So if a negative value is specified then mServerPortNumber stays at 0. So what have we accomplished? At a minimum won't it cause us to send garbage to Node.js since we will specify our ports as having the value 0 which is illegal? Sorry, this scenario is confusing me. Please explain.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 310 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqDw_Twjegkl_RlSzH:-KAqDw_Twjegkl_RlSzI:1551037445) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L310)):*
Shouldn't you check if the result was true or false?

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 394 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqReujydM0JXI0ltGk:-KAqReuk2fqICaeKIb1_:1206064793) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L394)):*
Why does this call have to be made on the main thread?

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 400 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqPEmxKqdzUz2uReHA:-KAqPEmxKqdzUz2uReHB:-926242298) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L400)):*
This value cannot ever be null because it is equal to who and if who was null then the attempt to set threadId above would have thrown a null pointer exception.

I think what you meant to do was check if errorMessage wasn't null but in that case shouldn't you check that the value is what you expect so you can make sure that the problem was that we couldn't connect to the internal port and not something else? Looking at the code there are more than just one reason why this callback could get an error.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 401 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqkNMxKXQIVt8dVseT:-KAqkNMyI5RE5lFqLIli:-1033636838) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L401)):*
There is a requirement that we rate limit how often we call this method to no more than once every 100 ms. Where is that throttling? I looked but I didn't see it here or in JXcoreExtension

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 535 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqTO6d4UxLNc6CvqKr:-KAqTO6d4UxLNc6CvqKs:1374064063) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L535)):*
Shouldn't we log the errorMessage even if peerProperties are null?

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 559 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqU5YMRg9Yd8AB8ead:-KAqU5YMRg9Yd8AB8eae:-582586885) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L559)):*
Can you please put in the bug number for the bug to resolve this? It's easy for us to forget since we don't use Marshmallow in CI yet.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 623 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqVFwcSAIEJ0g-V6vG:-KAqVFwd9hCt2tPam7HZ:-224649601) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L623)):*
Shouldn't this also cause a notifyPeerAvailabilityChanged event just like onPeerDiscovered?

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 646 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqV_lVVbxEAKTUiBxM:-KAqV_lVVbxEAKTUiBxN:-1569315114) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L646)):*
I think that this override as well as the next two should throw an exception since they should only be called when we are in bro mode and right now we aren't supporting bro mode. So their use is de-facto an error.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 673 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqW3mh-Iu7MoRaeb_J:-KAqW3mh-Iu7MoRaeb_K:1038625359) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L673)):*
If we aren't in BLE discovery mode isn't that an error? I know I'm being a pain about this but I really want the Thali_CordovaPlugin code to be very clean and only do exactly what we said.

---

*[src/android/java/io/jxcore/node/ConnectionModel.java, line 16 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqCEya7mLeD3uh_16a:-KAqCEybhWZbHK9BZQdK:-853284766) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionModel.java#L16)):*
Um....?

---

*[src/android/java/io/jxcore/node/ConnectionModel.java, line 149 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqF4HGPjc7zDeTuyHK:-KAqF4HGPjc7zDeTuyHL:-1930035200) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionModel.java#L149)):*
This call is giving me minor nightmares. Would it ever make sense for us to add the same incomingSocketThread twice? Because if it doesn't then we should be using a CopyOnWriteArraySet and not a CopyOnWriteArrayList in order to guarantee that we never add the same value more than once.

---

*[src/android/java/io/jxcore/node/ConnectivityInfo.java, line 66 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApibZ5Z9t9iHLMXHN7:-KApibZ5Z9t9iHLMXHN8:-1625905868) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectivityInfo.java#L66)):*
It seems like a bad idea to just swallow the exception. We depend on this class working for lots of stuff and with this exception the class won't work.

---

*[src/android/java/io/jxcore/node/IncomingSocketThread.java, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApfUcX60w5Gw-VbyTB:-KApfUcX60w5Gw-VbyTC:849060819) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/IncomingSocketThread.java#L18)):*
There is actually no reason that this port has to be for a HTTP server. In fact, it's not. It's going to be a generic TCP listener that runs our MUX infrastructure. I would suggest naming this and the associated methods something more generic like mTcpPort.

---

*[src/android/java/io/jxcore/node/IncomingSocketThread.java, line 41 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApgZJ94U0H9CHyy5Xn:-KApgZJ94U0H9CHyy5Xo:-1317273876) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/IncomingSocketThread.java#L41)):*
While it is true that there isn't any port 0 wouldn't a negative number make more sense since in essence we are saying that this function was called at a bad time?

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 91 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqZIHusO1YuLfGQO1d:-KAqZIHusO1YuLfGQO1e:315595118) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L91)):*
I'm a little concerned with the actual text from thaliMobileNative.js being here. The issue is - what happens when we update thaliMobileNative.js? We do that all the time as we find bugs. There have been 5 or so changes since the start of the year and another should be coming soon from Toby (you missed that in today's stand up while you and Juhana was off having fun! =).

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 163 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAq__sxnUtQx8aQC6wX:-KAq__syqJNsD8NzyK4H:-456777782) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L163)):*
This is my other concern about copying in the text. The text above is specifically about the start and stop methods in thaliMobileNativeWrapper, they don't apply to this code. It's kind of confusing for the reader.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 218 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqacFknrvEoP3u23D0:-KAqacFknrvEoP3u23D1:-209374105) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L218)):*
Shouldn't we make sure that the portNumber passes the giggle test? Also what happens if someone passes in something that isn't an Integer?

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 219 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqbHZBUzTdHaCr4k9s:-KAqbHZBUzTdHaCr4k9t:1652748314) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L219)):*
Am I hallucinating or are lines 219 through 243 identical to lines 99 through 121 with the exception of line 241? If so, could you please factor that code out into a re-usable function?

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 379 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqcF7-mNIhT_xn8_EM:-KAqcF7-mNIhT_xn8_EN:836807670) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L379)):*
blah blah blah only BLE blah blah blah (yes, even I get sick of myself)

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 403 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqcdnAPF8DK64TwmjV:-KAqcdnAPF8DK64TwmjW:-1033683831) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L403)):*
We actually had not intended to distinguish between these two states (e.g. connected vs connecting) since testing that is a timing nightmare from hell and the distinction isn't needed higher up in the stack. So feel free to simplify your life and just return "Already connect(ing/ed)" in both cases.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 476 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqdgUqCgvCQmNeMVCF:-KAqdgUqCgvCQmNeMVCG:452362913) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L476)):*
That's the problem with devs these days. They just can't follow directions! The spec clearly states that Android is not supposed to do this! Sigh... devs.... I'll go fix thaliMobileNativeWrapper.js to specify that Android does support this.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 496 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqeY_R9OgLqboCGf5p:-KAqeY_R9OgLqboCGf5q:-54956022) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L496)):*
We shouldn't need this for anything since we implement disconnect by killing the connection to the native TCP/IP listener. So can we remove this code? (One less thing to test) I know this breaks some tests but those tests aren't going to work with iOS anyway so isn't it time we killed them?

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 516 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqiJMrFWGROQyB3f7j:-KAqfvBfhzNH_mirS4HP:-2005891882) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L516)):*
Why do we need this? Isn't this information already in the network connections status update notification? In other words wouldn't we get back a "BLE not supported" result from that notification?

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 670 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqjc9U0KDGk2Vrl1M9:-KAqjc9U0KDGk2Vrl1MA:12146322) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L670)):*
You have the right value for this property. It should only be true in certain specific cases involving iOS. Android should always return false.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 841 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAql7pPGKEaPIbcGjHN:-KAql7pPGKEaPIbcGjHO:-1338618809) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L841)):*
Isn't this from the old spec? I don't think we support this anymore. If we can't connect to a remote peer that should show up as a peerAvailabilityChanged event.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 858 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqlN5QumguwXpoNCK8:-KAqlN5QumguwXpoNCK9:-1126818059) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L858)):*
Grumble... grumble... ble.... grumble... grumble.... If BLE isn't working and if Bluetooth isn't work then we aren't working.

---

*[src/android/java/io/jxcore/node/OutgoingSocketThread.java, line 21 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApmcshEtpkeb0QFJlO:-KApmcshEtpkeb0QFJlP:1576948609) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/OutgoingSocketThread.java#L21)):*
This delay should not be necessary. See http://stackoverflow.com/questions/2409277/can-connect-call-on-socket-return-sucessfully-without-server-calling-accept but my understanding is that as soon as we call new ServerSocket(0) and get a response then the port is bound and will accept incoming connections. But those connections will just be queued until a call is made to accept. So you should be able to call listener.onListeningForIncomingConnections() as soon as you initialize mServerSocket.

The reason I care is that 300 ms is a long time to wait. Humans can perceive anything above around 100 ms and get annoyed starting at 250 ms. So if we don't need a delay we sure as heck don't want to add it and I don't believe we need this delay.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-:-KApXwi2AqfhNLDiGhXO:1643540834)*

tompaana:
Review status: 3 of 19 files reviewed at latest revision, 35 unresolved discussions.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 43 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqHljZIN8PHzlavUZt:-KB87pM4795dWyM4pp5E:1690291792) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L43)):*
We've seen data transfer sometimes pause for several seconds. It can be fatal for the data transfer, if we power up the discovery while still ongoing. 

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 44 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqHtYAhczI2HDNX2el:-KB889V2sSXjfPuxqKPS:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L44)):*
Done.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 71 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqINtoT3Z34mCl1n_L:-KB88GZmgzHXegWzRNHt:1358094681) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L71)):*
Um.. I don't know.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 94 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqJ2pE6bu1JeH9Z33L:-KB88M7Dnly_tNZnL6z-:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L94)):*
Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 113 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqK3jIlF47xmRrYN7K:-KB88NsMyhXuWNBy0oJp:2138798012) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L113)):*
Improved. startListeningForAdvertisements does not want to set the port.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 123 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqKLqSBOidjx9tOMyK:-KB88_Y-SFwuvxi_AZJA:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L123)):*
Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 310 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqDw_Twjegkl_RlSzH:-KB88cTdvUWqZ0r9arEJ:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L310)):*
Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 394 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqReujydM0JXI0ltGk:-KB88e9Az29yxvlGgF_Z:1296404362) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L394)):*
No reason. Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 400 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqPEmxKqdzUz2uReHA:-KB88hSrcYLDIOHw_-oe:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L400)):*
Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 401 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqkNMxKXQIVt8dVseT:-KB88jPgG_3fTJev-1EM:432228903) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L401)):*
Fixed (in JXcoreExtension.java)

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 535 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqTO6d4UxLNc6CvqKr:-KB89HqKzdvT_1ssf7fb:-1546924208) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L535)):*
Yes, fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 559 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqU5YMRg9Yd8AB8ead:-KB89K2nDYyGjHjUxI-u:1104459294) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L559)):*
This callback is not applicable since we have ThaliPermissions.java. Made to throw an exception if called.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 623 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqVFwcSAIEJ0g-V6vG:-KB7tzuf9Pv4Jdj8wGzi:-1072157893) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L623)):*
The availability does not change.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 646 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqV_lVVbxEAKTUiBxM:-KB89VqkHqum3AJaOTE_:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L646)):*
Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 673 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqW3mh-Iu7MoRaeb_J:-KB89XbYGk4RPRdF99aH:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L673)):*
Fixed.

---

*[src/android/java/io/jxcore/node/ConnectionModel.java, line 16 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqCEya7mLeD3uh_16a:-KB89Z1VYa-kNMYPz4Ks:1740281843) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionModel.java#L16)):*
Removed.

---

*[src/android/java/io/jxcore/node/ConnectionModel.java, line 149 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqF4HGPjc7zDeTuyHK:-KB8-AC2wWw4zFuOTuSm:-1473052910) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionModel.java#L149)):*
Implemented a check to make sure a thread for the same peer isn't added twice.

---

*[src/android/java/io/jxcore/node/ConnectivityInfo.java, line 66 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApibZ5Z9t9iHLMXHN7:-KB8-L6AVHAXObR4NmbQ:1035749911) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectivityInfo.java#L66)):*
Will now return boolean and ConnectionHelper.start() will also return false in case of failure.

---

*[src/android/java/io/jxcore/node/IncomingSocketThread.java, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApfUcX60w5Gw-VbyTB:-KB8-UdqDQttkrMXUD1P:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/IncomingSocketThread.java#L18)):*
Fixed.

---

*[src/android/java/io/jxcore/node/IncomingSocketThread.java, line 41 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApgZJ94U0H9CHyy5Xn:-KB8-pCR21obGKWh3IHz:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/IncomingSocketThread.java#L41)):*
Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 91 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqZIHusO1YuLfGQO1d:-KB85ArJn1ZkT5qWQXR_:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L91)):*
Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 163 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAq__sxnUtQx8aQC6wX:-KB85BwByrCoY5TZFUEA:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L163)):*
Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 218 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqacFknrvEoP3u23D0:-KB85CpiQ_syyAG_bCOR:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L218)):*
Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 219 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqbHZBUzTdHaCr4k9s:-KB85E7j1YU6jB3iQY22:631373240) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L219)):*
No, you are not. Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 379 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqcF7-mNIhT_xn8_EM:-KB85JkrHmwFF0-TOyMf:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L379)):*
Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 403 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqcdnAPF8DK64TwmjV:-KB85Mwb1SMriLiq5upJ:677654516) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L403)):*
fx'd

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 476 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqdgUqCgvCQmNeMVCF:-KB85Rfk1w_N5uc-s3dY:-1969640475) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L476)):*
Must've been the case "didn't read instructions since the method has always been there". Fixed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 496 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqeY_R9OgLqboCGf5p:-KB86RSxZASZ2vAABK40:1740281843) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L496)):*
Removed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 516 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqiJMrFWGROQyB3f7j:-KB86dg46F0PFvAWQxvn:-152854574) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L516)):*
I think Ville wanted us to have this so he can easily (or somehow easier) detect which devices in CI have no BLE support.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 841 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAql7pPGKEaPIbcGjHN:-KB877JwNWQwZRbdVbJi:-1891592773) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L841)):*
Really? Random connection error -> peerAvailabilityChanged (false)?

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 858 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqlN5QumguwXpoNCK8:-KB87H8G_0j2PyJc-33F:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L858)):*
Fixed.

---

*[src/android/java/io/jxcore/node/OutgoingSocketThread.java, line 21 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KApmcshEtpkeb0QFJlO:-KB87R5mTM6fMhgJw76G:579023722) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/OutgoingSocketThread.java#L21)):*
Jukka was certain, this was needed. But I'll remove and test.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548)*

yaronyg:
Please resolve the remaining 4 comments as you see fit and then :shipit: without waiting for further review. Thanks!

---

Reviewed 1 of 15 files at r2, 3 of 11 files at r3, 13 of 13 files at r4.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 71 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqINtoT3Z34mCl1n_L:-KB9NZPRJUgC34THDFQO:43870172) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L71)):*
Um...please find out. We need to make sure we fail in a way that something reasonable happens like we log the hell out of it so we can debug when a customer comes screaming.

---

*[src/android/java/io/jxcore/node/ConnectionModel.java, line 149 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqF4HGPjc7zDeTuyHK:-KB9PccKfjhJr8hDKCw7:1523137078) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionModel.java#L149)):*
Because the method is synchronized and if you are super careful on mIncomingSocketThreads then this is safe. But I would suggest, just from the perspective of programming robustness, that you instead use addIfAbsent(). That guarantees thread safe functionality.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 476 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqdgUqCgvCQmNeMVCF:-KB9SATyNA0FeP-kMHLT:-1456034307) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L476)):*
No!!!! I already fixed the spec to say that Android DOES support this!!!! Put the code back! :)

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 841 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAql7pPGKEaPIbcGjHN:-KB9SORRu_5bnu4Y21_s:-1283377366) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L841)):*
Sorry I was thinking more along the lines of making a connect call and either we fail right off and therefore return the asynch callback for connect with an error or if it fails for some reason later (which Bluetooth in particular loves to do) then that would manifest by the native layer killing the local TCP/IP client that is connected to the Mux layer. This is automatically detected by the Mux layer who will then retry the connect.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-:-KB9Ut9DXOWJgBF8j6ti:893970714)*

tompaana:
Review status: 16 of 19 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/android/java/io/jxcore/node/ConnectionHelper.java, line 71 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqINtoT3Z34mCl1n_L:-KBDASkjx6UTvk6Kfk9m:1529070960) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionHelper.java#L71)):*
The answer is: no-one. From http://developer.android.com/reference/java/lang/Thread.UncaughtExceptionHandler.html: "Further exceptions thrown in this method prevent the remainder of the method from executing, but are otherwise ignored."

But we do log the error and it shows neatly in the blogs. I tested it. And the docs are correct: Throwing an exception did nothing as it was the last statement in the method.

---

*[src/android/java/io/jxcore/node/ConnectionModel.java, line 149 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqF4HGPjc7zDeTuyHK:-KBDE43Afut-P23owd3W:-1218077779) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/ConnectionModel.java#L149)):*
Switched to using addIfAbsent() and implemented SocketThreadBase.equals() to compare the peer IDs.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 476 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548#-KAqdgUqCgvCQmNeMVCF:-KBDFQgVRxx-0sYxGLnw:1094638715) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1e5184810f0750b7dc6f2be0d68d65af620bd8f7/src/android/java/io/jxcore/node/JXcoreExtension.java#L476)):*
Careful - I went and actually checked the recent commits to make sure you weren't joking.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548)*

vjrantal:
Reviewed 1 of 1 files at r8, 1 of 1 files at r9.
Review status: 14 of 20 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/548)*

PR #557, opened 2016-02-19
Avoid writes to stderr during test run

vjrantal:
This commit was done so that in a success test run, we would not write to
stderr, because in some environments, such output is considered an error
behavior and fails the test.

The PouchDB.defaults call was moved so that it gets called only once, because
otherwise, event emitters gets subscribed to many time which caused a warning
like `EventEmitter memory leak detected` that was written to stderr.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/557)


vjrantal:
Build still failed so stderr writes weren't the issue here.

PR #560, opened 2016-02-19
Fix #539, get rid of magic values and increase perf

### LINKED ISSUE #539: Test in testThaliNotificationBeacons.js sometimes fails
To reproduce:
```
$ git checkout vNext
$ ./thali/install/setUpDesktop.sh
$ cd test/www/jxcore
$ jx bv_tests/testThaliNotificationBeacons.js
```
The issue happens only rarely, but when it does, there is what it looks like:
```
# setup
# #parseBeacons addressBookCallback returns public key
not ok 19 should be equal
  ---
    operator: equal
    expected: 1
    actual:   2
    at: Test.<anonymous> (/Users/vjrantal/Software/Thali/Thali_CordovaPlugin/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js:404:5)
  ...
```
The easiest way to reproduce that I have found is just to loop the test as long as the error occurs. This can be done with a bash script like:
```
#!/bin/bash

set -e

COUNTER=0
while [  $COUNTER -lt 1000 ]; do
    jx bv_tests/testThaliNotificationBeacons.js
    let COUNTER=COUNTER+1
done
```

yaronyg:
We had a spurious match during decryption that would occasionally cause
our counts to be off.

testThaliNotificationBeacons - Removed or at least better explained
magic values

thaliNotificationBeacons - Got rid of magic values and also pulled
calculating sey out of parse beacons so we only perform the ECDH
operation once per beacon string instead of once per beacon! Did the same
for the HKDF hkey value.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/560)

### DIFF
--- a/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js
+++ b/test/www/jxcore/bv_tests/testThaliNotificationBeacons.js
@@ -114,20 +114,24 @@ test('#generatePreambleAndBeacons multiple keys to notify', function (t) {
     expiration
   );
 
-  var pubKe = results.slice(0, 65);
-  var expirationBuffer = results.slice(65, 65 + 8);
+  var pubKe = results.slice(0, notificationBeacons.PUBLIC_KEY_SIZE);
+  var expirationBuffer = results.slice(notificationBeacons.PUBLIC_KEY_SIZE,
+    notificationBeacons.PUBLIC_KEY_SIZE + notificationBeacons.LONG_SIZE);
 
-  t.equal(pubKe.length, 65);
-  t.equal(expirationBuffer.length, 8);
+  t.equal(pubKe.length, notificationBeacons.PUBLIC_KEY_SIZE);
+  t.equal(expirationBuffer.length, notificationBeacons.LONG_SIZE);
   var expirationDate =
     long.fromBits(expirationBuffer.readInt32BE(4),
                   expirationBuffer.readInt32BE(0))
       .toNumber();
   var outputExpiration = expirationDate - oldNow;
-  var errorRange = 250 + 100; // 250 is the maximum fuzz and 100 just incase
+  // The 100 is just a buffer to deal with slow environments, I had to pick
+  // some random value.
+  var errorRange = notificationBeacons.EXPIRATION_FUZZ_MAX_VALUE + 100;
   t.ok(outputExpiration <= expiration + errorRange &&
        outputExpiration >= expiration - errorRange);
-  t.equal(results.length, 65 + 8 + 48 + 48 + 48);
+  t.equal(results.length, notificationBeacons.PUBLIC_KEY_SIZE +
+    notificationBeacons.LONG_SIZE + (notificationBeacons.BEACON_SIZE * 3));
 
   t.end();
 });
@@ -137,7 +141,8 @@ test('#parseBeacons invalid ECDH public key in beaconStreamWithPreAmble',
     var localDevice = crypto.createECDH(notificationBeacons.SECP256K1);
     localDevice.generateKeys();
 
-    var beaconStreamWithPreAmble = new Buffer(62);
+    var beaconStreamWithPreAmble =
+      new Buffer(notificationBeacons.PUBLIC_KEY_SIZE - 1);
 
     var addressBookCallback = function () {
       t.fail();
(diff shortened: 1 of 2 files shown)


cicorias:
Reviewed 2 of 2 files at r1.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 305 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560#-KAvJVlk8kGTx873lPyo:-KAvJVllgSuNIjhH3jsZ:292297544) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/41263e197898e565407538037bd385b5f04c86a7/thali/NextGeneration/thaliNotificationBeacons.js#L305)):*
is this intentional?  if at this point, wouldn't it all fail from here forward on the continue? also, which failed ,the createDecipher, or buffer.concat.?  

should something happen with e.message?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560)*

yaronyg:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 305 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560#-KAvJVlk8kGTx873lPyo:-KAvKafs46G_3dXiVq1o:-357466210) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/41263e197898e565407538037bd385b5f04c86a7/thali/NextGeneration/thaliNotificationBeacons.js#L305)):*
Yes, the continue is expected. If the decrypt fails that means that the beacon we are examining isn't for us and we move on. The more awful case is that because JXcore doesn't support GCM it's possible to get a false positive where it looks like decryption did work when it didn't. But we are hardened against tat.

The real reason I don't check the exception is because exception handling in Javascript is a disaster. Pretty much nobody (including us) bothers to do anything even remotely sensible. For example, Node.js's decrypt API doesn't return an error object of some sub-type so we could detect by type what it is. All you can do is look at the exception value and if you look at the string you will see that it is so specific that any change in Node will cause the check to fail. :( But it's the best we can do for the moment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560)*

cicorias:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 305 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560#-KAvJVlk8kGTx873lPyo:-KB8GQHtZpc3FOGU_hOl:417842772) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/41263e197898e565407538037bd385b5f04c86a7/thali/NextGeneration/thaliNotificationBeacons.js#L305)):*
OK - and there was that "continue" - is that supposed to continue the for loop of perhaps a break out of it?

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 309 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560#-KB8GawJU7mrJDyRjuVF:-KB8GawJU7mrJDyRjuVG:2020542818) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/cc4a0ad1fae9f051a9433f0b974368c1f63fc130/thali/NextGeneration/thaliNotificationBeacons.js#L309)):*
should this be a break or continue? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560)*

cicorias:
Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560)*

yaronyg:

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 305 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560#-KAvJVlk8kGTx873lPyo:-KB9MN7rk6t18Vxar2Xa:114579362) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/41263e197898e565407538037bd385b5f04c86a7/thali/NextGeneration/thaliNotificationBeacons.js#L305)):*
It's continue because if we do get a decryption error that just tells us that this specific token isn't for us. It doesn't tell us anything about the next token. So the best we can do is a continue so we can see that next token.

---

*[thali/NextGeneration/thaliNotificationBeacons.js, line 309 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560#-KB8GawJU7mrJDyRjuVF:-KB9MuMAtGRTU1Ju9-tS:1116083657) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/cc4a0ad1fae9f051a9433f0b974368c1f63fc130/thali/NextGeneration/thaliNotificationBeacons.js#L309)):*
See previous

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/560)*

PR #561, opened 2016-02-22
tcpServersManager and tests

tobybrad:
- [X] Figure out the maximum call stack size exceeded crashing bug
- [x] Review createNativeListener to find all the places where the CBs make bad assumptions
- [x] Punch up the createNativeListener tests to test the weak points where things can fail
- [x] Add test that we properly close incoming connections when they time out
- [x] Fix incomingConnectionState events aren't emitting correct values
- [ ] Fix Appveyor failure.
- [ ] Check to see if we are using any other native methods and if so that they are expecting strings
- [ ] Check if routerPortConnectionFailed is emitted correctly
- [ ] Check if failedconnection is emitted correctly
- [ ] Fix the tests that are expecting the wrong events
- [ ] Implement terminateIncomingConnection
- [ ] Test implementation of terminateIncomingConnection
- [ ] Check all the cbs in the forward connection code for bad assumptions, also check the pipes for possible closed objects
- [ ] Check forward connection weak points to make sure we clean up properly
- [ ] Check the code that creates a listener in response to a peer availability event and make sure that at some point we clean port up.
- [ ] Figure out how to make sure that the tests run on devices, we don't have to run cross device but we do want the tests running there. Ville suggested we could detect when we are on mobile and run. But I'd like to get rid of the fixed ports and let it run in coordinated as well.
- [ ] Figure out if the start interface on tcpServersManager really needs to return the local port it is listening on for connections from the native interface. Does or should anyone but tcpServersManager care about that value?

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/561)


yaronyg:
@tobybrad Appveyor says that tests aren't passing. Please resolve. If you click on the appveyor link you'll see that the failures are relevant to this code.

Also, the Javascript code was written without running JSCS or JSHINT so it contains a lot of style violations. I'll fix them later but in the future please use an IDE that enforces JSCS and JSHINT rules.

---

Reviewed 3 of 5 files at r1.
Review status: 3 of 5 files reviewed at latest revision, 23 unresolved discussions, some commit checks failed.

---

*[thali/NextGeneration/tcpServersManager.js, line 11 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA1HhMBAxfaClqSlu-:-KBA1HhMBAxfaClqSlu0:1695645900) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L11)):*
Why don't you use thalilogger?

---

*[thali/NextGeneration/tcpServersManager.js, line 158 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KB9bXjiX4l8TBp4s016:-KB9bXjjXo8T9j82Vwb6:2141095549) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L158)):*
It's true that if you tried to add an object directly as a property member then you would have a problem because it's name has to be represented as a string and the issue you bring up causes a problem.

But you should never have to assign an ID yourself because Javascript is required to be able to distinguish different instances of the same object. Please see https://gist.github.com/yaronyg/dea38f150015ab62c5c7

So please remove this method and the id properties in general, we don't need them. We can instead just use '===' comparisons to determine membership.

(As a side note, if we were going to use _id then wouldn't it have made more sense to just put the IDs as properties on objects like mux._streams and that way we could at least do lookups in O(1) and removals (using delete) in O(1)? But IDs are like ref counting, it's just something to get wrong so let's keep the structure of your existing code using an array and get rid of the IDs)

---

*[thali/NextGeneration/tcpServersManager.js, line 349 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA-pc6dEuIZS2qoyg1:-KBA-pc6dEuIZS2qoyg2:-694734861) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L349)):*
A 164 line Javascript function? I'm fairly certain that constitutes cruel and depraved behavior in most jurisdictions. Perhaps it should be refactored into something slightly more digestible?

---

*[thali/NextGeneration/tcpServersManager.js, line 383 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA2_5Cn9T19cxYrq-Z:-KBA2_5Cn9T19cxYrq-_:-39399602) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L383)):*
The function's actual name is 'makeIntoCloseAllServer'. I'd appreciate it if you would use the full name only because every time I see "CloseAllServer" I think the server is being closed and wonder why that's happening at this point!

---

*[thali/NextGeneration/tcpServersManager.js, line 396 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA6WwQT17VkFbAhsYx:-KBA6WwQT17VkFbAhsYy:62793381) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L396)):*
Isn't this code doing exactly what makeIntoCloseAllServer is doing except it uses end (which leaves behind half opened connections) rather than destroy (which closes everything)?

---

*[thali/NextGeneration/tcpServersManager.js, line 419 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA7hkDf1WF1wT2JuWA:-KBA7hkEXHCRS475644r:-2095459917) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L419)):*
As the docs say, http://jxcore.com/docs/net.html#socketsettimeouttimeout-callback, a timeout doesn't close the connection. In fact, it doesn't actually do anything. So how can we safely remove the connection from _incoming? Don't we have to wait until close? Also, shouldn't we call destroy on ourselves so we can clean up this connection?

More generally, how do we ever manage to generate this event? It should only be generated in response to socket.setTimeout but I can't find us ever calling that method although I do believe it is required by the spec for what I suspect are reasonably obvious reasons.

---

*[thali/NextGeneration/tcpServersManager.js, line 462 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA9jAsz0aM2Cug_UoR:-KBA9jAsz0aM2Cug_UoS:-2085823478) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L462)):*
I couldn't find an 'end' event anywhere, did you mean 'close'?

---

*[thali/NextGeneration/tcpServersManager.js, line 470 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA9wBkj6TOGCaV_ssf:-KBA9wBkj6TOGCaV_ssg:-442718560) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L470)):*
Per http://jxcore.com/docs/net.html#event-error any time error is called you are guaranteed there will be a subsequent call to close. So we should be able to skip the call to removeArrayEelement and instead wait for the close event to trigger it.

---

*[thali/NextGeneration/tcpServersManager.js, line 489 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBABnSRryS_KjTxOqks:-KBABnSSf_92KFFb_Xop:265168260) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L489)):*
Where was mux._incoming defined?

---

*[thali/NextGeneration/tcpServersManager.js, line 490 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KB9kxKfPuj0jlSlW2c0:-KB9kxKgTu5wmdFkYbVl:-416338305) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L490)):*
I think this line is an error. Isn't this already handled in the close handler for the incoming connection?

---

*[thali/NextGeneration/tcpServersManager.js, line 505 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBACsyAfp340VLHduTG:-KBACsyAfp340VLHduTH:-1945908465) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L505)):*
If listen fails shouldn't we reject(err) the result so it propagates to start and gets returned to the caller? Otherwise aren't we just black holing this error outside of the log?

---

*[thali/NextGeneration/tcpServersManager.js, line 704 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAF7bTtIH_eciaEE2E:-KBAF7bTtIH_eciaEE2F:-1365413680) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L704)):*
313 lines long? A little refactoring, maybe even into different files, could really help with readability.

---

*[thali/NextGeneration/tcpServersManager.js, line 728 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAELhrsNsZOHdDw5oz:-KBAELhsYdjLhT3LRSrg:250176135) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L728)):*
Shouldn't this call resolve with the port of the peerServer per the spec's requirement that if createPeerListener is called on a peer we are already listening for then we will return the port we are using?

---

*[thali/NextGeneration/tcpServersManager.js, line 868 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBANRECs889aFcKSFE1:-KBANRECs889aFcKSFE2:2002445816) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L868)):*
Isn't firstConnection always going to be true at this point because it was set at the start of onNewConnection and nothing will have run between then and now to have changed its value? In which case, why is it in the if statement

I think the real issue is that there seems to be an assumption in the code that somehow firstConnection is going to stick around since it gets set in various subfunctions but it will disappear as soon as we exit the parent method. It isn't defined on any object and while we write to it 3 different places we only check it once and that once is here before anything else could have happened. So what firstConnection was meant to do it can't be doing it. Something isn't quite right.

---

*[thali/NextGeneration/tcpServersManager.js, line 880 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAPR90elN8Bo1RcocJ:-KBAPR90elN8Bo1RcocK:281106141) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L880)):*
Shouldn't this be hooked into the promise so that we only get resolve to the parent promise if we successfully were able to connect to the peer?

---

*[thali/NextGeneration/tcpServersManager.js, line 922 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAGv2OCL4F8f5lQxoh:-KBAGv2OCL4F8f5lQxoi:1144969850) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L922)):*
In the spec we didn't just say the oldest server but rather the oldest inactive server. So I would have expected that anytime a server has active content being sent through it that we would update the [0] date value. That way the server to be closed is the one that was used in the most distant past. We don't care when it was originally created, only when it was last used.

---

*[thali/NextGeneration/tcpServersManager.js, line 943 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAIVMXGlm32l-uc9XW:-KBAIVMYyvYEESGij5s4:-2134881699) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L943)):*
This is an asynchronous callback so how do we ever end up calling reject? I can see the code below that will call reject if server is null but this error handle won't cause server to be null nor can it cause server to be null because it won't be called until after we have returned from createServer. It seems like there is a connection missing here between this error state and cleaning up a failed peerServers entry and removing it.

Personally I would go farther and say that createPeerListener MUST NOT resolve() until it knows that listen succeeded. When combined with pleaseConnect I suspect we will have to potentially track two promises in an array and use Promise.all to only return when both have been resolved.

---

*[thali/NextGeneration/tcpServersManager.js, line 948 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KB9z4bj9NJB9Z-_PGum:-KB9z4bkgt8GPcSUr-rS:2015462279) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L948)):*
Please have some pity for the future folks who have to maintain this file. Wouldn't it have been nicer to have had a value like { lastUsedDate: Date.now(), serverInstance: sever } rather than what we have now which is contextless references to _peerServers[peerIdentifier][0] and _peerServers[peerIdentifiery][1] everywhere?

Also I believe it is considered more correct to use Date.now() than new Date() since this lets us skip creating a new object.

---

*[thali/NextGeneration/tcpServersManager.js, line 955 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAI270ATev6H5mXv0m:-KBAI270ATev6H5mXv0n:2132187591) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L955)):*
In the createServer code we add an entry to _peerServers but if we are in this code then that entry is 'bad'. Shouldn't it be removed?

---

*[thali/NextGeneration/tcpServersManager.js, line 997 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAM5LPuWxjrbvaGrHJ:-KBAM5LPuWxjrbvaGrHK:-1093128147) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L997)):*
The error could happen at just about any time, including far in the future, long after a response to the promise has been sent. Now the good news is that trying to resolve/reject the same promise more than once is not, I believe, an error. I think after the first call to promise/reject any subsequent calls are ignored. But this whole thing feels kind of like a race condition. If we are lucky and the error happens early enough then we get the reject otherwise it is invisible. But maybe that is o.k.?

---

*[thali/NextGeneration/tcpServersManager.js, line 1000 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAMRGnYHA42jaW-pUq:-KBAMRGnYHA42jaW-pUr:1672161310) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L1000)):*
Where did the call to outgoing.setTimeout happen to cause this event to fire?

---

*[thali/NextGeneration/tcpServersManager.js, line 1004 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAMdowGvoIAG_G-8z9:-KBAMdowGvoIAG_G-8zA:-869916621) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L1004)):*
I thought we cannot rely on the output of address() until we got the listen callback? So don't we have to wait until there before we can resolve?

---

*[thali/NextGeneration/tcpServersManager.js, line 1008 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAMn37uNpaEbvC3akq:-KBAMn38YOJ13b16RMeX:-869916621) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L1008)):*
I thought we cannot rely on the output of address() until we got the listen callback? So don't we have to wait until there before we can resolve?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-:-KB9Xjy5qKdvcmd1W8Lp:64672192)*

ThaliTester:
PR is added to the queue for testing as 1. task. (7fe97b0)

ThaliTester:
Test (Fail) 60124347 build is completed (7fe97b0)

See https://github.com/ThaliTester/TestResults/tree/601243477fe97b0_tcpServersManager_and_tests_tobybrad/ for the logs

vjrantal:
*[thali/NextGeneration/tcpServersManager.js, line 885 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBqIu-l0tbpi30EWOYA:-KBqIu-l0tbpi30EWOYB:-1439794001) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7fe97b0340568c4e2c916fa2c3a56323293823fc/thali/NextGeneration/tcpServersManager.js#L885)):*
After quite a bit of debugging, I realized why this (and some following code) does not work on devices.

That is, because both iOS and Android return the connection object as string. If one checks `testThaliMobileNative.js`, it can be seen that the callbacks to connect have something like `connection = JSON.parse(connection);`, which would be needed here as well.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561)*

ThaliTester:
PR is added to the queue for testing as 1. task. (6e0f5f5)

tobybrad:
Review status: 2 of 8 files reviewed at latest revision, 24 unresolved discussions.

---

*[thali/NextGeneration/tcpServersManager.js, line 349 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA-pc6dEuIZS2qoyg1:-KBsexk3c3VUTXVr3rj7:-929299872) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L349)):*
Will get to...

---

*[thali/NextGeneration/tcpServersManager.js, line 419 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA7hkDf1WF1wT2JuWA:-KBhovcn18Rq6WQI4J6S:1567340907) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L419)):*
we didn't create incoming, the native layer did. Should we really be setting timeouts on and closing socket we didn't originate ?? Nonetheless.. I've set it to 30s.

---

*[thali/NextGeneration/tcpServersManager.js, line 462 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBA9jAsz0aM2Cug_UoR:-KBDEjt0rZJzpSCqghI6:422969870) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L462)):*
https://nodejs.org/api/net.html#net_event_end

---

*[thali/NextGeneration/tcpServersManager.js, line 704 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBAF7bTtIH_eciaEE2E:-KBsedDlRsJe-R9HUETp:-929299872) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/482d94bb2f633ce8479f29c35c4f2660aca66bf2/thali/NextGeneration/tcpServersManager.js#L704)):*
Will get to...

---

*[thali/NextGeneration/tcpServersManager.js, line 885 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561#-KBqIu-l0tbpi30EWOYA:-KBrcUXq43ZzHQsb4S4x:-1571535261) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7fe97b0340568c4e2c916fa2c3a56323293823fc/thali/NextGeneration/tcpServersManager.js#L885)):*
Yeah. so this is a limitation of JXCore. You can either return JSON or a list of params. You can't return a mix of the two.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561)*

ThaliTester:
PR is added to the queue for testing as 2. task. (0083d9b)

ThaliTester:
PR is added to the queue for testing as 3. task. (c8c8bd4)

ThaliTester:
Test (Fail) 60124347 build is completed (6e0f5f5)

See https://github.com/ThaliTester/TestResults/tree/601243476e0f5f5_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Fail) 60124347 build is completed (6e0f5f5)

See https://github.com/ThaliTester/TestResults/tree/601243476e0f5f5_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Fail) 60124347 build is completed (0083d9b)

See https://github.com/ThaliTester/TestResults/tree/601243470083d9b_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Fail) 60124347 build is completed (0083d9b)

See https://github.com/ThaliTester/TestResults/tree/601243470083d9b_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Fail) 60124347 build is completed (c8c8bd4)

See https://github.com/ThaliTester/TestResults/tree/60124347c8c8bd4_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Fail) 60124347 build is completed (c8c8bd4)

See https://github.com/ThaliTester/TestResults/tree/60124347c8c8bd4_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (97f7ca7)

ThaliTester:
Test 6012434797f7ca7(97f7ca7) has failed

See https://github.com/ThaliTester/TestResults/tree/6012434797f7ca7_tcpServersManager_and_tests_tobybrad/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (e9830ee)

ThaliTester:
Test (Fail) 60124347 build is completed (e9830ee)

See https://github.com/ThaliTester/TestResults/tree/60124347e9830ee_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (d4f5b03)

ThaliTester:
Test (Success) 60124347 build is completed (d4f5b03)

See https://github.com/ThaliTester/TestResults/tree/60124347d4f5b03_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test 60124347d4f5b03(d4f5b03) has failed

See https://github.com/ThaliTester/TestResults/tree/60124347d4f5b03_tcpServersManager_and_tests_tobybrad/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (64ab483)

ThaliTester:
Test (Success) 60124347 build is completed (64ab483)

See https://github.com/ThaliTester/TestResults/tree/6012434764ab483_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test 6012434764ab483(64ab483) has failed

See https://github.com/ThaliTester/TestResults/tree/6012434764ab483_tcpServersManager_and_tests_tobybrad/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (40c277f)

ThaliTester:
Test (Fail) 60124347 build is completed (40c277f)

See https://github.com/ThaliTester/TestResults/tree/6012434740c277f_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (1b7fae3)

ThaliTester:
Test (Fail) 60124347 build is completed (1b7fae3)

See https://github.com/ThaliTester/TestResults/tree/601243471b7fae3_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Fail) 60124347 build is completed (1b7fae3)

See https://github.com/ThaliTester/TestResults/tree/601243471b7fae3_tcpServersManager_and_tests_tobybrad/ for the logs

yaronyg:
Reviewed 1 of 5 files at r2, 2 of 5 files at r3, 1 of 4 files at r5, 10 of 14 files at r8, 2 of 4 files at r10, 3 of 5 files at r11, 2 of 2 files at r12, 1 of 1 files at r13.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561)*

ThaliTester:
PR is added to the queue for testing as 1. task. (13fea37)

ThaliTester:
Test (Success) 60124347 build is completed (13fea37)

See https://github.com/ThaliTester/TestResults/tree/6012434713fea37_tcpServersManager_and_tests_tobybrad/ for the logs

yaronyg:
Reviewed 1 of 1 files at r14.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/561)*

ThaliTester:
Test 6012434713fea37(13fea37) has failed

See https://github.com/ThaliTester/TestResults/tree/6012434713fea37_tcpServersManager_and_tests_tobybrad/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (d1a8dac)

ThaliTester:
PR is added to the queue for testing as 2. task. (d1a8dac)

ThaliTester:
Test (Success) 60124347 build is completed (d1a8dac)

See https://github.com/ThaliTester/TestResults/tree/60124347d1a8dac_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test 60124347d1a8dac(d1a8dac) has failed

See https://github.com/ThaliTester/TestResults/tree/60124347d1a8dac_tcpServersManager_and_tests_tobybrad/ for the fail logs

ThaliTester:
Test (Success) 60124347 build is completed (d1a8dac)

See https://github.com/ThaliTester/TestResults/tree/60124347d1a8dac_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test (Success) 60124347 build is completed (e678b8e)

See https://github.com/ThaliTester/TestResults/tree/60124347e678b8e_tcpServersManager_and_tests_tobybrad/ for the logs

ThaliTester:
Test 60124347e678b8e(e678b8e) has failed

See https://github.com/ThaliTester/TestResults/tree/60124347e678b8e_tcpServersManager_and_tests_tobybrad/ for the fail logs

PR #562, opened 2016-02-22
More features to thaliMobile

vjrantal:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/562)

### DIFF
--- a/test/www/jxcore/bv_tests/testThaliMobile.js
+++ b/test/www/jxcore/bv_tests/testThaliMobile.js
@@ -1,10 +1,12 @@
 'use strict';
 
 var ThaliMobile = require('thali/NextGeneration/thaliMobile');
+var ThaliMobileNativeWrapper = require('thali/NextGeneration/thaliMobileNativeWrapper');
 var tape = require('../lib/thali-tape');
 var testUtils = require('../lib/testUtils.js');
 var express = require('express');
 var validations = require('thali/validations');
+var sinon = require('sinon');
 
 var verifyCombinedResultSuccess = function (t, combinedResult, message) {
   t.equal(combinedResult.wifiResult, null, message || 'error should be null');
@@ -120,6 +122,26 @@ test('network changes emitted', function (t) {
     testUtils.toggleWifi(true);
   });
   testUtils.toggleWifi(false);
+});
+
+test('does not send duplicate availability changes', function (t) {
+  var dummyPeer = {
+    peerIdentifier: 'dummy',
+    portNumber: 8080
+  };
+  var spy = sinon.spy(ThaliMobile.emitter, 'emit');
+  ThaliMobileNativeWrapper.emitter.emit('nonTCPPeerAvailabilityChangedEvent',
+                                        [dummyPeer]);
+  process.nextTick(function () {
+    t.equals(spy.callCount, 1, 'should be called once');
+    ThaliMobileNativeWrapper.emitter.emit('nonTCPPeerAvailabilityChangedEvent',
+                                          [dummyPeer]);
+    process.nextTick(function () {
+      t.equals(spy.callCount, 1, 'should not have been called more than once');
+      ThaliMobile.emitter.emit.restore();
+      t.end();
+    });
+  });
 });
 
 if (!tape.coordinated) {
--- a/test/www/jxcore/bv_tests/testThaliMobile.js
+++ b/test/www/jxcore/bv_tests/testThaliMobile.js
@@ -168,36 +168,38 @@ test('does not send duplicate availability changes', function (t) {
 test('calls correct starts when network changes', function (t) {
   var listeningSpy = null;
   var advertisingSpy = null;
-  testUtils.toggleRadios(false)
-  .then(function () {
-    return ThaliMobile.start(express.Router());
-  })
-  .then(function () {
-    return ThaliMobile.startListeningForAdvertisements();
-  })
-  .then(function (combinedResult) {
-    t.equals(combinedResult.wifiResult.message, 'Radio Turned Off',
-             'specific error expected');
-    return ThaliMobile.startListeningForAdvertisements();
-  })
-  .then(function (combinedResult) {
-    t.equals(combinedResult.wifiResult.message, 'Radio Turned Off',
-          'specific error expected');
-    listeningSpy = sinon.spy(ThaliMobile,
-      'startListeningForAdvertisements');
-    advertisingSpy = sinon.spy(ThaliMobile,
-      'startUpdateAdvertisingAndListening');
-    return testUtils.toggleRadios(true);
-  })
-  .then(function () {
-    t.equals(listeningSpy.callCount, 1,
-      'startListeningForAdvertisements should have been called');
-    t.equals(advertisingSpy.callCount, 1,
-      'startUpdateAdvertisingAndListening should have been called');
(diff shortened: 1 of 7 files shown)


yaronyg:
Please resolve the comments as you see fit and then :shipit:. Thanks!

---

Reviewed 7 of 7 files at r1.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 231 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBFXmmBG8w_Ia4fi25:-KBBFXmmBG8w_Ia4fi26:1779311948) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/test/www/jxcore/lib/wifiBasedNativeMock.js#L231)):*
Wouldn't it be easier to just set peerAvailabilities to {}?

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 505 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBG-E7X-cNC0AZL3bz:-KBBG-E8kBQbSorDtYL_:-911709433) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/test/www/jxcore/lib/wifiBasedNativeMock.js#L505)):*
Unless we are going to test tcpServersManager on iOS devices I don't see much of a choice but for us to put in the ability to simulate the ID behavior. :( It really shouldn't be that hard. Each node just needs to generate a UUID and then compare it.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 626 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBGHgNGeALIfqeiAUV:-KBBGHgNGeALIfqeiAUW:214937112) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/test/www/jxcore/lib/wifiBasedNativeMock.js#L626)):*
It's generally not a good idea to call a callback until all related work has completed. Otherwise one can end up in race conditions where the caller thinks certain work has happened but it has not. It can also cause situations where the code in the callback is re-entrant or otherwise causes state changes that interfere with the function. This was one of the reasons that promises were introduced (besides exiting the nesting pyramid of doom), they guarantee that the next user code will only run after the function is complete. In this case we have to use callbacks but we should make sure that the callback is always the very last thing the function does before it exits.

---

*[thali/NextGeneration/thaliMobile.js, line 38 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBA_93t00SEbz09BxNF:-KBA_93t00SEbz09BxNG:152659201) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/thali/NextGeneration/thaliMobile.js#L38)):*
Why not just declare this as function handleNEtworkChange (networkChangedValue) {..} ?

---

*[thali/NextGeneration/thaliMobile.js, line 66 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBA_S9at_3vHSYA3QK9:-KBA_S9at_3vHSYA3QKA:-669469586) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/thali/NextGeneration/thaliMobile.js#L66)):*
Maybe it's just me but when we are defining an inline, non-exported, function I find function checkErrors (){} easier to read.

---

*[thali/NextGeneration/thaliMobile.js, line 580 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBEs4u4JTktTc1xIxK:-KBBEs4u4JTktTc1xIxL:-365045677) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/thali/NextGeneration/thaliMobile.js#L580)):*
Is this even theoretically possible? When we advertise a peer across WiFi or BLE or MPCF don't we always use a new ID every time we change the peer?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-:-KBBI2Fa3ROx_OJNMQWf:-497206490)*

vjrantal:
Review status: 6 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 231 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBFXmmBG8w_Ia4fi25:-KBC56yxm-KXwVQtCk5Z:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/test/www/jxcore/lib/wifiBasedNativeMock.js#L231)):*
Done.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 505 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBG-E7X-cNC0AZL3bz:-KBC59IFZAEyOKQu43e9:-1839342599) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/test/www/jxcore/lib/wifiBasedNativeMock.js#L505)):*
We will most likely get into this once we start testing `tcpServersManager`.

---

*[test/www/jxcore/lib/wifiBasedNativeMock.js, line 626 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBGHgNGeALIfqeiAUV:-KBC65C05Wm43yG7nFcG:1194202379) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/test/www/jxcore/lib/wifiBasedNativeMock.js#L626)):*
Done.

I agree. I did it that way since I wanted that the `networkChangedCallback` gets fired after the callback is called, but I changed the code to do it after a timeout, which should also do the trick.

---

*[thali/NextGeneration/thaliMobile.js, line 580 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562#-KBBEs4u4JTktTc1xIxK:-KBC6wOkf1fLwQoSUKla:625280022) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/5791f113ed016b69b0faec7ed156873b677289fa/thali/NextGeneration/thaliMobile.js#L580)):*
This piece of code gets triggered, for example, every time we get an advertisement over SSDP. If nothing has changed with the peer who advertised, the peer identifier remains the same and thus we need this check in thaliMobile to make sure we bubble up only actual changes.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562)*

vjrantal:
Reviewed 4 of 7 files at r1, 3 of 3 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/562)*

PR #564, opened 2016-02-23
Mostly linting

yaronyg:
testThaliTCPServersManager - lint fixes

makeIntoCloseAllServer - Some linting and making sure that close got called
first and THEN we close the connections and THEN call the callback.

tcpServersManager - Linting

thalilogger - Linting

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/564)

### DIFF
--- a/test/www/jxcore/bv_tests/testThaliTCPServersManager.js
+++ b/test/www/jxcore/bv_tests/testThaliTCPServersManager.js
@@ -43,7 +43,7 @@ test("calling startNativeListener directly throws", function(t) {
   var serversManager = new ThaliTCPServersManager(4242);
   serversManager.start()
   .then(function(localPort) {
-    serversManager.createNativeListener();
+    serversManager._createNativeListener();
   })
   .catch(function(err) {
     t.equal(err.message, "Don't call directly!", "Should throw");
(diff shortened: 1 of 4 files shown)


vjrantal:
Reviewed 4 of 4 files at r1.

---

*[thali/NextGeneration/makeIntoCloseAllServer.js, line 56 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/564#-KBCTbugL_u4jRMgWlS9:-KBCTbugL_u4jRMgWlSA:-507802071) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2417faeb6452444b9422b25116ebb65f253a9208/thali/NextGeneration/makeIntoCloseAllServer.js#L56)):*
I might be wrong, but I believe the callback for the close function is called only after the server is successfully closed. Now that connection destroy operations were moved to be done after the callback is called, I think that never happens and thus the closeAll call might overall never return (unless someone somewhere external to this code closes the connections).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/564)*

Issue #563, opened 2016-02-22
[Android] ThaliTest app crashes during start-up


tompaana:
Reproduced with commit [`8e7e29ab668e5569dc9c8c212030295f41ed47bd`](https://github.com/thaliproject/Thali_CordovaPlugin/commit/8e7e29ab668e5569dc9c8c212030295f41ed47bd) and on Lollipop.

No clear reason, but saw this: https://github.com/jxcore/jxcore-cordova/blob/master/src/android/java/io/jxcore/node/jxcore.java#L168

In addition, commenting out line 84 in ThaliPermissions.java (https://github.com/thaliproject/Thali_CordovaPlugin/blob/8e7e29ab668e5569dc9c8c212030295f41ed47bd/src/android/java/org/thaliproject/p2p/ThaliPermissions.java#L84) <strike>fixed the problem, but is obviously not an actual fix</strike> removes the crash, but nothing is happening after - the app just freezes functionality-wise.

tompaana:
The things that has broken the code has happened after (9b58dbf88f0f3cc7922802f1f50034f2c357665d)[https://github.com/thaliproject/Thali_CordovaPlugin/commit/9b58dbf88f0f3cc7922802f1f50034f2c357665d] in vNext branch.

tompaana:
Commenting out stuff in app.js (UnitTest_app.js) makes the bug go away:
```js
//testUtils.toggleRadios(true)
//.then(function () {
  Mobile('GetDeviceName').callNative(function (name) {
    console.log('My device name is: %s', name);
    testUtils.setName(name);
    require('./runTests.js');
  });
//})
//.catch(function (error) {
//  console.log('Something went wrong: ' + error);
//});
```

vjrantal:
Without commenting out anything, i.e., when the crash occurs, I see something like this in logcat:

```
14:52:24.907 511/511                     DEBUG  I   *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
14:52:24.907 511/511                     DEBUG  I   Build fingerprint: 'htc/HTCOneM9vzw/htc_himawl:5.0.2/LRX22G/511781.15:user/release-keys'
14:52:24.907 511/511                     DEBUG  I   Revision: '0'
14:52:24.907 511/511                     DEBUG  I   ABI: 'arm'
[...]

PR #572, opened 2016-02-24
Get rid of the radio toggling at the end of tests

vjrantal:
The radio toggling at the beginning of the tests was removed as a workaround
for #563. That would be good to put back once a root cause and fix is found,
but even then, it should not be necessary to turn off radios at the end of
tests.

This was originally done for CI where devices didn't reboot in between tests,
but now devices do reboot and thus tests always start from clean radio states.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/572)


tompaana:
Reviewed 4 of 4 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/572)*

PR #585, opened 2016-02-26
ThalinotificatioServer for review

juhanak:
[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/585)

### DIFF
--- a/test/www/jxcore/bv_tests/testThaliNotificationServer.js
+++ b/test/www/jxcore/bv_tests/testThaliNotificationServer.js
@@ -46,9 +46,17 @@ var GlobalVariables = function () {
   MockThaliMobile.startUpdateAdvertisingAndListening = function () {
     return Promise.resolve();
   };
+
+  // Mocks ThaliMobile.stopAdvertisingAndListening function
+  MockThaliMobile.stopAdvertisingAndListening = function () {
+    return Promise.resolve();
+  };
   
   this.spyStartUpdateAdvertisingAndListening = 
     sinon.spy(MockThaliMobile, 'startUpdateAdvertisingAndListening');
+
+  this.spyStopAdvertisingAndListening = 
+    sinon.spy(MockThaliMobile, 'stopAdvertisingAndListening');
   
 };
 
@@ -64,7 +72,6 @@ GlobalVariables.prototype.init = function () {
       if (err != null) {
         reject(err);
       } else {
-
         self.TESTURL = 'http://' + '127.0.0.1' + ':' + 
         self.expressServer.address().port + 
         ThaliNotificationServer.NOTIFICATION_BEACON_PATH;
@@ -145,8 +152,13 @@ test('Start and stop ThaliNotificationServer', function (t) {
   
   globalVariables.notificationServer.start(publicKeysToNotify).then(function () {
     globalVariables.notificationServer.stop().then(function () {
+
       t.equals(globalVariables.spyStartUpdateAdvertisingAndListening.callCount,
-      2, 'StartUpdateAdvertisingAndListening should be called twice');
+      1, 'ThaliMobile.StartUpdateAdvertisingAndListening should be called once');
+      
+      t.equals(globalVariables.spyStopAdvertisingAndListening.callCount,
+      1, 'ThaliMobile.StopAdvertisingAndListening should be called once');
+      
       t.end();
     }).catch(function (failure) {
       t.fail('Stopping failed:' + failure);
@@ -171,7 +183,10 @@ test('Pass an empty array to ThaliNotificationServer.start', function (t) {
       t.equals(globalVariables.spyStartUpdateAdvertisingAndListening.callCount,
       0, 'StartUpdateAdvertisingAndListening should not be called');
       ThaliHttpTester.runTest(globalVariables.TESTURL, httpResponseHandler);
-      
+
+      t.equals(globalVariables.spyStopAdvertisingAndListening.callCount,
+      1, 'ThaliMobile.StopAdvertisingAndListening should be called once');
+
     }).catch(function (failure) {
       t.fail('Stopping failed:' + failure);
       t.end();
--- a/test/www/jxcore/bv_tests/testThaliNotificationServer.js
+++ b/test/www/jxcore/bv_tests/testThaliNotificationServer.js
@@ -1,0 +1,248 @@
+var tape = require('../lib/thali-tape');
+var express = require('express');
+var crypto = require('crypto');
+var sinon = require('sinon');
+var Promise = require('lie');
+
+var proxyquire = require('proxyquire').noCallThru();
+var NotificationBeacons = 
+  require('thali/NextGeneration/notification/thaliNotificationBeacons');
+var ThaliNotificationServer = 
+  require('thali/NextGeneration/notification/thaliNotificationServer');
+var thaliHttpTester = require('../lib/httpTester');
+
+var SECP256K1 = 'secp256k1';
+var PORT = 3001;
+var TESTURL = 'http://localhost:' + PORT + 
+  ThaliNotificationServer.NOTIFICATION_BEACON_PATH; 
+
+var globalVariables = {};
+
+/**
+ * Initializes globalVariables before the each test run.
+ */
+function initializeGlobalVariables() {
+  globalVariables = {
+    clientRequestCounter : 0,
+    spyStartUpdateAdvertisingAndListening : null,
+    expressApp : express(),
+    expressServer : null,
+    ThaliNotificationServerProxyquired : null
+  };
+  
+  globalVariables.ThaliNotificationServerProxyquired = 
+    proxyquireThaliNotificationServer();
+}
+
+/**
+ * Frees reserved resources from globalVariables after the each test run.
+ */
+function destructGlobalVariables() {
+  if (globalVariables) {
+    if (globalVariables.expressServer) {
+      globalVariables.expressServer.close();
+    }
+  }
+}
+
+/**
+ * Creates a proxyquired ThaliNotificationServer class. 
+ */
+function proxyquireThaliNotificationServer() {
+  var MockThaliMobile = { };
+  var ThaliNotificationServerProxyquired =
+    proxyquire('thali/NextGeneration/notification/thaliNotificationServer',
+      { '../thaliMobile':
+      MockThaliMobile});
+
+  MockThaliMobile.startUpdateAdvertisingAndListening = function () {
+    return Promise.resolve();
+  };
+  
+  globalVariables.spyStartUpdateAdvertisingAndListening = 
+    sinon.spy(MockThaliMobile, 'startUpdateAdvertisingAndListening');
+  
+  return ThaliNotificationServerProxyquired;
+}
+
+var test = tape({
+  setup: function (t) {
+    initializeGlobalVariables();
+    t.end();
+  },
+  teardown: function (t) {
+    destructGlobalVariables();
+    t.end();
+  }
+});
+
+/**
+ * Initializes ThaliNotificationServer from proxyquired 
+ * ThaliNotificationServerProxyquired class.
+ */
+function initializeThaliNotificationServer() {
+  var expressRouter = express.Router();
+  var myPublicKey = crypto.createECDH(SECP256K1);
+  myPublicKey.generateKeys();
+  return new globalVariables.ThaliNotificationServerProxyquired( 
+    expressRouter, myPublicKey, 90000);
+}
+
+/**
+ * Generates two public keys for the test. 
+ */
+function generatePublicKeys() {
+  var publicKeys = [];
+  var device1 = crypto.createECDH(SECP256K1);
+  var device1Key = device1.generateKeys();
+  var device2 = crypto.createECDH(SECP256K1);
+  var device2Key = device2.generateKeys();
+  publicKeys.push(device1Key, device2Key);
+  return publicKeys;
+}
+
+test('Start and stop ThaliNotificationServer', function (t) {
+
+  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+  
+  thaliNotificationServerProxyquired.start(generatePublicKeys()).then(function () {
+    t.ok(true, 'Starting works');
+    thaliNotificationServerProxyquired.stop().then(function () {
+      thaliNotificationServerProxyquired;
+      if (globalVariables.spyStartUpdateAdvertisingAndListening.calledTwice) {
+        t.ok(true, 'Stopping works');
+      } else {
+        t.fail('StartUpdateAdvertisingAndListening is not called twice');
+      }
+      t.end();
+    }).catch(function (failure) {
+      t.fail('Stopping failed:' + failure);
+      t.end();
+    });
+  }).catch(function (failure) {
+    t.fail('Starting failed:' + failure);
+    t.end();
+  });
+});
+
+test('ThaliNotificationServer.start with wrong parameters', function (t) {
+
+  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+  
+  thaliNotificationServerProxyquired.start('wrong key').then(function () {
+      t.fail('start should have failed.');
+      t.end();
+
+    }).catch(function () {
+      t.pass('Expected result.');
+      t.end();
+    });
+});
+
+test('ThaliNotificationServer.start with empty array', function (t) {
+
+  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+  
+  thaliNotificationServerProxyquired.start([]).then(function () {
+      t.pass('Expected result.');
+      t.end();
+    }).catch(function () {
+      t.fail('start should have failed.');
+      t.end();
+    });
+});
+
+
+test('Makes an HTTP request to /NotificationBeacons', function (t) {
+  
+  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+
+  var httpResponseHandler = function (error, response) { 
+    if (error == null && response.statusCode == 200) {
+      t.equal(response.headers['cache-control'], 'no-cache');
+      t.equal(response.headers['content-type'], 'application/octet-stream');
+    } else {
+      t.fail('Test failed. Response code is not 200');
+    }
+    t.end();
+  };
+  
+  var testArray = [
+    { url:TESTURL, delay:0, handler: httpResponseHandler}
+  ];
+
+  thaliNotificationServerProxyquired.start(generatePublicKeys()).then(function () {
+    globalVariables.expressApp.use('/', 
+      thaliNotificationServerProxyquired._router);
+    globalVariables.expressServer = globalVariables.expressApp.listen(PORT, 
+      function () {
+      thaliHttpTester.runTest(testArray);
+    });
+  });
+});
+
+test('Makes an HTTP request to /NotificationBeacons (no public keys)', function (t) {
+  
+  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+
+  var httpResponseHandler = function (error, response) { 
+    if (error == null && response.statusCode == 204) {
+      t.pass('Response code is 204 as expected');
+    } else {
+      t.fail('Test failed. Response code is not 204');
+    }
+    t.end();
+  };
+  
+  var testArray = [
+    { url:TESTURL, delay:0, handler: httpResponseHandler}
+  ];
+
+  thaliNotificationServerProxyquired.start(null).then(function () {
+    globalVariables.expressApp.use('/', 
+      thaliNotificationServerProxyquired._router);
+    globalVariables.expressServer = globalVariables.expressApp.listen(PORT, 
+      function () {
+      thaliHttpTester.runTest(testArray);
+    });
+  });
+});
+
+test('Test to exceed the rate limit of /NotificationBeacons', function (t) {
+  
+  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+  var responseCounter = 0;
+  var httpResponseHandler = function (error, response) {
+    responseCounter++;
+    if (responseCounter >= ThaliNotificationServer.WINDOW_SIZE + 1){
+      if (response.statusCode == 503) {
+        t.pass('Response is 503 as expected when the rate is exceeded');
+      } else {
+        t.fail('Server should return 503 when the rate is exceeded');
+      }
+      t.end();
+    } else {
+      if (response.statusCode != 200) {
+        t.fail('Server should return 200 when the rate limit is not exceeded');
+        t.end();
+      }
+    }
+  };
+  var testArray = [];
+
+  // RATE+2 ensures that the rate limit is hit
+  var delayMultiplayer = 1 / (ThaliNotificationServer.RATE+2); 
+  for (var i = 0 ; i < ThaliNotificationServer.WINDOW_SIZE + 1 ; i++) {
+    testArray.push({ url:TESTURL, delay:i*delayMultiplayer, 
+                      handler: httpResponseHandler});
+  }
+
+  thaliNotificationServerProxyquired.start(generatePublicKeys()).then(function () {
+    globalVariables.expressApp.use('/', 
+      thaliNotificationServerProxyquired._router);
+    globalVariables.expressServer = globalVariables.expressApp.listen(PORT, 
+      function () {
+        thaliHttpTester.runTest(testArray);
+      });
+  });
+});
--- a/test/www/jxcore/bv_tests/testThaliNotificationServer.js
+++ b/test/www/jxcore/bv_tests/testThaliNotificationServer.js
@@ -1,3 +1,4 @@
+'use strict';
 var tape = require('../lib/thali-tape');
 var express = require('express');
 var crypto = require('crypto');
@@ -9,111 +10,143 @@ var NotificationBeacons =
   require('thali/NextGeneration/notification/thaliNotificationBeacons');
 var ThaliNotificationServer = 
   require('thali/NextGeneration/notification/thaliNotificationServer');
-var thaliHttpTester = require('../lib/httpTester');
+  
+var MakeIntoCloseAllServer = 
+  require('thali/NextGeneration/makeIntoCloseAllServer');
+  
+var ThaliHttpTester = require('../lib/httpTester');
 
 var SECP256K1 = 'secp256k1';
-var PORT = 3001;
-var TESTURL = 'http://localhost:' + PORT + 
-  ThaliNotificationServer.NOTIFICATION_BEACON_PATH; 
 
 var globalVariables = {};
 
 /**
- * Initializes globalVariables before the each test run.
+ * @classdesc This class is a container for all variables and 
+ * functionality that are common to most of the ThaliNoficationServer
+ * tests. 
  */
-function initializeGlobalVariables() {
-  globalVariables = {
-    clientRequestCounter : 0,
-    spyStartUpdateAdvertisingAndListening : null,
-    expressApp : express(),
-    expressServer : null,
-    ThaliNotificationServerProxyquired : null
-  };
+var GlobalVariables = function () {
+
+  this.sourceKeyExchangeObject = crypto.createECDH(SECP256K1);
+  this.sourcePublicKey = this.sourceKeyExchangeObject.generateKeys();
+  this.sourcePublicKeyHash = 
+    NotificationBeacons.createPublicKeyHash(this.sourcePublicKey);
+
+  this.expressApp = express();
+  this.expressRouter = express.Router();
   
-  globalVariables.ThaliNotificationServerProxyquired = 
-    proxyquireThaliNotificationServer();
-}
-
-/**
- * Frees reserved resources from globalVariables after the each test run.
- */
-function destructGlobalVariables() {
-  if (globalVariables) {
-    if (globalVariables.expressServer) {
-      globalVariables.expressServer.close();
-    }
-  }
-}
-
-/**
- * Creates a proxyquired ThaliNotificationServer class. 
- */
-function proxyquireThaliNotificationServer() {
+  // Creates a proxyquired ThaliNotificationServer class.
   var MockThaliMobile = { };
-  var ThaliNotificationServerProxyquired =
+  this.ThaliNotificationServerProxyquired =
     proxyquire('thali/NextGeneration/notification/thaliNotificationServer',
       { '../thaliMobile':
       MockThaliMobile});
-
+  
+  // Mocks ThaliMobile.startUpdateAdvertisingAndListening function
   MockThaliMobile.startUpdateAdvertisingAndListening = function () {
     return Promise.resolve();
   };
   
-  globalVariables.spyStartUpdateAdvertisingAndListening = 
+  this.spyStartUpdateAdvertisingAndListening = 
     sinon.spy(MockThaliMobile, 'startUpdateAdvertisingAndListening');
   
-  return ThaliNotificationServerProxyquired;
-}
-
-var test = tape({
-  setup: function (t) {
-    initializeGlobalVariables();
-    t.end();
-  },
-  teardown: function (t) {
-    destructGlobalVariables();
-    t.end();
-  }
-});
+};
 
 /**
- * Initializes ThaliNotificationServer from proxyquired 
- * ThaliNotificationServerProxyquired class.
+ * Initializes express router and starts to listen incoming http requests. 
  */
-function initializeThaliNotificationServer() {
-  var expressRouter = express.Router();
-  var myPublicKey = crypto.createECDH(SECP256K1);
-  myPublicKey.generateKeys();
-  return new globalVariables.ThaliNotificationServerProxyquired( 
-    expressRouter, myPublicKey, 90000);
-}
+GlobalVariables.prototype.init = function () {
+  var self = this;
+  return new Promise(function (resolve, reject) {
+    // Initializes the server with the expressRouter
+    self.expressApp.use('/', self.expressRouter);
+    self.expressServer = self.expressApp.listen(0, function (err) {
+      if (err != null) {
+        reject(err);
+      } else {
+
+        self.TESTURL = 'http://' + '127.0.0.1' + ':' + 
+        self.expressServer.address().port + 
+        ThaliNotificationServer.NOTIFICATION_BEACON_PATH;
+
+        MakeIntoCloseAllServer(self.expressServer);
+
+        self.notificationServer = new self.ThaliNotificationServerProxyquired( 
+          self.expressRouter, self.sourceKeyExchangeObject, 90000);
+    
+        resolve();
+      }
+    });
+  });
+};
+
 
 /**
- * Generates two public keys for the test. 
+ * Frees reserved resources from globalVariables after the each test run.
  */
-function generatePublicKeys() {
-  var publicKeys = [];
+GlobalVariables.prototype.kill = function () {
+  var self = this;
+  return new Promise(function (resolve, reject) {
+    if (self.expressServer) {
+      self.expressServer.closeAll(function (error) {
+        self.expressServer = null;
+        if (error != null) {
+          reject(error);
+        } else {
+          resolve();
+        }
+      });
+    } else {
+      resolve();
+    }
+  });
+};
+
+/**
+ * Generates public keys for two target devices. 
+ */
+GlobalVariables.prototype.createPublicKeysToNotify = function () {
+  this.publicKeysToNotify = [];
+  this.targetDeviceKeyExchangeObjects = [];
+  
   var device1 = crypto.createECDH(SECP256K1);
   var device1Key = device1.generateKeys();
   var device2 = crypto.createECDH(SECP256K1);
   var device2Key = device2.generateKeys();
-  publicKeys.push(device1Key, device2Key);
-  return publicKeys;
-}
+  
+  this.publicKeysToNotify.push(device1Key, device2Key);
+  this.targetDeviceKeyExchangeObjects.push(device2, device2);
+  return this.publicKeysToNotify;
+};
+
+var test = tape({
+  setup: function (t) {
+    globalVariables = new GlobalVariables();
+    globalVariables.init().then(function () {
+      t.end();
+    }).catch(function (failure) {
+      t.fail('Test setting up failed:' + failure);
+      t.end();
+    });
+  },
+  teardown: function (t) {
+    globalVariables.kill().then(function () {
+      t.end();
+    }).catch(function (failure) {
+      t.fail('Server cleaning failed:' + failure);
+      t.end();
+    });
+  }
+});
 
 test('Start and stop ThaliNotificationServer', function (t) {
 
-  var thaliNotificationServerProxyquired = initializeThaliNotificationServer();
+  var publicKeysToNotify = globalVariables.createPublicKeysToNotify();
   
-  thaliNotificationServerProxyquired.start(generatePublicKeys()).then(function () {
-    t.ok(true, 'Starting works');
-    thaliNotificationServerProxyquired.stop().then(function () {
-      thaliNotificationServerProxyquired;
-      if (globalVariables.spyStartUpdateAdvertisingAndListening.calledTwice) {
-        t.ok(true, 'Stopping works');
-      } else {
-        t.fail('StartUpdateAdvertisingAndListening is not called twice');
-      }
+  globalVariables.notificationServer.start(publicKeysToNotify).then(function () {
+    globalVariables.notificationServer.stop().then(function () {
+      t.equals(globalVariables.spyStartUpdateAdvertisingAndListening.callCount,
+      2, 'StartUpdateAdvertisingAndListening should be called twice');
       t.end();
     }).catch(function (failure) {
       t.fail('Stopping failed:' + failure);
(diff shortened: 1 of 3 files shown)


yaronyg:
The first thing to check when you do a PR is if you passed appveyor (and hopefully starting next week, also if you passed CI). In this case you didn't. The tests failed on one of your tests. I didn't see any useful error information. However the test failed while running 'jx npm run test-coordinated' which you can also run on your desktop. Just see the README.md in the test file for directions.

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 30 unresolved discussions, some commit checks failed.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVBqefsDyoAR8oVWqo:-KBVBqegVvrbWja6HUiT:-753207049) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L1)):*
Please make sure to run JSHint and JSCS validation. It would have told you, for example, that this file isn't using 'use strict';.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVJFgSJykaP9YBT70R:-KBVJFgSJykaP9YBT70S:990461691) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L2)):*
There should be an additional test where we try to hit the beacon endpoint before we call start on thaliNotificationServer. This should give us a 401. It proves that we don't register the route until start is called.

Also the tests below that call start with various legal values should all be extended to then call stop and do a beacon request and prove that they got the proper empty response.

There should also be at least one test where you actually check the beacon value you got back from the HTTP server and prove that it matches the beacon value you got when you generated the beacons with the beacon library.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 16 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVCrn_5GRt-yn-dazd:-KBVCrn_5GRt-yn-daze:1124818905) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L16)):*
Over the years I've learned that sometimes 'localhost' (the DNS name) gets bound to strange places. I've learned to directly use 127.0.0.1. It seems to work better in practice.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 33 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVD4G0aOnY2fuLCi6I:-KBVD4G0aOnY2fuLCi6J:-17581302) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L33)):*
Why isn't this variable set in the object above? There is no semantic difference between globalVariables.ThaliNotificationServerProxyrquired = ... and including in the previous , ThaliNotificationServerProxyquired : proxyquireThaliNotificationServer()

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 43 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVDxhkcampDNME7Tpj:-KBVDxhl0CU29oPViYjZ:1548516520) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L43)):*
Close is not a synchronous function. It takes a callback that is called when the server is actually closed. Until you get that callback the port is still held onto. For your sanity I would make this function return a promise and then turn the callback into a promise resolve()/reject() based on the callback's result. Keep in mind that the callback will have a first argument that will be null if everything is fine and non-null if there is an error.

Also keep in mind that if the server you create doesn't close all of its incoming connections (which is NOT automatically handled by the close method) then the callback will never return and the test will hang. I would suggest that your test server use the makeIntoCloseAllServer function from Thali and use its closeAll method.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 106 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVEO6yaDhJIc_wpgf1:-KBVEO6yaDhJIc_wpgf2:1491456968) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L106)):*
Every single test starts with initializeThaliNotificationServer which makes sense but then shouldn't it just be in setup?

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 109 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVEiQlfaRMbx33DkXl:-KBVEiQlfaRMbx33DkXm:1554498239) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L109)):*
Content free tests (e.g. this test always passes) are seriously distracting when going through our already enormous debug logs. So please remove them. They should only be needed while debugging and you can use the thalilogger for that with logger.debug statements.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 111 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVF0VZyzrXhoeTz6rV:-KBVF0V_FrisL5O87WOl:1953171135) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L111)):*
What is the above statement supposed to do? It isn't an assignment. It isn't a function call. What is it? Note that one of the reasons I use IntelliJ is because it has awesome built in Javascript linting code that detects stuff like this automatically.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 112 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVFWYEfQLVxOWX4mbP:-KBVFWYEfQLVxOWX4mbQ:-1370529411) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L112)):*
This needs to be rewritten as t.ok(globalVariables.spyStartUpdateAdvertising...., 'StartUpdateAdvertisingAndListening is called twice); Again, please, don't use tests as your log.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 137 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVGsA_Lh0SZizjexyv:-KBVGsA_Lh0SZizjexyw:-492824471) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L137)):*
This is technically right but is it really useful? I think we can remove it.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 156 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVCfPaM5sbl3WtqOeF:-KBVCfPaM5sbl3WtqOeG:240065550) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L156)):*
As soon as I ran 'jx npm run test-coordinated' I got an EADDRINUSE error on this test.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 161 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVH67_sgQLQFWi4IMS:-KBVH67_sgQLQFWi4IMT:961543972) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L161)):*
Wouldn't it be simpler to get rid of the if/else and instead just say
t.equal(error, null, 'no error');
t.equal(response.statusCode, 200, '200 error code');
?

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 175 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVHW2hM13jLzpscDMS:-KBVHW2hM13jLzpscDMT:1335090603) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L175)):*
This approach to testing makes me a bit queasy. What I would rather see us do is that as part of setting up each test we take the router object that we will pass to thaliNotificationServer and before the test begins we automatically stick it into an app host listening on a port. At worst we will just tear it down when the test ends. No big deal. But the point is that by doing it this way we show that we set the right value on the right router object. That there wasn't a copy error or something weird going on. With the test the way it is currently written we could have somehow create another router object and we wouldn't know.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 189 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVIbJJJhX395k3yh3j:-KBVIbJJJhX395k3yh3k:-497696405) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L189)):*
Again, per above, it's cleaner to write t.equal(error, null, 'no error'); and t.equal(response.statusCode, 204, '204 status code); and be done. No if, no else, no pass and no fail.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 216 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVK7yXq4CTEz5skAw9:-KBVK7yXq4CTEz5skAwA:-1246916215) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L216)):*
This test is super duper fragile and probably will fail in really annoying ways. The reason for that is because when we run Node.js under constrained conditions (e.g. in a mobile device, in AppVeyor, etc.) the timers all get really crazy. I've seen 10 ms setTimeouts that didn't run for 4000 ms! So unfortunately you can't rely on the timers running when you think they will.

However there is a work around that I think is particularly useful in our case because the logic here is very simple and straight forward, in sinon there is a feature called Fake timers. It allows you to take over setTimeout and other objects and it lets you manually set the system time. So you can exactly control how far the clock moves forward every time one of your delayed requests runs. Using that will give you much more reliable test results. Just make sure to create a global variable called clock that is initialized to null. ONLY set that variable in this test and make sure the teardown code tests if clock isn't null and if so calls restore.

Of course if you agree to my proposal to get rid of the throttling then this test just goes away. :)

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVJwj_8O4fRD9rgFvO:-KBVJwj_8O4fRD9rgFvP:-614544246) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L227)):*
Why is there a t.end() both here and above? Why not just put a t.end() at the bottom of this function?

---

*[test/www/jxcore/lib/httpTester.js, line 17 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVAld1Nubi7zk29qM5:-KBVAld21dukc9y2t1Qy:-796012838) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/lib/httpTester.js#L17)):*
This is a JSCS violation. Please run JSCS validation. Oh and as in my previous comment, JSHint isn't being enforced here either. This would have told you things like you don't need promiseQueue above.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 29 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUks6GTKi0UUwRRI5n:-KBUks6GTKi0UUwRRI5o:287022682) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L29)):*
(NB means "Non Blocking", which means you can resolve this comment anyway you want, it shouldn't, on its own, prevent a checkin)
NB: You can replace these with assert tests. It's easier. E.g. assert(router !== null, "router must not be null");

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 68 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV5MxbHVxzLl-65GwA:-KBV5MxbHVxzLl-65GwB:-1879179147) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L68)):*
10? Wow... that seems really slow. Shouldn't we at least support say 100 requests/second? Especially when we are on WiFi that is quite reasonable.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 97 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUnAlbAPgsmAFheaW7:-KBUnAlbAPgsmAFheaW8:320505898) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L97)):*
I made a change to replication that i'd like to also make here. I specify publicKeystoNotify as {buffer[]} and I don't make the argument optional (e.g. remove the brackets). This means that the value submitted MUST always be an array. It can be empty but it must be an array. This simplified a lot of checks and use cases for me in the replication code (for example I could always use array operations against publicKeystoNotify because I always knew it was an array, even if empty).

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 115 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUnTrhQnYAG_mvibtj:-KBUn8Fav2GAfF7iS8D9:-1385616718) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L115)):*
So please accept empty arrays as that effectively becomes our new null so the check above would be assert(publicKeysToNotify !== null && publicKeysToNotify instanceof Array, "publicKeysToNotify must be an array");. Followed by a check for if (publicKeysToNotify.length === 0) { self.preambleAndBeacons = null } and then the rest of your logic here.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 147 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUozsyDnmzTnCwHG1p:-KBUozsyDnmzTnCwHG1q:580483078) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L147)):*
I don't think you need to check self._firstStartCall. The reason is that if this really is the first call then previousPreambleAndBeacons will be null because we initialize self._preambleAndBeacons to null in the constructor. 

Also removing self._firstStartCall has another nice side effect which is that you can stick "self._firstStartCall = false" (since you will have registered the notification path) which makes the code simpler. It means the if check for self._firstStartCall below can be removed.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 177 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUptj9OfZ6lJR_w0yh:-KBUptjAV8YDnmiy5xWR:1820586920) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L177)):*
Think about this... someone is calling stop and they are now being ask to call start somewhere else? That sounds... weird. There are two fixes here.

An obvious fix is that we could put in logic below to catch a 'Call Start!' error and swallow it. But I might suggest this isn't a great idea because error handling in Javascript is, at best, an inexact art where a single change of a string can ruin a bunch of code. There are tricks using sub-typing in Javascript to make for more robust error handling but they require a level of ceremony that is inconsistent with javascript's natural approach.

Which brings us to option 2. Why don't you set self._preambleAndBeacons to null and call ThaliMobile.stopAdvertisingAndListening()?

Now, you're probably thinking "Wait a minute, didn't YOU tell ME to make this into a call to start with a null argument?" And the answer is, yes, I did. And I was wrong. Beings wrong is a habit I haven't been able to break. :(

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 182 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUpiz5u62SFRfkWbkc:-KBUpiz5u62SFRfkWbkd:1720775104) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L182)):*
Depending on what you do with my previous comment on how I specified this wrong this would at the least turn into this.start([]). But hopefully we'll get rid of this line all together.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 218 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUvjQCACmOoaUHvJXn:-KBUvjQCACmOoaUHvJXo:823760422) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L218)):*
Why is this here? No one seems to check it and I can't immediately come up with a good reason for it.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUwz9ECF_GHbLzMGh5:-KBUwz9ECF_GHbLzMGh6:-1673027479) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L227)):*
Ask yourself a question.... why would any sane person want to use Javascript to write code? O.k. sure, if you are in a browser, you have no choice. But what about server side logic like Thali? No one with more than a few brain cells would want to use Javascript! Except... you see... there is this magical source of problem solving called NPM and NPM is super powerful. For example, in literally 10 seconds of querying I found three different libraries that will implement rate limiting on Express servers in NPM.  And keep in mind that if you widen the search to rate limiting Node.js servers you get an ocean of projects. So why would we write our own? But oh well, too late now.

But see my proposal below to rip all this code out.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 239 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV2JLtBVQ-Qzrr0IF1:-KBV2JLtBVQ-Qzrr0IF2:-690413119) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L239)):*
The current window size of 5 seems a bit small as it means a temporary bump will cause us to fail out. I would probably use at least say 100.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 240 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV2sO9mPGO1pKwbnf_:-KBV2sOAv60xqJQfT_uF:462104744) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L240)):*
I do wonder if shift is the wisest choice. It's certainly the simplest but I wonder what happens under really heavy load? The GC in Javascript has a tendency to explode under load. It's probably safer to allow a fixed array of WINDOW_SIZE and then use a wrap around index to decide where the oldest value is and where to put the newest value. But you can file that as a bug and stick it in icebox. Please mark it with the security label. But I don't think we need to fix it now and I'm not 100% sure it's actually an issue. We would have to run a load test where we did an enormous number of push/shifts and see what the GC does. Again, not worth worrying about right now. (Note that if you accept my proposal below to get rid of rate limiting then we don't need to file any bug on this since the code will go away)

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 243 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV3pdGMwSvZ_IaKTCo:-KBV3pdGMwSvZ_IaKTCp:69934834) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L243)):*
Please run JSHint. This file has at least 5 different JSHint rule violations. One of which is above. Note, btw, that the VSCode JSHint extension doesn't seem to work with our project.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 245 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVLWytdBP53-4nfDre:-KBVLWyuCAlylvkTfIzb:1945688307) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L245)):*
Ask yourself, what is the point of rate limiting? Wouldn't it be cheaper to just return the bloody string than do all this window calculating on each request? Not to mention memory manipulations with the array? The point is to protect the device from load but not returning the string isn't going to achieve that. Once we are over our rate we need to shut everything down! We are under attack and we don't want our battery destroyed and our other connections starved out. So at a minimum we need to call stop on advertising via thaliMobile! We then need to set a timer for some reasonable period of time and only start up again after that timer. For bonus points we should also hook network change events and maybe reset our timer earlier if we see that the network has changed. But, a persistent attacker will already know our address and use it to attack us so it's not enough to just shut down advertising, we have to also shut down listening as well! And we need to coordinate all of this with the layers above us so they understand that we are in an emergency situation!!!!!!

All of which is nuts. So, here is my proposal. What say we:
1. Rip out the rate limiting code.
2. Open a bug that we need to write a handbook on things to do to secure Thali and include in there that one really should use one of the rate limiting libraries and hook the rate exceeded notification (some of them have that) to shut everything down for some period of time and then restart later. We should then put this bug into the icebox because we aren't going to worry about it for March.

What say you?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-:-KBUkDUZb8c4oobW_iAv:-2058750686)*

yaronyg:
@juhanak and also make sure that before you make any changes you sync again with vNext. This should become compulsive on whatever branch you are on so you know you always have the latest. And please don't assign the bug to me back to review your responses until you have this passing in AppVeyor. Thanks!

juhanak:
Review status: 0 of 3 files reviewed at latest revision, 30 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVBqefsDyoAR8oVWqo:-KBgDw00SuMMeidTa2Tt:1463251153) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L1)):*
I think my jshint  was old and VSCode JSHint addon was using it because it was showing just few validation errors like lines that exceed 80 character max limit. Now I can see all errors and will fix them..

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVJFgSJykaP9YBT70R:-KBgF94CSlCqxFY7AHud:-684453256) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L2)):*
here should be an additional test where we try to hit the beacon endpoint before we call start on thaliNotificationServer. This should give us a 401. It proves that we don't register the route until start is called.
-Added test: Make an HTTP request to /NotificationBeacons before calling start

Also the tests below that call start with various legal values should all be extended to then call stop and do a beacon request and prove that they got the proper empty response.
-Changed test: Pass an empty array to ThaliNotificationServer.start
-Changed test: Make an HTTP request to /NotificationBeacons

There should also be at least one test where you actually check the beacon value you got back from the HTTP server and prove that it matches the beacon value you got when you generated the beacons with the beacon library.
-Changed test: "Make an HTTP request to /NotificationBeacons" to check the response.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 16 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVCrn_5GRt-yn-dazd:-KBgonq7cwPw6EacM0fQ:-973552330) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L16)):*
Ok. Changed localhost to 127.0.0.1. 

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 33 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVD4G0aOnY2fuLCi6I:-KBgoyJQEG5OIJrjHM1Q:-1506373321) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L33)):*
My stupid architecture was the reason. Internally function proxyquireThaliNotificationServer() changed the content of the globalVariables so I couldn't call the function when initializing globalVariables.

I changed the structure quite a lot. Now there is GlobalVariables class and all initialization of the generic parameters happens in its constructor or init function which I think is way more cleaner solution. 

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 43 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVDxhkcampDNME7Tpj:-KBgszjQVQXRtMIrLLYX:1189967937) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L43)):*
Good observation. I changed it to use Promise and closing happens with makeIntoCloseAllServer.closeAll. New implementation is in the GlobalVariables.prototype.kill function. Also listen method seems to be async so I changed it to use async pattern too.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 106 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVEO6yaDhJIc_wpgf1:-KBgtZ8Is_ZKWu-etCl3:1743462987) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L106)):*
I wanted to leave it there if I need to initialize ThaliNotificationServer differently. However there seems to be no need for this so it is now part of Init-functionality.  

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 109 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVEiQlfaRMbx33DkXl:-KBgts7zguUCP7LKkpVW:-507324092) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L109)):*
Fixed. All test which are true by default changed to real tests like "t.equal(error, null, 'no error');"

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 111 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVF0VZyzrXhoeTz6rV:-KBgu8S3t2HBO8eHo0AY:-1544711085) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L111)):*
ups.. removed

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 112 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVFWYEfQLVxOWX4mbP:-KBgvLc9tB0yAgeVFGmM:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L112)):*
Fixed.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 137 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVGsA_Lh0SZizjexyv:-KBgvVYwmeYq6rtLK4zK:44741100) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L137)):*
Ok. It's gone now. But I added a test that ensures globalVariables.spyStartUpdateAdvertisingAndListening.callCount is correct. 

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 156 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVCfPaM5sbl3WtqOeF:-KBhOlk2yps37mK2qJim:842132728) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L156)):*
Fixed. I was using a constant port 3001 for testing and it was bad idea when you have two simultaneous tests running. Now listen(0,..) takes care that a free port is always assigned for the test.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 161 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVH67_sgQLQFWi4IMS:-KBgwgqmIIVZfrEYNcTC:1619333749) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L161)):*
Yes indeed. I changed it. 

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 175 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVHW2hM13jLzpscDMS:-KBgxoeajM-rl-youl24:1924097123) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L175)):*
I moved server initialization to GlobalVariables.init function which is called in the begin each test.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 189 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVIbJJJhX395k3yh3j:-KBgxl-0cKqclxt4Ia2L:1758359711) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L189)):*
Fixed.

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 216 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVK7yXq4CTEz5skAw9:-KBhKJo_5PCtGQCUKCaA:467674166) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L216)):*
Very valid point. For instance rate like 100 msg/s would require 10ms accuracy with this test.. I removed throttling so this should not be a problem anymore. 

---

*[test/www/jxcore/bv_tests/testThaliNotificationServer.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVJwj_8O4fRD9rgFvO:-KBhKjnObTHKq5WZMWLv:1625806597) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/bv_tests/testThaliNotificationServer.js#L227)):*
Fixed. No good reason why it was like that. 

---

*[test/www/jxcore/lib/httpTester.js, line 17 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVAld1Nubi7zk29qM5:-KBhKp3oslgqd3LBoU0Q:1842356418) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/test/www/jxcore/lib/httpTester.js#L17)):*
Fixed. Actually changed runTest to not accept an array but just a single HTTP address because no more multiple HTTP requests are needed. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 29 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUks6GTKi0UUwRRI5n:-KBhSLyGC9vdkqiCmAmf:-1601141619) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L29)):*
Changed them to asserts. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 68 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV5MxbHVxzLl-65GwA:-KBhN6OZbf1K12m3L-2m:-1229359878) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L68)):*
Well, This problem is gone now but it was slow indeed. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 97 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUnAlbAPgsmAFheaW7:-KBhSlVBSJQLB-q69DzB:1262213168) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L97)):*
Changed. Empty arrays are now accepted as null and empty call throws an error. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 115 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUnTrhQnYAG_mvibtj:-KBhTlrJy8jp4IUk5RBJ:-609445857) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L115)):*
Changed.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 147 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUozsyDnmzTnCwHG1p:-KBhTzk4niBlZ5IDV33c:3436729) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L147)):*
Changed. But if one calls the start with an empty array as a first call then we don't call ThaliMobile.startUpdateAdvertisingAndListening but I guess it doesn't need to be called then.. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 177 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUptj9OfZ6lJR_w0yh:-KBhqfN2eCd1nLl9Hdf9:616106515) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L177)):*
Option 2 sounds logical. I implemented it. One thing left TODO is to document errors that stopAdvertisingAndListening may return.. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 182 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUpiz5u62SFRfkWbkc:-KBhqrqpGI4yPAoQPwbu:1409383609) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L182)):*
Yes, this line is gone.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 218 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUvjQCACmOoaUHvJXn:-KBhV7flFrMBI9t8EBgG:564196029) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L218)):*
Fixed. No reason to be there. It was left from my earlier implementation

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUwz9ECF_GHbLzMGh5:-KBhVRnWv0GUvUE3_9H7:-2128604931) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L227)):*
Yes it is gone now. For the record I did 10 seconds of searching and I think it was this one that I found ( https://github.com/nfriedly/express-rate-limit ) however it required access to app instance which thaliNotificationServer is not having so then I decided to write my own. But indeed it sounds weird that individual route handlers would control the rate.. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 239 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV2JLtBVQ-Qzrr0IF1:-KBhWgixRjwiSeht4jjw:999702174) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L239)):*
Yes, indeed they were too small. But it is no more problem of this class :)

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 240 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV2sO9mPGO1pKwbnf_:-KBhX3kIAMXc2DF0TQd6:963385097) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L240)):*
True a fixed array should perform better. But as I got rid of the whole implementation no actions related to this. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 243 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBV3pdGMwSvZ_IaKTCo:-KBhYZpR4f4UC0dzB0A2:1088250631) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L243)):*
Ok. I got it working now better. I think I'm able to catch all problems now. 

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 245 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBVLWytdBP53-4nfDre:-KBhYjcBY5MfHJuf_L8b:-381270328) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L245)):*
I agree with this. If there are other services than "/NotificationBeacons" it makes sense to have a one place where to control all rate limits. Probably in that object who owns the express app and starts listening etc.  

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585)*

yaronyg:
Reviewed 3 of 3 files at r2.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 147 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUozsyDnmzTnCwHG1p:-KBiO720TLMS0Go8hG5_:-1388319713) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L147)):*
NB: It is true that if the very first call to thaliNotificationServer.start is [] then we don't need to call startUpdateAdvertisingAndListening but wouldn't that still be cause by the if statement? previousPreambleAndBeacons would be null so would self._premableAndBeacons. So won't we not call startUPdateAdvertisingAndListening anyway?

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 227 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUwz9ECF_GHbLzMGh5:-KBiOu7sjLvmqtv47g_Y:-656248576) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L227)):*
NB: For whatever it's worth we (in thaliMobileNativeWrapper and thaliWifiInfrastructure) create the app servers so we could have handled that. but it's irrelevant now.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 152 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBiV0XAKkSGv9mpNr4k:-KBiV0XAKkSGv9mpNr4l:-453544495) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c9641dd0ab5c5aa86d4bd66506a7ea7355019adf/thali/NextGeneration/notification/thaliNotificationServer.js#L152)):*
It's generally considered better style to say if (self._firstStartCall) { return resolve(); } Then unindent the other clause.

But why even bother? It's explicitly legal to call stopAdvertisingAndListening even if start was never called so why not just call it even if _firstStartCall is true? No harm done, simpler code.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585)*

juhanak:
Review status: 2 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 147 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBUozsyDnmzTnCwHG1p:-KBl6iWi35De6TPQqZcG:517951295) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b1f3e0061373a059b023127dfdf7194a371ce216/thali/NextGeneration/notification/thaliNotificationServer.js#L147)):*
Right, so if the first call is [] then no ThaliMobile.startUpdateAdvertisingAndListening is called.

---

*[thali/NextGeneration/notification/thaliNotificationServer.js, line 152 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585#-KBiV0XAKkSGv9mpNr4k:-KBl1dZkJd-tOrcjFgDh:832794537) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c9641dd0ab5c5aa86d4bd66506a7ea7355019adf/thali/NextGeneration/notification/thaliNotificationServer.js#L152)):*
Fixed. I removed _firstStartCall check. Also improved the function documentation to reflect what the function currently does.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/585)*

PR #586, opened 2016-02-26
Fix the broken queue

yaronyg:
Our promise queue used a share array for all instances, that was just a bug. Now it uses a different array of promises for each queue instance. Thanks to @vjrantal  for fixing this.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/586)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -73,7 +73,9 @@ copying and pasting that Cordova development normally requires.
 To set up your desktop environment for development go to Thali_CordovaPlugin/thali/install and run 
 `jx npm run setupDesktop`.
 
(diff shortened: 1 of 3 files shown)


mpodwysocki:
Why aren't you using the [pique](https://www.npmjs.com/package/pique) library we already created from this project?  That was a bug I fixed a long time ago when this was published.

yaronyg:
Because pique brings in a lot more complexity in terms of build, update, etc. It is a bigger beast than just a handful of lines of code that we started with. So for now we stick with the handful of lines of code, bugs and all. :)

PR #604, opened 2016-03-08
Added release configuration file

jareksl:
Add an external config json file with the information about the dependency and the version number which will be used during installation to create the url with the correct version. The config file contains also the main settings for the thali repository.

[Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/604)


yaronyg:
We have to move the pointer to the Android binary as well. That is defined in Thali_CordovaPlugin/src/android/JXcore.gradle. We would need to change the install script to edit the gradle file on the fly during install to point to the version of the BTLibrary we want to use.

---

Reviewed 2 of 2 files at r1.

---

*[thali/install/install.js, line 101 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCMJSDZzg5SjNqMrlj4:-KCMJSDZzg5SjNqMrlj5:1417531928) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2a69ccbdc63162ce8731ed52d61edbdb73cb7188/thali/install/install.js#L101)):*
You don't need to do this. You can just remove the .catch clause entirely. If there is an error it will be propagated out of the function automatically.

---

*[thali/install/install.js, line 104 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCMJe_Ieorr7bFHrBe_:-KCMJe_JCdk53zGKdGg_:1691223062) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2a69ccbdc63162ce8731ed52d61edbdb73cb7188/thali/install/install.js#L104)):*
You don't need Promise.resolve. You can just have return JSON.parse(data).

A problem though is that JSON.parse can in fact throw an exception. So you need to wrap it in a try/catch and if the exception is thrown then you need to return a Promise.reject(new Error(err));

---

*[thali/install/install.js, line 328 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCMLL5K5aBp_o3PelhE:-KCMLL5K5aBp_o3PelhF:-1928066533) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2a69ccbdc63162ce8731ed52d61edbdb73cb7188/thali/install/install.js#L328)):*
I would recommend that you move this into the getReleaseConfig function. It's best when functions are self contained so that someone calling this function knows that they will either get a good value out or an error. They shouldn't have to check anything outside of the function.

---

*[thali/install/releaseConfig.json, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCMOsn1EMHYP2elTj4O:-KCMOsn1EMHYP2elTj4P:1367057710) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2a69ccbdc63162ce8731ed52d61edbdb73cb7188/thali/install/releaseConfig.json#L1)):*
In theory the package.json format is fully extensible. Instead of introducing a new config file in some random location why not create a new value 'thaliInstall' in the package.json in thali/package.json (not the one in install) and put the values there? That way all the versioning information for the production code is in one place.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-:-KCMMlIGpvxwxfqpJSRY:1817558465)*

yaronyg:
Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[scripts/androidBeforeCompile.js, line 48 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCrM8EC88UKoiRYpv9I:-KCrM8EDmGxTnc7SKRac:-1224754493) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/0ecc813a8b9986faf44ca03fd5de7bf663487167/scripts/androidBeforeCompile.js#L48)):*
We are flat out replacing the destination file. That is really dangerous since we might not be the only ones playing with this file. What would be more robust is either:

A) If the target file shouldn't exist (e.g. we are the ones creating it) then in your copySync submit the clobber: false option so we will fail if the file does exist. That way we will know something went wrong and can evaluate.
B) If the file should exist (e.g. Cordova or someone else creates it before we get there) then we should check that it exists and we should edit it rather than replace it since we could break something else if we just replace it.

---

*[scripts/androidBeforeCompile.js, line 49 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCrK_ONzi1EKTR2E4Bs:-KCrK_ONzi1EKTR2E4Bt:801978172) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/0ecc813a8b9986faf44ca03fd5de7bf663487167/scripts/androidBeforeCompile.js#L49)):*
This is dangerous because copy is async so it means that the call below to updateBtconnectorlibVersion will exit and go to the next statement before this copy is guaranteed to be finished. This leads to really unhappy race conditions.

That's why if you look, for example, at the replaceJXcoreExtension function we do the write file synchronously. That way we know that either it succeeded before any further code runs or we know we are out of luck and we can cleanly throw and correctly identify where it all hit the fan.

Now, it's absolutely true that in the normal course of Node.js events calling a synchronous function is a really bad idea since Node.js Javascript execution is single threaded and synchronous operations shut everything done. But in the case of a build script it's acceptable since there really isn't anything else we should be doing until we know if the copy is good.

So please change this to copySync

---

*[thali/install/install.js, line 347 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-KCrUBUJlJjw4Kv0xg0M:-KCrUBUKp5MAGm29u9xO:-653564894) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/0ecc813a8b9986faf44ca03fd5de7bf663487167/thali/install/install.js#L347)):*
Since you are effectively just returning the result from the previous then and since we call this result in the next then 'thaliCordovaPluginUnZipResult' it would be good to use the same naming for this then rather than 'directory'.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604)*

yaronyg:
@jareksl Also note that this PR is no longer up to date, you need to merge from vNext again

yaronyg:
:Shipit: (once you can sign the CLA :( )

---

Reviewed 2 of 2 files at r3, 2 of 2 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-:-KD-KYsn6G4SUpIzEZdv:-25696025)*

yaronyg:
:lgtm:

---

Reviewed 3 of 3 files at r6.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/604#-:-KE7VUrZp-oIjs0m6B-o:1417088721)*

PR #620, opened 2016-03-14
416: Implementation of ThalinotificationAction

juhanak:
This change is [Review on Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/620)


ThaliTester:
PR is added to the queue for testing as 1. task. (bf95981)

ThaliTester:
Test (Fail) 62754403 build is completed (bf95981)

See https://github.com/ThaliTester/TestResults/tree/62754403bf95981_First_implementation_of_thalinotificationAction_juhanak/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (19c4f54)

ThaliTester:
Test (Success) 62754403 build is completed (19c4f54)

See https://github.com/ThaliTester/TestResults/tree/6275440319c4f54_416__Implementation_of_ThalinotificationAction_juhanak/ for the logs

ThaliTester:
Test 6275440319c4f54(19c4f54) has failed

See https://github.com/ThaliTester/TestResults/tree/6275440319c4f54_416__Implementation_of_ThalinotificationAction_juhanak/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 1. task. (b276699)

ThaliTester:
Test (Success) 62754403 build is completed (b276699)

See https://github.com/ThaliTester/TestResults/tree/62754403b276699_416__Implementation_of_ThalinotificationAction_juhanak/ for the logs

ThaliTester:
Test 62754403b276699(b276699) has failed

See https://github.com/ThaliTester/TestResults/tree/62754403b276699_416__Implementation_of_ThalinotificationAction_juhanak/ for the fail logs

yaronyg:
@juhanak This is great work! Super solid, I especially liked the testing. You tested a bunch of scenarios that can be tricky but are very likely to happen in the real world. Good job!

---

Reviewed 4 of 7 files at r1, 1 of 1 files at r2, 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationAction.js, line 72 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCsV8GfipQK4v5ZE8dM:-KCsV8GfipQK4v5ZE8dN:-1903752474) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/test/www/jxcore/bv_tests/testThaliNotificationAction.js#L72)):*
We now support closeAllPromise which should make this code neater

---

*[test/www/jxcore/bv_tests/testThaliNotificationAction.js, line 157 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCsW0F0bVUQ8UFAgzz9:-KCsW0F1sv8ROO1143E7:-1680958817) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/test/www/jxcore/bv_tests/testThaliNotificationAction.js#L157)):*
Shouldn't you also test that you got back the right device key? In the new code you will need to test 3 values. See the testThaliNotificationBeacons for an example.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 67 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCr__2jA0vwiO2RmmeO:-KCr__2kA7HMt_cLd698:-1141113356) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L67)):*
This value isn't defined on an instance but instead is defined on the constructor object itself. E.g.

```Javascript
var Foo = function () {};
Foo.bar = 3;
var blah = new Foo();
console.log(Foo.bar);
console.log(blah.bar);
```

This will output 3 and undefined.

I think what you wanted was ThaliNotificationAction.prototype.eventEmitter. Try it btw, if you replace Foo.bar=3 above then you will get the opposite output.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 109 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrrrWREzk1lpH80Y40:-KCrrrWREzk1lpH80Y41:-172048611) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L109)):*
Nice catch. It is theoretically possible (depending on how the promise library is implemented) for someone to sneak in between the response to p and the start of this promise handler and that interloper could have called kill on the action. So you do, in fact, need to check. And returning null is also appropriate because if this was a start after kill then the base class implementation would have caught it. So this is right. I know this is just one line of code but this is exactly the kind f evil subtle bug that cooperative multitasking systems like Javascript introduce that can be awful to debug. So good catch!

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 110 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrsgYDMQM81xktlOJa:-KCrsgYDMQM81xktlOJb:-1315722465) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L110)):*
NB: From a style perspective 'return resolve(null);' is preferred. But to be clear it makes no different to functionality, it's just style.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 122 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrwyNy2ME4M-b6Xi10:-KCrwyNzW33x5lRNcbig:-1389752078) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L122)):*
Humor me please and stick in the family option set to 4. This is just paranoia on my part.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 125 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrvxeN-gblHT0G_je1:-KCrvxeN-gblHT0G_je2:-755280357) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L125)):*
I know I'm always in favor of factoring codes out to functions but in this case I think it actually makes things harder to follow given how small responseCallback is. It also requires you to shove resolve and reject onto self. Sticking things on self is always a bit sticky because it introduces more state that can change in places one can't immediately see. If it were me I would have at least turned _responseCallback into a function that takes resolve/reject as arguments and returns a function that http.request can use as a callback.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 133 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrwQrcY2_TU2Nb7ras:-KCrwQrdmymboG3gU9ZI:1582297540) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L133)):*
Ick.... this is basically secret passing state. The only way the person using the _responseCallback code even knows this is there is if they are in on the secret. Again, I would pass this as an argument to the responseCallback function so it's explicitly what state is being passed.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 142 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrxOVAEQ9woRRhhF9_:-KCrxOVAEQ9woRRhhF9a:343490997) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L142)):*
If we get an upgrade or connect then the connection will be closed. But do we get an error or some other indication that something has gone wrong so we at least know to signal that the request failed?

What if the other side just kills the connection dead without any notice? (I think you can simulate that with the destroy() method, but i may be wrong) What will happen? Do we have to wait until we time out?

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 171 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs383kCclyzfPZKmxj:-KCs383kCclyzfPZKmxk:-750518594) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L171)):*
Isn't _resolution just replicating _actionState?

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 202 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs3Sa2SC2-7bZIdfU3:-KCs3Sa2SC2-7bZIdfU4:1206506649) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L202)):*
The content-type header should be on the initial part of the request before the data so you can shortcut wrong responses by checking this immediately rather than waiting for the end event.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 204 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs3vRaiepXwBKVRIe-:-KCs3vRaiepXwBKVRIe0:1734101333) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L204)):*
I really don't think it's good to have functions that hide what they do. Also the function itself is really small. So my suggestion is to just get rid of the function and move the content directly into _responseCallback.

I realize this is a judgement call but my usual rule of thumb is if a function is less than say 50-60 lines then it can be together, otherwise it probably should be split apart.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 249 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs5f6y-LgF625jCYit:-KCs5f6y-LgF625jCYiu:-563988433) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L249)):*
Unfortunately this isn't enough information to use the notification. When receiving a notification the client has to create the PSK_Identity_Field which requires both the preamble as well as the specific beacon that matched the client. Now right away this introduces a problem which is that parseBeacons doesn't return the information we need!!!!!

The PSK_Identity_Field then has to be fed to the generatePSK function which we haven't actually implemented yet. Thankfully it's trivial.

We then need to return the PSK_Identity_Field and the psk secret in peerAdvertisesDataForUs. So we need those two values, along with unecryptedKeyId, to be returned in the emitted event.

So I'm going to submit to you a PR that extends parseBeacons to return the preAmble, unencryptedKeyId and encryptedBeaconKeyId. That is what the action needs to emit. Then when you write notification client you will use the new generatePskIdentityField and generatePskSecret functions to emit the final notification.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 282 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCsAV5MZ5bWORSCP60G:-KCsAV5NTnAvsW-dzMWG:579332278) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L282)):*
4k is a more reasonable value for now. I would set it to 1024 * 4. It can also be useful to specify the type in the name, e.g. MAX_CONTENT_SIZE_IN_BYTES.

---

*[thali/NextGeneration/notification/thaliNotificationCommon.js, line 1 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrYanh47Ys7zjGDd4_:-KCrYanh47Ys7zjGDd4a:-368213258) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationCommon.js#L1)):*
Do you want to stick this into thaliConfig? I'd ask @vjrantal if that makes sense. Because I can hear @vjrantal saying in my head "No, that file is just for user configuration parameters, not constants". Which I would agree with. But I don't want to see a proliferation of tiny files with just a single value either. Thoughts?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-:-KCsYFxljXCWzLnzIliD:506049322)*

ThaliTester:
PR is added to the queue for testing as 2. task. (643bc2c)

ThaliTester:
Test (Fail) 62754403 build is completed (643bc2c)

See https://github.com/ThaliTester/TestResults/tree/62754403643bc2c_416__Implementation_of_ThalinotificationAction_juhanak/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (91a6bae)

ThaliTester:
Test (Success) 62754403 build is completed (91a6bae)

See https://github.com/ThaliTester/TestResults/tree/6275440391a6bae_416__Implementation_of_ThalinotificationAction_juhanak/ for the logs

juhanak:
Review status: 1 of 10 files reviewed at latest revision, 15 unresolved discussions.

---

*[test/www/jxcore/bv_tests/testThaliNotificationAction.js, line 72 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCsV8GfipQK4v5ZE8dM:-KCtY_GESSmaJcfRcbeF:-539943903) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/test/www/jxcore/bv_tests/testThaliNotificationAction.js#L72)):*
Thanks. Instead of 10 lines now it takes 3 lines of code!

---

*[test/www/jxcore/bv_tests/testThaliNotificationAction.js, line 157 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCsW0F0bVUQ8UFAgzz9:-KCve-JH-tThLpag4ige:-1582986446) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/test/www/jxcore/bv_tests/testThaliNotificationAction.js#L157)):*
Sure, This is now tested in: Test BEACONS_RETRIEVED_AND_PARSED locally

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 67 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCr__2jA0vwiO2RmmeO:-KCuPQPR36G2jpWqDSVf:1904293775) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L67)):*
Thanks, good point! I was trying to quickly finish that change yesterday and here is the result..

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 109 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrrrWREzk1lpH80Y40:-KCucZtuQ_BvVR2r2vTI:-1577226821) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L109)):*
Thanks! I changed it to: return resolve(null);.  

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 110 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrsgYDMQM81xktlOJa:-KCt_vDZGoDKqgkRsGBy:-782529931) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L110)):*
Changed. It indeed makes it cleaner. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 122 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrwyNy2ME4M-b6Xi10:-KCta3W74RsAx2c4r-0L:1827872899) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L122)):*
sure added it. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 125 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrvxeN-gblHT0G_je1:-KCtsWQMyS4hyvt4dJWe:-1819857863) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L125)):*
I think I can't avoid having _resolve and _reject as members of my class. I'll need them when kill function calls _complete. However I changed _responseCallback to take a self as an input parameter and return a callback that http.request understands. It is a little bit cleaner now.

 I'd like to avoid having full responseCallback inside start as it is already getting quite big. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 133 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrwQrcY2_TU2Nb7ras:-KCu6B1AgujWM4rZW1cO:-843309223) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L133)):*
This is now fixed as mentioned above. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 142 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrxOVAEQ9woRRhhF9_:-KCuFkHXnXgVCGyo7bb-:813998381) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L142)):*
I added new test: 
'Close the server socket while the client is waiting a response from the server'. This will cause '-2 socket hang up' to be passed to: self._httpRequest.on('error', function ()

It will be handled fine in the notificationAction. I did not try to force close the socket from the client side but if some 'outsider' closes the socket that HTTP client is internally using it must cause some similar error, right?

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 171 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs383kCclyzfPZKmxj:-KCuIGDXC4Ion_GyXg1C:1804473574) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L171)):*
yes in this case. I changed it to: 
return this.getActionState() === PeerAction.actionState.KILLED;

Unfortunately I can't get completely rid of the _resolution. It was the simplest way for me to ensure that in the _complete function  emit is getting called only once. There are some special cases we call _complete two times:
1.when the kill is called. It calls _complete which calls http.abort which causes http.error callback which calls _complete again
2. When we exceed content size in _responseCallback it will call _complete which then calls http.abort which causes http.error callback which calls _complete again.. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 202 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs3Sa2SC2-7bZIdfU3:-KCtyxn4bMLbzc_6e-Xj:-2062131419) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L202)):*
Makes sense. Moved it there!

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 204 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs3vRaiepXwBKVRIe-:-KCtz5DlnBwOdDYrzmB-:885614409) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L204)):*
You're right it fits well to _responseCallback.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 249 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCs5f6y-LgF625jCYit:-KCveRt4-3SK3eE4BHTr:-2098062337) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L249)):*
OK! This is now added into the latest release

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 282 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCsAV5MZ5bWORSCP60G:-KCtzFkYTM_m62mv1HMd:-1159710788) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L282)):*
Sure. I changed it to 1024*4. 

---

*[thali/NextGeneration/notification/thaliNotificationCommon.js, line 1 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrYanh47Ys7zjGDd4_:-KCuYritm-gKWReyvIcP:1819032759) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationCommon.js#L1)):*
I'll move it to thaliConfig. @vjrantal agreed with it. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620)*

ThaliTester:
Test 6275440391a6bae(91a6bae) has failed

See https://github.com/ThaliTester/TestResults/tree/6275440391a6bae_416__Implementation_of_ThalinotificationAction_juhanak/ for the fail logs

yaronyg:
Reviewed 1 of 2 files at r4, 3 of 6 files at r5, 6 of 6 files at r6.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 142 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrxOVAEQ9woRRhhF9_:-KD-LZ8ORmN_h6LDkzfz:-722359408) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L142)):*
Some error will happen but what error and will we handle it properly? I only care because connection failures will be the rule for our scenarios, not the exception.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 125 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-ftAsXkvXGlMKZh92:-KD-ftAsXkvXGlMKZh93:-1411085755) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L125)):*
Instead of passing self to _responseCallback it would probably be more idiomatic to issue self._responseCallback.bind(self). This is built into the Javascript language and does the same thing.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 133 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-MahtVLS2Bum7rE2B:-KD-MahuYpIWOiSffON6:-196686706) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L133)):*
NB: It would be best if this wasn't left as a dangling comment

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 160 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-clRNpVXEph0_x9Zh:-KD-clROv3ff7VBvCUSk:1502865571) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L160)):*
I would have expected this line to come before the previous line so we are sure the state is marked as dead. That way if anything asynchronous happens they will be cut off by the dead state.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 170 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-dD1OCGDRaU0JuZTq:-KD-dD1OCGDRaU0JuZTr:-1920769815) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L170)):*
Why have a whole method for this? Why not just replace any call to this method with the one line below? I actually checked, at least in this file you seem to call the function exactly once anyway.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 184 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-g4u0hZz-9htDb0to:-KD-g4u0hZz-9htDb0tp:-1918094453) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L184)):*
If you take my proposal above then instead of passing in self you would instead just use 'this' knowing it had been properly bound via the bind call. Although it's also fine to define self as a var at the top of this function and assign it this.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620)*

ThaliTester:
PR is added to the queue for testing as 2. task. (9ea0a99)

juhanak:
Review status: 8 of 10 files reviewed at latest revision, 6 unresolved discussions.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 142 \[r3\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KCrxOVAEQ9woRRhhF9_:-KD2vpW4EjJtDH14JbNX:195507104) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b2766997eb2f5cf049d82f717e1ef63c5b441c83/thali/NextGeneration/notification/thaliNotificationAction.js#L142)):*
Added test: 'Close the client socket while the client is waiting a response from the server. Local test.'

It gives following error: 
{ [Error: socket hang up -1] code: 'ECONNRESET' }

I'm not specifically handling different type of errors. All they 'catched' here
---
      self._httpRequest.on('error', function () {
        self._complete(
          ThaliNotificationAction.ActionResolution.NETWORK_PROBLEM,
          null, 'Could not establish TCP connection');
      });
---

And they will cause NETWORK_PROBLEM event and ThalinotificationAction.start will return rejected promise with 'Could not establish TCP connection' 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 125 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-ftAsXkvXGlMKZh92:-KD3AMtClPhobfOFoJDo:2144993512) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L125)):*
Changed to bind(self).

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 133 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-MahtVLS2Bum7rE2B:-KD2vnW8UC5M4iF5cWu9:1675637715) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L133)):*
removed.

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 160 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-clRNpVXEph0_x9Zh:-KD2QaQOQHLCG9SrYWFq:-1301118115) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L160)):*
Thanks for pointing out. I was trying to get rid of '_resolution'  so I changed logic inside _complete function from: 
if (!this._resolution) {
to
if(!this.isKilled()) 

and for that to work I needed to call _complete first. As there were only synchronous stuff happening inside _complete it would have not caused problems. (I was thinking to use it like destructor where child is killed first..).

There were some other cases which stopped me using isKilled in the _complete function so I'll just change the order back as you mentioned. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 170 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-dD1OCGDRaU0JuZTq:-KD2xvUhfJax2OPZTnR4:207181691) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L170)):*
As explained above I was planning to use it inside _complete at least. But now it is used only once so I'll do as you suggested and remove it. 

---

*[thali/NextGeneration/notification/thaliNotificationAction.js, line 184 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-KD-g4u0hZz-9htDb0to:-KD2y7vmJa8GgquDOnJ1:1462269509) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/91a6baebc49883a194d9928a343d2a3ab4974217/thali/NextGeneration/notification/thaliNotificationAction.js#L184)):*
Sure I'll fixed it. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620)*

ThaliTester:
PR is added to the queue for testing as 3. task. (d2f0346)

ThaliTester:
Test (Success) 62754403 build is completed (9ea0a99)

See https://github.com/ThaliTester/TestResults/tree/627544039ea0a99_416__Implementation_of_ThalinotificationAction_juhanak/ for the logs

ThaliTester:
Test (Success) 62754403 build is completed (d2f0346)

See https://github.com/ThaliTester/TestResults/tree/62754403d2f0346_416__Implementation_of_ThalinotificationAction_juhanak/ for the logs

ThaliTester:
Test 627544039ea0a99(9ea0a99) has failed

See https://github.com/ThaliTester/TestResults/tree/627544039ea0a99_416__Implementation_of_ThalinotificationAction_juhanak/ for the fail logs

ThaliTester:
Test 62754403d2f0346(d2f0346) has failed

See https://github.com/ThaliTester/TestResults/tree/62754403d2f0346_416__Implementation_of_ThalinotificationAction_juhanak/ for the fail logs

yaronyg:
:lgtm:

---

Reviewed 1 of 2 files at r7, 1 of 1 files at r8.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/620#-:-KD4p6w5OsFeUyoSxzkI:1417088721)*

PR #646, opened 2016-03-19
Fix less travelled paths

tobybrad:
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/646)

### DIFF
--- a/src/ios/MultipeerConnectivity/THEMultipeerClient.m
+++ b/src/ios/MultipeerConnectivity/THEMultipeerClient.m
@@ -229,7 +229,13 @@ static const int MAX_PENDING_REVERSE_CONNECTIONS = 64; // Don't allow more than 
     clientSession = (THEMultipeerClientSession *)p;
     if (clientSession)
     {
-      if (![clientSession visible])
+      NSString *sessionGeneration = [THEMultipeerPeerSession peerGenerationFromPeerIdentifier: [clientSession remotePeerIdentifier]];
+      NSString *requestedGeneration = [THEMultipeerPeerSession peerGenerationFromPeerIdentifier: peerIdentifier];
+
+      NSLog(@"sessionGeneration: %d requestedGeneration: %d", [sessionGeneration intValue], [requestedGeneration intValue]);
+      
+      // Peer is unreachable if not visible or we're being asked for a stale generation
+      if (![clientSession visible] || [sessionGeneration intValue] > [requestedGeneration intValue])
       {
         success = NO;
         NSLog( @"client: connect: unreachable %@", remotePeerUUID);
@@ -277,7 +283,7 @@ static const int MAX_PENDING_REVERSE_CONNECTIONS = 64; // Don't allow more than 
             [_nearbyServiceBrowser invitePeer:[clientSession remotePeerID]
                                     toSession:[clientSession session]
                                   withContext:contextData
-                                      timeout:30];
+                                      timeout:10];
 
             success = YES;
           }
@@ -288,10 +294,16 @@ static const int MAX_PENDING_REVERSE_CONNECTIONS = 64; // Don't allow more than 
             success = NO;
           }
         }
+        else if ([clientSession connectionState] == THEPeerSessionStateConnecting)
+        {
+          NSLog(@"client: already connecting to %@", peerIdentifier);
+          connectCallback(@"Already connecting", 0);
+          success = NO;
+        }
         else
         {
-          NSLog(@"client: already connect(ing/ed) to %@", peerIdentifier);
-          connectCallback(@"Already connect(ing/ed)", 0);
+          NSLog(@"client: already connected to %@", peerIdentifier);
+          connectCallback(@"Already connected", 0);
           success = NO;
         }
       }
@@ -433,6 +445,8 @@ static const int MAX_PENDING_REVERSE_CONNECTIONS = 64; // Don't allow more than 
       if ([remotePeerIdentifier compare:[clientSession remotePeerIdentifier]] != NSOrderedSame)
       {
         NSLog(@"client: found updated peer: %@", remotePeerIdentifier);
+        // Updting the peer id implies declaring the old one no longer reachable
+        [_multipeerDiscoveryDelegate didLosePeerIdentifier:[clientSession remotePeerIdentifier]];
         [clientSession updateRemotePeerIdentifier:remotePeerIdentifier];
         previouslyVisible = FALSE;
       }
--- a/src/ios/MultipeerConnectivity/THEMultipeerClientSession.m
+++ b/src/ios/MultipeerConnectivity/THEMultipeerClientSession.m
@@ -137,7 +137,10 @@
   {
     NSLog(@"client session: onLinkFailure: %@", [self remotePeerUUID]);
 
-    assert([self connectionState] != THEPeerSessionStateNotConnected);
+    if ([self connectionState] == THEPeerSessionStateNotConnected)
+    {
+      return;
+    }
 
     THEPeerSessionState prevState = [self connectionState];
     [self disconnect];
--- a/src/ios/MultipeerConnectivity/THEMultipeerManager.m
+++ b/src/ios/MultipeerConnectivity/THEMultipeerManager.m
@@ -344,7 +344,7 @@
   NSLog(@"multipeer manager: start client restart timer: %p", self);
 
   dispatch_async(dispatch_get_main_queue(), ^{
-    _clientRestartTimer = [NSTimer scheduledTimerWithTimeInterval:20
+    _clientRestartTimer = [NSTimer scheduledTimerWithTimeInterval:10
                                                            target:self
                                                          selector:@selector(restartClient:)
                                                          userInfo:nil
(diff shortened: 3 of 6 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (c6bc463)

ThaliTester:
Test (Fail) 63458193 build is completed (c6bc463)

See https://github.com/ThaliTester/TestResults/tree/63458193c6bc463_Fix_less_travelled_paths_tobybrad/ for the logs

ThaliTester:
Test (Fail) 63458193 build is completed (c6bc463)

See https://github.com/ThaliTester/TestResults/tree/63458193c6bc463_Fix_less_travelled_paths_tobybrad/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (508fffe)

ThaliTester:
Test (Fail) 63458193 build is completed (508fffe)

See https://github.com/ThaliTester/TestResults/tree/63458193508fffe_Fix_less_travelled_paths_tobybrad/ for the logs

ThaliTester:
Test (Fail) 63458193 build is completed (508fffe)

See https://github.com/ThaliTester/TestResults/tree/63458193508fffe_Fix_less_travelled_paths_tobybrad/ for the logs

vjrantal:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMvi6PADlhEUbDJpYI:-KDMvi6PADlhEUbDJpYJ:-1570752065) (no related file):*
This test still hangs for me locally and in both CI environments. We now support running individual tests in coordinated desktop environment so you can reproduce this like `jx runCoordinatedTests.js --filter bv_tests/testThaliMobileNative.js`. Please check first locally that this test passes before pushing the next commit.

The latest commit seemed to be about making the data sending test cross-platform, but just wanted to point out that without the changes done in `vNext_vjrantal_500_tobe_fix_ios` this test already worked on desktop and on Android.

There is some whitespace and linting issues, but I didn't comment on those since I think most important would be to get the basic logic right. I am fine merging with the linting issue since I can easily fix them afterwards.

---

*[src/ios/MultipeerConnectivity/THEMultipeerClient.m, line 300 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMsgGi9Z9yR4CeK0xh:-KDMsgGi9Z9yR4CeK0xi:-1501902908) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/src/ios/MultipeerConnectivity/THEMultipeerClient.m#L300)):*
If you want to change the error message, it must be also changed in specification and Android implementation, because otherwise, it is hard to write cross-platform node code. If not changing everywhere, I suggest we don't change here either.

---

*[test/www/jxcore/bv_tests/testThaliMobileNative.js, line 145 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMt8-zlHFv7W5WrEPh:-KDMt8-zlHFv7W5WrEPi:1183333173) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/test/www/jxcore/bv_tests/testThaliMobileNative.js#L145)):*
Why would a peer become unreachable in an environment where the devices are next to each other (which is the case when we run the tests)? If this happens, I'd fail the test overall since sounds like an error scenario.

Now that there is a return here, doesn't it mean that if this code path would get run, we would not call successCb nor would be schedule a connect retry so wouldn't the test just hang?

---

*[test/www/jxcore/bv_tests/testThaliMobileNative.js, line 147 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMtY9cmcocZNNq5SDA:-KDMtY9cmcocZNNq5SDB:-1402160879) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/test/www/jxcore/bv_tests/testThaliMobileNative.js#L147)):*
In what scenarios the peer would already be connected while this test runs? The test starts from a stopped state and we first call all the starts. The, we try to connect to the first peer we see. Why would that peer already be connected? Sounds like an error if we end up to this code path.

---

*[test/www/jxcore/bv_tests/testThaliMobileNative.js, line 220 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMtx3BfMIOb2sXyY7T:-KDMtx3BfMIOb2sXyY7U:339164265) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/test/www/jxcore/bv_tests/testThaliMobileNative.js#L220)):*
If we get a failure to connect after 10 retries, I think the test should overall be marked as a failure so I'd remove this check.

---

*[test/www/jxcore/bv_tests/testThaliMobileNative.js, line 235 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMu7WWDMySPhc1DFsM:-KDMu7WWDMySPhc1DFsN:2065525912) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/test/www/jxcore/bv_tests/testThaliMobileNative.js#L235)):*
Now that the connecting-check was turned into a connected-check, here is what I think might happen.

We start concurrently connecting to every peer we discover. For example, if in CI there would be more than two Thali apps running. If all would go great and all connect-attempts would succeed, the test would fail, because t.end() would be called more than once.

I suggest we keep the test logic like this:
- We do peer discovery as long as we have found a peer
- We try to connect to only that peer max 10 times and if that fails, we mark the test overall as a failure

---

*[test/www/jxcore/bv_tests/testThaliMobileNative.js, line 271 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMuwj_94vqcC7TjVwD:-KDMuwj_94vqcC7TjVwE:1633943164) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/test/www/jxcore/bv_tests/testThaliMobileNative.js#L271)):*
This isn't used anywhere.

---

*[test/www/jxcore/bv_tests/testThaliMobileNative.js, line 357 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMv76Ts_hN6VdBwOy2:-KDMv76Ts_hN6VdBwOy3:848445480) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/test/www/jxcore/bv_tests/testThaliMobileNative.js#L357)):*
Same as earlier comment.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646)*

yaronyg:
Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDMvi6PADlhEUbDJpYI:-KDPpjePEuyRdKC2Pju1:-1989447248) (no related file):*
I don't get a hang, what I get is an infinite list of identified peers creating an infinite number of connections. I think this is largely because of a bug in the wifiBasedNativeMock. So I'm sending this back to @tobybrad to review and respond to the comments but I'm going to go fix the native mock and see if that fixes the issues with the tests on the desktop. For now @tobybrad should just worry about the tests passing on iOS.

---

*[src/ios/MultipeerConnectivity/THEMultipeerManager.m, line 347 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646#-KDPWPYFzxV5WY5bWP3S:-KDPWPYFzxV5WY5bWP3T:184465380) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/508fffe7af6717b5de9db4897ee153f79dee731a/src/ios/MultipeerConnectivity/THEMultipeerManager.m#L347)):*
So I vaguely remember that we had issues with announcements just sort of disappearing and we needed to restart the announcements on a fairly frequent basis to keep things moving along. I also remember that this wasn't supposed to disturb any existing MCSessions and that we would use the same peerIdentifier across different starts so that other clients would realize that this is, in fact, the same client that they saw previously. So are you saying that the announcements are so unstable that we have to restart every 10 seconds?

But how does this lead to the connection issues that @vjrantal raised in the js code? Shouldn't these restarts be effectively invisible to the node.js layer outside of getting an additional stream of peer updates that always have the same peer ID?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/646)*

ThaliTester:
PR is added to the queue for testing as 1. task. (508fffe)

ThaliTester:
Test (Fail) 63458193 build is completed (508fffe)

See https://github.com/ThaliTester/TestResults/tree/63458193508fffe_Fix_less_travelled_paths_tobybrad/ for the logs

ThaliTester:
Test (Fail) 63458193 build is completed (508fffe)

See https://github.com/ThaliTester/TestResults/tree/63458193508fffe_Fix_less_travelled_paths_tobybrad/ for the logs

yaronyg:
@baydet This is now morphing per the mail thread into figuring out how to diff this with the version I talked about in email and then decide what to keep and what to drop.

yaronyg:
I'm closing this since we are rewriting the code base

Issue #555, opened 2016-02-19
Write comprehensive unit tests for Android framework


yaronyg:
We have no formal unit tests what so ever. We need to fix that.

PR #762, opened 2016-07-15
Native UT for cordova plugin (android) #555

jareksl:
The test classes are in the node subdirectory of src/android/test directory. There are also few mocks there. There is also the modification of the build script and androidBeforeCompile.js allowing to copy UT files in case when the "./build.sh UT" is issued. The test runner "ThaliTestRunner" is registered to be executed in the JXcoreExtension. If the class RegisterExecuteUT exists, which means that all test classes are copied, the ExecuteNativeTests method is registered in the jxcore and executed.  At the end there is also modification of the app.js allowing to execute UT before the e2e test. The modification of the appveyor.yml was, because there are some issues when using the latest cordova version.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/762)


ThaliTester:
PR is added to the queue for testing as 1. task. (35c04ce)

ThaliTester:
Test (Fail) 77622416 build is completed (35c04ce)

See https://github.com/ThaliTester/TestResults/tree/7762241635c04ce_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

jareksl:
*[src/android/test/io/jxcore/node/jxcore.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj85eykrFdadD57nYw:-KMj85eykrFdadD57nYx:-1260894603) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/jxcore.java#L1)):*
> ```Java
> // License information is available from LICENSE file
> ```

This file should be removed, the whole modification is now done in the JXcoreExtension

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

jareksl:
*[scripts/androidBeforeCompile.js, line 54 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj8dDSs9oQI2RF3vRP:-KMj8dDSs9oQI2RF3vRQ:1797345465) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L54)):*
> ```JavaScript
> };
> 
> var replaceJXCore = function (appRoot) {
> ```

This function should be removed, as the jxcore is not replaced anylonger

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

yaronyg:
Reviewed 14 of 36 files at r1.
Review status: 14 of 36 files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[scripts/androidBeforeCompile.js, line 54 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj8dDSs9oQI2RF3vRP:-KMl0n-guB8Kblgk4OhM:-461457794) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L54)):*
<details><summary><i>Previously, jareksl (Jarosław Słota) wrote…</i></summary>
> This function should be removed, as the jxcore is not replaced anylonger
</details>
I don't remember anymore what it was that we were fixing but what changed that we don't need to fix it anymore?

---

*[scripts/androidBeforeCompile.js, line 77 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl13XR3QhFIOTiX8Xn:-KMl13XSWmppS6Wrg-O9:407746474) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L77)):*
> ```JavaScript
>   var targetFile = path.join(appRoot, 'platforms/android/src/io/jxcore/node');
>   try {
>     fs.copySync(sourceFile, targetFile);
> ```

There is the same code structure here three times, can we replace this and the next two functions with a call to a single function that takes the source, target and some kind of error message as args and then does the copySync and handles the error by exiting?

---

*[scripts/androidBeforeCompile.js, line 146 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl1VFGpF0-Yli5n0HH:-KMl1VFHUOlCMeiO7dS5:392110975) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L146)):*
> ```JavaScript
> var copyTestFiles = function (appRoot) {
>   try {
>     var st = fs.lstatSync('platforms/android/unittests');
> ```

Please use a more descriptive name than 'st'

---

*[src/android/java/io/jxcore/node/IncomingSocketThread.java, line 42 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl2EEFSFV_JClCd9Ak:-KMl2EEGoVXZVuX3_nx0:-279776276) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/IncomingSocketThread.java#L42)):*
> ```Java
>      * @throws IOException Thrown, if the constructor of the base class, SocketThreadBase, fails.
>      */
>     public IncomingSocketThread(BluetoothSocket bluetoothSocket, Listener listener,
> ```

Is this constructor just for testing purposes? If so we should mark it as such in the docs.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 81 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl30wx6CHYPFcOsDzm:-KMl30wx6CHYPFcOsDzn:-963336023) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/JXcoreExtension.java#L81)):*
> ```Java
>      */
> 
>     private static void RegisterUTExecuteMethod() {
> ```

I'm really not happy with having this code in our mainline JXcoreExtension. If I'm reading this right then this version of JXcoreExtension would be in our production code and we are just depending on the class detector failing to make sure it doesn't run. But that just isn't good coding practice. We don't mix our test code with our production code. 

In fact, why is this here at all? I thought the UT tests are pure native Android. Why do we need JXcoreExtension? Please explain on our next stand up. If we do need to call out to this then I would rather we monkey patched this file as part of the test install environment.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 99 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl2qLMkf7zGy0_8LJu:-KMl2qLMkf7zGy0_8LJv:962742703) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/JXcoreExtension.java#L99)):*
> ```Java
>     private static Class<?> RetrieveRegisterUTClass() {
>         try {
>             return Class.forName("com.test.thalitest.RegisterExecuteUT");
> ```

Shouldn't we put the names in our own actual DNS registered namespace? Something like org.thaliproject.java.thalitest.RegisterExecuteUT?

---

*[src/android/java/io/jxcore/node/ListenerOrIncomingConnection.java, line 59 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl3the3Hzak-q3NpCL:-KMl3the3Hzak-q3NpCM:-2060682334) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/ListenerOrIncomingConnection.java#L59)):*
> ```Java
> 
>     public int getClientPortNumber() {
>         return mClientPortNumber;
> ```

If we lived with a bug this bad do we really need this method?

---

*[src/android/java/io/jxcore/node/OutgoingSocketThread.java, line 39 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl3zXf7-MhWOynk6uz:-KMl3zXf7-MhWOynk6v-:-507865954) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/OutgoingSocketThread.java#L39)):*
> ```Java
>      * @param listener        The listener.
>      * @param inputStream     The InputStream.
>      * @param outputStream    The OutputStream.
> ```

I'm guessing this is another test specific method? If so please mark it as such.

---

*[src/android/java/io/jxcore/node/SocketThreadBase.java, line 63 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl4CT3UuP45dKKetaZ:-KMl4CT3UuP45dKKeta_:-230909707) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/SocketThreadBase.java#L63)):*
> ```Java
>      * @param outputStream    The OutputStream.
>      */
>     public SocketThreadBase(BluetoothSocket bluetoothSocket, Listener listener,
> ```

I promise, this will be the last time I make this specific comment. But anytime we put in a constructor or method just for testing purposes we should prominently mark it as such so people trying to read the code don't get confused. We will also eventually need to exclude these methods when we do code coverage tests.

---

*[src/android/test/com/test/thalitest/MainActivity.java, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl4fUI_3Lturj0v5jG:-KMl4fUI_3Lturj0v5jH:1867604115) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/com/test/thalitest/MainActivity.java#L2)):*
> ```Java
> /*
>        Licensed to the Apache Software Foundation (ASF) under one
> ```

Um... this isn't our copyright notice. We use MIT. Why is this here?

---

*[src/android/test/com/test/thalitest/ThaliTestSuite.java, line 19 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl5QGMiTMKuGjPEbV6:-KMl5QGMiTMKuGjPEbV7:1949533753) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/com/test/thalitest/ThaliTestSuite.java#L19)):*
> ```Java
> 
> @RunWith(Suite.class)
> @Suite.SuiteClasses({
> ```

Is there no way to generate this list dynamically? The reason I ask is that it wouldn't be the 1st or 10th time a framework that required explicitly listing tests resulted in a situation where people adding new tests didn't realize they had to register them and they never get run and nobody every notices. :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

yaronyg:
Reviewed 22 of 36 files at r1.
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-eQloYB63_LcU01Un:-KN-eQlpYBpcruS00UP6:-2059307203) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L1)):*
> ```Java
> package io.jxcore.node;
> ```

Have y'all tried to run any kind of code coverage statistics so we can see how much of the code we are actually covering?

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 74 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-bHyTWlMafEiuBPd8:-KN-bHyTWlMafEiuBPd9:2088019344) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L74)):*
> ```Java
>     @After
>     public void tearDown() throws Exception {
>         mConnectionHelper.killAllConnections();
> ```

Please put this code and its equivalent version in tearDownAfterClass into a single function and call that.

Or, alternatively, why are we repeating the code at all? If we are going to shut everything down after every test then why do we need to do it at the end of the entire class's test run?

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 659 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-dBRzaVZfYoCUakeG:-KN-d8O6SuJHuDJxsNaw:-1115305113) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L659)):*
> ```Java
>     @Test
>     public void testOnPeerUpdated() throws Exception {
>         assertThat(true, is(true));
> ```

Yes, it's true that true is true but I"m guessing something is actually supposed to be here?

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 672 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-dTGDJiFHQLsUBZMi:-KN-dTGDJiFHQLsUBZMj:741060480) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L672)):*
> ```Java
>         mIncomingSocketThreadMock.setPeerProperties(mPeerProperties);
>         mConnectionModel.addConnectionThread(mIncomingSocketThreadMock);
>         mConnectionHelper.onPeerLost(mPeerProperties);
> ```

Is this actually checking that onPeerLost is called?

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 684 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-decZv3merKRZGmtz:-KN-dec_XdrSysqW8P4o:-347321759) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L684)):*
> ```Java
>     public void testOnProvideBluetoothMacAddressRequest() throws Exception {
>         thrown.expect(UnsupportedOperationException.class);
>         mConnectionHelper.onProvideBluetoothMacAddressRequest("1111");
> ```

Shouldn't we be checking that this one and the next two do call the handlers when they are supposed to be?

---

*[src/android/test/io/jxcore/node/IncomingSocketThreadTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-gnm_DGinUnOB36YI:-KN-gnm_DGinUnOB36YJ:-1769915933) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/IncomingSocketThreadTest.java#L1)):*
> ```Java
> package io.jxcore.node;
> ```

The most interesting part of this class is the run method and it doesn't seem to be tested

---

*[src/android/test/io/jxcore/node/jxcore.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj85eykrFdadD57nYw:-KN-h3djb-jekaHZ89vd:-1010244026) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/jxcore.java#L1)):*
<details><summary><i>Previously, jareksl (Jarosław Słota) wrote…</i></summary>
> This file should be removed, the whole modification is now done in the JXcoreExtension
</details>
Is this still the case? I'm ignoring this file for the moment until I get clarification. Thanks!

---

*[src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-jYraByp496_bmxHl:-KN-jYraByp496_bmxHm:-528495908) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java#L1)):*
> ```Java
> package io.jxcore.node;
> ```

Run is where all the fun is but it isn't being tested!

---

*[src/android/test/io/jxcore/node/StreamCopyingThreadTest.java, line 71 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-keZpPFaNCSpJgsQ_:-KN-keZpPFaNCSpJgsQa:1404120076) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/StreamCopyingThreadTest.java#L71)):*
> ```Java
> 
>     @Test
>     public void testRun() throws Exception {
> ```

Given how earth shatteringly important this code is to everything we do wouldn't it make sense to test that we can handle more than just copying a single string? I would want to test several different passes of data just to make sure the engine works right.

---

*[src/android/test/io/jxcore/node/StreamCopyingThreadTest.java, line 103 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-kYiydb7FFVUHzS61:-KN-kYiydb7FFVUHzS62:549973867) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/StreamCopyingThreadTest.java#L103)):*
> ```Java
>         @Override
>         public void onStreamCopyError(StreamCopyingThread who, String errorMessage) {
> ```

Shouldn't we get unhappy if this is called? And actually shouldn't we test how we behave in that case?

---

*[test/www/jxcore/UnitTest_app.js, line 33 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-lH3qn6uvcT4fVpRs:-KN-lH3qn6uvcT4fVpRt:-2066818728) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/test/www/jxcore/UnitTest_app.js#L33)):*
> ```JavaScript
>       });
>     })
>       .then(function (result) {
> ```

NB: Please indent this code one, out style is to put .then on the new line at the same level as the previous promise code.

---

*[test/www/jxcore/UnitTest_app.js, line 45 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-lUVlUw4I2mncxQXO:-KN-lUVlUw4I2mncxQXP:-196561628) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/test/www/jxcore/UnitTest_app.js#L45)):*
> ```JavaScript
>         }
> 
>         Promise.all(promiseList)
> ```

I"m confused why the previous code was shoved between starting the radios if they aren't on and checking if the radios actually go started. Those are two pieces of heavily related code. I suspect the previous promise which runs UT should be first and in fact should probably run before the getNetworkStatus call?

---

*[thali/install/setUpTests.sh, line 66 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-ljPcLcpThfHHvEoK:-KN-ljPcLcpThfHHvEoL:608696052) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/thali/install/setUpTests.sh#L66)):*
> ```Shell
> cp -v $1 app.js
> 
> # In case of UT crete a file
> ```

NB: /crete/create

---

*[thali/install/setUpTests.sh, line 67 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-lvx_WEL1yWfi5KEO:-KN-lvxac6EGX3rBag3q:110906531) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/thali/install/setUpTests.sh#L67)):*
> ```Shell
> 
> # In case of UT crete a file
> if [ $2 == "UT" ]; then
> ```

Rather than having two if statements how about just using OR?

---

*[thali/install/setUpTests.sh, line 68 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-lsYhovpMT6jmAt4f:-KN-lsYhovpMT6jmAt4g:-2048542497) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/thali/install/setUpTests.sh#L68)):*
> ```Shell
> # In case of UT crete a file
> if [ $2 == "UT" ]; then
>   echo "UT files will be copied to the platform directory"
> ```

Come on, you can set a shell variable with that string instead of repeating it three times

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[scripts/androidBeforeCompile.js, line 54 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj8dDSs9oQI2RF3vRP:-KNBJhexaTMogeBkJZGI:-546098859) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L54)):*
> I don't remember anymore what it was that we were fixing but what changed that we don't need to fix it anymore?
</details>
We were executing out UT in jxcore_ready method of jxcore.java. Now we are monkey patching JXcoreExtension, so we do not need it anymore. 

---

*[scripts/androidBeforeCompile.js, line 77 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl13XR3QhFIOTiX8Xn:-KNBKIjo3Yj9HowSNDxE:1847680004) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L77)):*
> There is the same code structure here three times, can we replace this and the next two functions with a call to a single function that takes the source, target and some kind of error message as args and then does the copySync and handles the error by exiting?
</details>
Agree, will be fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 81 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl30wx6CHYPFcOsDzm:-KNBKdnC1YJ6HGNtTpJ_:1788945150) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/JXcoreExtension.java#L81)):*
> I'm really not happy with having this code in our mainline JXcoreExtension. If I'm reading this right then this version of JXcoreExtension would be in our production code and we are just depending on the class detector failing to make sure it doesn't run. But that just isn't good coding practice. We don't mix our test code with our production code. 
> 
> In fact, why is this here at all? I thought the UT tests are pure native Android. Why do we need JXcoreExtension? Please explain on our next stand up. If we do need to call out to this then I would rather we monkey patched this file as part of the test install environment.
</details>
Agree, scripts are already written.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 99 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl2qLMkf7zGy0_8LJu:-KNBKkMAFpVjxV-VyjFX:-1764077876) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/JXcoreExtension.java#L99)):*
> Shouldn't we put the names in our own actual DNS registered namespace? Something like org.thaliproject.java.thalitest.RegisterExecuteUT?
</details>
The reason why we put this in com.test.thalitest is that Cordova by default creates this package while creating project, which we named thalitest and puts there MainActivity.java, the file where all starts from. So we just adjusted to it to not make a bigger mess.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[src/android/test/com/test/thalitest/MainActivity.java, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl4fUI_3Lturj0v5jG:-KNBLQk1Aswtdl0ebNDy:-515397310) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/com/test/thalitest/MainActivity.java#L2)):*
> Um... this isn't our copyright notice. We use MIT. Why is this here?
</details>
Will be replaced by MIT license.

---

*[src/android/test/com/test/thalitest/ThaliTestSuite.java, line 19 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl5QGMiTMKuGjPEbV6:-KNBLW47MeCXj5vQ-9pV:914737352) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/com/test/thalitest/ThaliTestSuite.java#L19)):*
> Is there no way to generate this list dynamically? The reason I ask is that it wouldn't be the 1st or 10th time a framework that required explicitly listing tests resulted in a situation where people adding new tests didn't realize they had to register them and they never get run and nobody every notices. :(
</details>
Scripts are already written.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 74 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-bHyTWlMafEiuBPd8:-KNBL_OLp-Ydj2p5fmDD:499427960) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L74)):*
> Please put this code and its equivalent version in tearDownAfterClass into a single function and call that.
> 
> Or, alternatively, why are we repeating the code at all? If we are going to shut everything down after every test then why do we need to do it at the end of the entire class's test run?
</details>
tearDownAfterClass will be removed.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 659 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-dBRzaVZfYoCUakeG:-KNBLcVxj2af-_HsL1Db:638874332) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L659)):*
> Yes, it's true that true is true but I"m guessing something is actually supposed to be here?
</details>
This method returns nothing, all it does is some logging about peer change, so we won't test it anymore. 

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 684 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-decZv3merKRZGmtz:-KNBM0AiJUB4hCzAmKNC:-471604708) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L684)):*
> Shouldn't we be checking that this one and the next two do call the handlers when they are supposed to be?
</details>
This methods only throw exception when they are called, there is no handler to call. They override methods from DiscoveryManager.

---

*[src/android/test/io/jxcore/node/IncomingSocketThreadTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-gnm_DGinUnOB36YI:-KNBQkH435VMFHx7dELD:-291378625) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/IncomingSocketThreadTest.java#L1)):*
> The most interesting part of this class is the run method and it doesn't seem to be tested
</details>
There is some problem with testing run method. After starting this thread, exception is thrown so we are unable to test it. 
Exception: Failed to create the local streams: failed to connect to localhost/127.0.0.1 (port 0): connect failed: ECONNREFUSED (Connection refused)

---

*[src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-jYraByp496_bmxHl:-KNBOlqzvWpZ4uhUczJ8:1300495714) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java#L1)):*
> Run is where all the fun is but it isn't being tested!
</details>
Test is already written. But, at some point, this socket/thread is waiting for some incoming connection, so either we will simulate some incoming connection or just test if everything is okay before blocking call.

---

*[test/www/jxcore/UnitTest_app.js, line 45 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-lUVlUw4I2mncxQXO:-KNBQWyI3dI3B8mqvH_h:1967378551) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/test/www/jxcore/UnitTest_app.js#L45)):*
> I"m confused why the previous code was shoved between starting the radios if they aren't on and checking if the radios actually go started. Those are two pieces of heavily related code. I suspect the previous promise which runs UT should be first and in fact should probably run before the getNetworkStatus call?
</details>
Yes, UT should execute before checking radios and before e2e tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[scripts/androidBeforeCompile.js, line 54 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj8dDSs9oQI2RF3vRP:-KNBRbjSEvepq4u_7pbw:677922019) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/scripts/androidBeforeCompile.js#L54)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> We were executing out UT in jxcore_ready method of jxcore.java. Now we are monkey patching JXcoreExtension, so we do not need it anymore. 
</details>
our*

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 81 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl30wx6CHYPFcOsDzm:-KNBRuGs7EeQvML6qPrb:181445994) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/JXcoreExtension.java#L81)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> Agree, scripts are already written.
</details>
And the reason why we are firing UT from JXcoreExtension is that we need to wait for jxcore to load properly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 26 unresolved discussions, some commit checks failed.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-eQloYB63_LcU01Un:-KNBguk-VB_KkOQfhb_d:1789689612) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L1)):*
> Have y'all tried to run any kind of code coverage statistics so we can see how much of the code we are actually covering?
</details>
Code coverage could be a problem because we are actually using device resources in our tests so they are not exactly UT but rather Instrumentation tests. I don't think there are tools that provide code coverage on non-rooted phones. Also, our tests are not in different package than classes we are testing, because this classes are avalaible only in their package, so this could be also a problem for coverage tools. I will look into that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

jareksl:
*[src/android/java/io/jxcore/node/ListenerOrIncomingConnection.java, line 59 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl3the3Hzak-q3NpCL:-KNBpLP3bo9EbTOHKoPa:256046187) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/ListenerOrIncomingConnection.java#L59)):*
> If we lived with a bug this bad do we really need this method?
</details>
None of the methods: getClientPortNumber() , setClientPortNumber(),  getServerPortNumber() , setServerPortNumber() is used, that's the reason why we didn't spot this bug before. Should we remove them if we don't use them?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

jareksl:
*[src/android/test/com/test/thalitest/MainActivity.java, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl4fUI_3Lturj0v5jG:-KNBuGGm2p-FflWg5KPE:222702326) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/com/test/thalitest/MainActivity.java#L2)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> Will be replaced by MIT license.
</details>
This file is generated by Cordova. It is leftover from the previous implementation, now it is not needed. Will be removed from the repository.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

jareksl:
*[src/android/test/io/jxcore/node/jxcore.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMj85eykrFdadD57nYw:-KNC2GRmfoElIeSX-MQr:1834636005) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/jxcore.java#L1)):*
> Is this still the case? I'm ignoring this file for the moment until I get clarification. Thanks!
</details>
Yes, this file is not modified anylonger.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

ThaliTester:
PR is added to the queue for testing as 2. task. (6f426cd)

ThaliTester:
PR is added to the queue for testing as 3. task. (6db488b)

yaronyg:
Reviewed 14 of 14 files at r2.
Review status: all files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[scripts/androidBeforeCompile.js, line 59 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNEK7qfwHETMvmM4ZH-:-KNEK7qfwHETMvmM4ZH0:1722016499) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/scripts/androidBeforeCompile.js#L59)):*
> ```JavaScript
> 
>   try {
>     console.log('Copying ' + message);
> ```

NB: Is it really worth logging this? We are already really chatty in our logs and it makes finding errors difficult. if something fails then we get a log in the error handler so maybe we can get rid of this log?

---

*[scripts/androidBeforeCompile.js, line 112 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNEKJt4dBxcvCfFje7e:-KNEKJt4dBxcvCfFje7f:-1418089665) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/scripts/androidBeforeCompile.js#L112)):*
> ```JavaScript
>  * @param {Object} appRoot
>  */
> // jscs:enable jsDoc
> ```

Why did you have to disable jsdoc? We should be passing jsdoc validation.

---

*[scripts/androidBeforeCompile.js, line 143 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNEKgGpqpJ6TA78oCbI:-KNEKgGpqpJ6TA78oCbJ:2140120145) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/scripts/androidBeforeCompile.js#L143)):*
> ```JavaScript
>  * @param {Object} appRoot
>  */
> // jscs:enable jsDoc
> ```

Same issue, why are we disabling jsOoc validation? Please apply this comment to the rest of the file. :)

---

*[scripts/androidBeforeCompile.js, line 175 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNEMN4Ksylr_DNPCliX:-KNEMN4LWOFvAlpGHNqx:2085738813) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/scripts/androidBeforeCompile.js#L175)):*
> ```JavaScript
> // jscs:disable jsDoc
> /**
>  * Updates JXcoreExtension with a method is used to register the native UT executor
> ```

Please throw in a note that we are doing this because we do not want to edit the original extension file that goes into production with test code so instead we add it dynamically only when we are testing.

---

*[src/android/java/io/jxcore/node/JXcoreExtension.java, line 96 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNEMfXoXVilfqkZmwO_:-KNEMfXoXVilfqkZmwOa:1516394239) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/src/android/java/io/jxcore/node/JXcoreExtension.java#L96)):*
> ```Java
>         });
> 
>         lifeCycleMonitor.start();
> ```

We should probably stick in a comment here informing the reader that the native tests plan on sticking their content dynamically after this line so anyone editing this line needs to go check out updateJXCoreExtensionWithUTMethod in androidBeforeCompile.js

---

*[src/android/java/io/jxcore/node/ListenerOrIncomingConnection.java, line 59 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KMl3the3Hzak-q3NpCL:-KNDgFGXKYXdsMb2xm18:-1904551135) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/java/io/jxcore/node/ListenerOrIncomingConnection.java#L59)):*
<details><summary><i>Previously, jareksl (Jarosław Słota) wrote…</i></summary>
> None of the methods: getClientPortNumber() , setClientPortNumber(),  getServerPortNumber() , setServerPortNumber() is used, that's the reason why we didn't spot this bug before. Should we remove them if we don't use them?
</details>
If we aren't using them, kill 'em. They are certainly easy to put back in later.

---

*[src/android/test/com/test/thalitest/UTMethod, line 10 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENBYyDtmMRvtJ3HCT:-KNENBYyDtmMRvtJ3HCU:-1200846719) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/src/android/test/com/test/thalitest/UTMethod#L10)):*
> ```
> 
>     private static void RegisterUTExecuteMethod() {
>         Class<?> clazz = RetrieveRegisterUTClass();
> ```

Isn't this code actually bad in the sense that if we have included UTMethod then we did so because we expect to run tests and if tests don't run because the class isn't there then we know something horrible has gone wrong and we should have an error.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 659 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-dBRzaVZfYoCUakeG:-KNDgdQiZ63-mJqw1d_4:-1096177212) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L659)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> This method returns nothing, all it does is some logging about peer change, so we won't test it anymore. 
</details>
Actually this is a pretty seriously critical method since it what tells JXcore when we have found a peer. That's a pretty big deal. Shouldn't we test that it's getting called correctly?

If you agree then please open a separate issue to add the test. I don't want to hold up this PR over writing that test.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 684 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-decZv3merKRZGmtz:-KNDhOKYYTTYp_vHVYJb:490795559) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L684)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> This methods only throw exception when they are called, there is no handler to call. They override methods from DiscoveryManager.
</details>
So shouldn't we be testing that they throw? And if so then please add as a separate bug as I don't want to hold up this PR over this issue.

---

*[src/android/test/io/jxcore/node/IncomingSocketThreadTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-gnm_DGinUnOB36YI:-KNDhaBuy5GxC7klkOSS:1913399222) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/IncomingSocketThreadTest.java#L1)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> There is some problem with testing run method. After starting this thread, exception is thrown so we are unable to test it. 
> Exception: Failed to create the local streams: failed to connect to localhost/127.0.0.1 (port 0): connect failed: ECONNREFUSED (Connection refused)
</details>
It is trying to connect in order to move data. The whole point of the class. So we need to give it an endpoint to connect to (it's localhost so we are good there) and make sure we are behaving right. Please add an issue that we should test this but let's not hold up the PR over it.

---

*[src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-jYraByp496_bmxHl:-KNDhs_9PRkHqRGeUioX:1585354779) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java#L1)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> Test is already written. But, at some point, this socket/thread is waiting for some incoming connection, so either we will simulate some incoming connection or just test if everything is okay before blocking call.
</details>
Right, we have to create something that creates a local socket and sends in data. This all looks like normal sockets so testing it should be easy (ish). As with the other similar comments please just file a bug on this issue, I don't want to hold up the PR for it.

---

*[src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java, line 75 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNEN_ICPpQcAc316lQT:-KNEN_IDn6hWD3IiPMUX:61353812) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/src/android/test/io/jxcore/node/OutgoingSocketThreadTest.java#L75)):*
> ```Java
>         assertThat("mServerSocket.isBound should return true", mServerSocket.isBound(),
>                 is(true));
>         //TODO Simulate incoming connection
> ```

yeah, exactly. I'm o.k. though with us filing this as a separate issue and dealing with it later.

---

*[test/www/jxcore/UnitTest_app.js, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENiitoo3MEksnVPqs:-KNENiitoo3MEksnVPqt:-1749508620) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/test/www/jxcore/UnitTest_app.js#L18)):*
> ```JavaScript
> 
> 
> Mobile('ExecuteNativeTests').callNative(function (result) {
> ```

So what happens when I'm running on iOS and there is no ExecuteNativeTests?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 29 unresolved discussions, some commit checks failed.

---

*[src/android/test/com/test/thalitest/UTMethod, line 10 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENBYyDtmMRvtJ3HCT:-KNG_9oTPv8D5K27LGQX:890937637) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/src/android/test/com/test/thalitest/UTMethod#L10)):*
> Isn't this code actually bad in the sense that if we have included UTMethod then we did so because we expect to run tests and if tests don't run because the class isn't there then we know something horrible has gone wrong and we should have an error.
</details>
We are already logging every error, also when there is no RegisterExecuteUT file, exception is thrown so we will definitely be aware of every bad thing. If you have any suggestions  how we should do it, please write it in reply.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 659 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-dBRzaVZfYoCUakeG:-KNGi_bbczltgnLe_r9o:1912606076) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L659)):*
> Actually this is a pretty seriously critical method since it what tells JXcore when we have found a peer. That's a pretty big deal. Shouldn't we test that it's getting called correctly?
> 
> If you agree then please open a separate issue to add the test. I don't want to hold up this PR over writing that test.
</details>
If I understand it correctly, this method Override method  onPeerUpdated from DiscoveryManager. We are not testing DiscoveryManager but ConnectionHelper and in this class, this method only do logs. Or, you meant onPeerDiscovered and onPeerLost, this methods are actually telling jxcore about some events and I agree we should test them. 

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 684 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-decZv3merKRZGmtz:-KNGeHc2PXIjiRgOkYbF:-523393094) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L684)):*
> So shouldn't we be testing that they throw? And if so then please add as a separate bug as I don't want to hold up this PR over this issue.
</details>
I agree and we are actually doing it by expecting that they will throw UnsupportedOperationException. 

---

*[test/www/jxcore/UnitTest_app.js, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENiitoo3MEksnVPqs:-KNGfu8Nc_tzCWyHKj8_:-1098510845) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/test/www/jxcore/UnitTest_app.js#L18)):*
> So what happens when I'm running on iOS and there is no ExecuteNativeTests?
</details>
As we agreed, on iOS, method used to execute tests will have the same name. Even when there is no such registered method, which I doubt as far as this file is used specifically for running tests, log with error will occur: E/JX-Cordova: JavaCall recevied a call for unknown method ExecuteNativeTests. And it has no affect on anything at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

yaronyg:
Reviewed 5 of 5 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[scripts/androidBeforeCompile.js, line 174 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNIw243my_Kqb1DGqAt:-KNIw244gGCh_lj-vhby:-1719000453) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6f742a4d5dc45886641c4e2f5b824aa743bd9441/scripts/androidBeforeCompile.js#L174)):*
> ```JavaScript
> // jscs:disable jsDoc
> /**
>  * Updates JXcoreExtension with a method used to register the native UT executor.
> ```

NB: For some reason in the line above 'a method is' was turned into 'a method'. Please put the is back.

---

*[src/android/test/com/test/thalitest/UTMethod, line 10 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENBYyDtmMRvtJ3HCT:-KNIt5Vrq7xDKyimNnzE:497438200) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/src/android/test/com/test/thalitest/UTMethod#L10)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> We are already logging every error, also when there is no RegisterExecuteUT file, exception is thrown so we will definitely be aware of every bad thing. If you have any suggestions  how we should do it, please write it in reply.
</details>
I was thinking that this whole class can just go away. The native code in jxcoreextension.java should just directly call the test class. If something went wrong and the extension method isn't in jxcoreextension then we'll get an error when we call it. Better yet, we'll get an error when we compile.

---

*[src/android/test/io/jxcore/node/ConnectionHelperTest.java, line 659 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-dBRzaVZfYoCUakeG:-KNItfcJpV79mrxoYJoE:-246865129) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/ConnectionHelperTest.java#L659)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> If I understand it correctly, this method Override method  onPeerUpdated from DiscoveryManager. We are not testing DiscoveryManager but ConnectionHelper and in this class, this method only do logs. Or, you meant onPeerDiscovered and onPeerLost, this methods are actually telling jxcore about some events and I agree we should test them. 
</details>
D'oh!!!! I'm sorry. I was looking at onPeerDiscovered and not onPeerUpdate. My bad.

---

*[test/www/jxcore/UnitTest_app.js, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENiitoo3MEksnVPqs:-KNItuTHBloUyHvTFvxC:2074643071) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/test/www/jxcore/UnitTest_app.js#L18)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> As we agreed, on iOS, method used to execute tests will have the same name. Even when there is no such registered method, which I doubt as far as this file is used specifically for running tests, log with error will occur: E/JX-Cordova: JavaCall recevied a call for unknown method ExecuteNativeTests. And it has no affect on anything at all.
</details>
Then that itself is a bug. :) We that exception is thrown we have to make sure that somehow or another it causes CI to be clear that this device has failed its tests. When CI gets the "No UT executed." result what will it do with it? Will it register it as a failure?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

mlesnic:
Review status: all files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[src/android/test/com/test/thalitest/UTMethod, line 10 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENBYyDtmMRvtJ3HCT:-KNWHqeBnOWCjaWGH6AF:2036163077) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/src/android/test/com/test/thalitest/UTMethod#L10)):*
> I was thinking that this whole class can just go away. The native code in jxcoreextension.java should just directly call the test class. If something went wrong and the extension method isn't in jxcoreextension then we'll get an error when we call it. Better yet, we'll get an error when we compile.
</details>
Ok, this file will be removed, instead of we will just call the static method from RegisterExecuteUT and to do it we will need additional import. So if something will go wrong, compile error should be thrown. 

---

*[test/www/jxcore/UnitTest_app.js, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENiitoo3MEksnVPqs:-KNWKNQ8TWjcLT7cmZpp:1678683403) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/test/www/jxcore/UnitTest_app.js#L18)):*
> Then that itself is a bug. :) We that exception is thrown we have to make sure that somehow or another it causes CI to be clear that this device has failed its tests. When CI gets the "No UT executed." result what will it do with it? Will it register it as a failure?
</details>
We will inform CI about tests results by logging specific message.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

ThaliTester:
Test (Fail) 77622416 build is completed (6f426cd)

See https://github.com/ThaliTester/TestResults/tree/776224166f426cd_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
Test (Fail) 77622416 build is completed (6db488b)

See https://github.com/ThaliTester/TestResults/tree/776224166db488b_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

yaronyg:
There is a problem in CI and @czyzm will investigate tomorrow

ThaliTester:
PR is added to the queue for testing as 1. task. (75fe645)

ThaliTester:
Test (Fail) 77622416 build is completed (75fe645)

See https://github.com/ThaliTester/TestResults/tree/7762241675fe645_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (209fb6c)

ThaliTester:
Test (Fail) 77622416 build is completed (209fb6c)

See https://github.com/ThaliTester/TestResults/tree/77622416209fb6c_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (2e8529c)

ThaliTester:
Test (Fail) 77622416 build is completed (2e8529c)

See https://github.com/ThaliTester/TestResults/tree/776224162e8529c_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (2945f57)

ThaliTester:
Test (Fail) 77622416 build is completed (2945f57)

See https://github.com/ThaliTester/TestResults/tree/776224162945f57_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

yaronyg:
Will figure out which devices we pass on and for now we will just test on those

ThaliTester:
PR is added to the queue for testing as 1. task. (ad9274d)

ThaliTester:
Test (Success) 77622416 build is completed (ad9274d)

See https://github.com/ThaliTester/TestResults/tree/77622416ad9274d_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
Test 77622416ad9274d(ad9274d) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/77622416ad9274d_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

jareksl:
*[scripts/androidBeforeCompile.js, line 174 \[r5\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNIw243my_Kqb1DGqAt:-KNfkkeJ_zzV0NAIPKwO:1157347499) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6f742a4d5dc45886641c4e2f5b824aa743bd9441/scripts/androidBeforeCompile.js#L174)):*
> NB: For some reason in the line above 'a method is' was turned into 'a method'. Please put the is back.
</details>
sorry, but I don't quite see what you mean. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

jareksl:
*[src/android/test/io/jxcore/node/StreamCopyingThreadTest.java, line 103 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KN-kYiydb7FFVUHzS61:-KNg9iymwQtTGxvZ8zri:1733832561) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/35c04ce59a0de1948d05f4a15cd62c0e963977f8/src/android/test/io/jxcore/node/StreamCopyingThreadTest.java#L103)):*
> Shouldn't we get unhappy if this is called? And actually shouldn't we test how we behave in that case?
</details>
Will be implemented as a part of the 777

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

ThaliTester:
PR is added to the queue for testing as 1. task. (b768259)

ThaliTester:
Test (Success) 77622416 build is completed (b768259)

See https://github.com/ThaliTester/TestResults/tree/77622416b768259_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
Test 77622416b768259(b768259) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/77622416b768259_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

yaronyg:
:lgtm:

I do have comments that I have included and I would ask that you resolve them but please resolve them as you see fit and then check in. This PR needs to finally be in vNext!!!! Congrats and thanks!!!!

---

Reviewed 6 of 8 files at r6, 1 of 3 files at r7, 1 of 1 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, 3 of 3 files at r16.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[build.sh, line 65 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi00stvGI1pO2RNLVP:-KNi00sueDQO6XVfPIyg:-1888486298) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/build.sh#L65)):*
> ```Shell
> # Run first the tests that can be run on desktop
> thali/install/setUpDesktop.sh;ERROR_ABORT
> cd test/www/jxcore/;ERROR_ABORT
> ```

NB: Please put in a TODO comment here that explains why we disabled these tests.

---

*[scripts/androidBeforeCompile.js, line 6 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi09eAtrAEn7zjPlAr:-KNi09eAtrAEn7zjPlAs:-508512335) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/scripts/androidBeforeCompile.js#L6)):*
> ```JavaScript
> var path = require('path');
> 
> // jscs:enable jsDoc
> ```

NB: Actually the jscs comments should be removed completely. E.g. we don't need //jscs:enable jsDoc, we shouldn't need anything here.

---

*[src/android/test/com/test/thalitest/RegisterExecuteUT.java, line 19 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi26Mb9flCktGt2eU1:-KNi26Mco8swEIDECHrA:562727135) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/src/android/test/com/test/thalitest/RegisterExecuteUT.java#L19)):*
> ```Java
>     }
> 
>     public static void RegisterExecuteUT() {
> ```

NB: My first reaction when I saw this code is "There are no static constructors in Java!" But of course that isn't what you are doing. But I think what you did do is a bit confusing since giving the static method the same name as the class confuses the static constructor issue. In general I think the advice given in the first answer in http://stackoverflow.com/questions/7486012/static-classes-in-java is good. Use a different name for the static method from the class. Declare the class final and make the constructor private. I would probably also throw in a 'throw' statement into the constructor just to make it clear that nobody is ever supposed to call it.

---

*[test/www/jxcore/UnitTest_app.js, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENiitoo3MEksnVPqs:-KNhzRw1xOyXmxjXHMA6:-94934727) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/test/www/jxcore/UnitTest_app.js#L18)):*
<details><summary><i>Previously, mlesnic (Marcin Lesniczek) wrote…</i></summary>
> We will inform CI about tests results by logging specific message.
</details>
NB: Have you actually tested a failure and made sure that CI will properly fail? Once you do, then please check in. :)

---

*[test/www/jxcore/UnitTest_app.js, line 38 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi3NWF1jjsqMRPs0s4:-KNi3NWGZBT6vcGtAmz3:-2012681304) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/test/www/jxcore/UnitTest_app.js#L38)):*
> ```JavaScript
> 
> // finish testing here (the node part will be omitted)
> console.log('****TEST_LOGGER:[PROCESS_ON_EXIT_SUCCESS]****');
> ```

NB: Make me a happy person and stick a return here. The trick of telling the CI runner that we are done and having it kill the process before we get the node tests running is a little too tricky for me. If you just stick in a return then we know the node tests won't run.

---

*[thali/install/setUpTests.sh, line 41 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi3aet-NME2_cBt2Gy:-KNi3aeujMH2UXVAZ7Ol:-479933462) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/thali/install/setUpTests.sh#L41)):*
> ```Shell
> 
> cd ThaliTest
> #cordova platform add ios
> ```

NB: Please stick in a comment explaining why this is commented out with a TODO on it

---

*[thali/install/setUpTests.sh, line 74 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi3fbcaQUFR9sez8H6:-KNi3fbcaQUFR9sez8H7:-1396797664) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/thali/install/setUpTests.sh#L74)):*
> ```Shell
> cordova build android --release --device
> 
> #if [ $runningInMinGw == false ]; then
> ```

NB: Please stick in a comment here as well

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-:-KNi3wCyy3rIYvztwnwf:-1393800703)*

ThaliTester:
PR is added to the queue for testing as 1. task. (3c26f5f)

ThaliTester:
Test (Success) 77622416 build is completed (3c26f5f)

See https://github.com/ThaliTester/TestResults/tree/776224163c26f5f_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
Test 776224163c26f5f(3c26f5f) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/776224163c26f5f_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (c9749bc)

ThaliTester:
Test (Success) 77622416 build is completed (c9749bc)

See https://github.com/ThaliTester/TestResults/tree/77622416c9749bc_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

ThaliTester:
Test 77622416c9749bc(c9749bc) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/77622416c9749bc_Native_UT_for_cordova_plugin__android___555_jareksl/ for the logs

jareksl:
*[build.sh, line 65 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi00stvGI1pO2RNLVP:-KNl0ydCEExxA4_yo_JJ:-2144980986) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/build.sh#L65)):*
> NB: Please put in a TODO comment here that explains why we disabled these tests.
</details>
Will be merged with the "Commented out ios build specific code for CI (#798)" which contains all this changes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

jareksl:
Review status: 1 of 37 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[build.sh, line 65 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi00stvGI1pO2RNLVP:-KNl1UuugBK_XvJkmEyf:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/build.sh#L65)):*
<details><summary><i>Previously, jareksl (Jarosław Słota) wrote…</i></summary>
> Will be merged with the "Commented out ios build specific code for CI (#798)" which contains all this changes
</details>
Done.

---

*[scripts/androidBeforeCompile.js, line 6 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi09eAtrAEn7zjPlAr:-KNl2HcDa5I3OSd94GrK:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/scripts/androidBeforeCompile.js#L6)):*
> NB: Actually the jscs comments should be removed completely. E.g. we don't need //jscs:enable jsDoc, we shouldn't need anything here.
</details>
Done.

---

*[src/android/test/com/test/thalitest/RegisterExecuteUT.java, line 19 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi26Mb9flCktGt2eU1:-KNl2JzC7dJ8Jf9pz08P:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/src/android/test/com/test/thalitest/RegisterExecuteUT.java#L19)):*
> NB: My first reaction when I saw this code is "There are no static constructors in Java!" But of course that isn't what you are doing. But I think what you did do is a bit confusing since giving the static method the same name as the class confuses the static constructor issue. In general I think the advice given in the first answer in http://stackoverflow.com/questions/7486012/static-classes-in-java is good. Use a different name for the static method from the class. Declare the class final and make the constructor private. I would probably also throw in a 'throw' statement into the constructor just to make it clear that nobody is ever supposed to call it.
</details>
Done.

---

*[test/www/jxcore/UnitTest_app.js, line 18 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNENiitoo3MEksnVPqs:-KNl2MTMB3awBHQGC177:-1788683283) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/6db488b78bfa1d6aa7c6544618af52207be10ad2/test/www/jxcore/UnitTest_app.js#L18)):*
> NB: Have you actually tested a failure and made sure that CI will properly fail? Once you do, then please check in. :)
</details>
yes, in case of the UT failure there is a proper comment like: Test 75789268a387c77(a387c77) has failed

---

*[test/www/jxcore/UnitTest_app.js, line 38 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi3NWF1jjsqMRPs0s4:-KNl2fS3DzrN7J9NYcC5:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/test/www/jxcore/UnitTest_app.js#L38)):*
> NB: Make me a happy person and stick a return here. The trick of telling the CI runner that we are done and having it kill the process before we get the node tests running is a little too tricky for me. If you just stick in a return then we know the node tests won't run.
</details>
Done.

---

*[thali/install/setUpTests.sh, line 41 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi3aet-NME2_cBt2Gy:-KNl2j5BqX-B_fvt9PVf:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/thali/install/setUpTests.sh#L41)):*
> NB: Please stick in a comment explaining why this is commented out with a TODO on it
</details>
Done.

---

*[thali/install/setUpTests.sh, line 74 \[r16\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762#-KNi3fbcaQUFR9sez8H6:-KNl2lVPL5p0qF7t6ba9:-499146009) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b7682596893b8207b095c45f8a72ea200a2325dd/thali/install/setUpTests.sh#L74)):*
> NB: Please stick in a comment here as well
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/762)*

PR #776, opened 2016-07-21
Create a custom Thali PouchDB constructor utility

### LINKED ISSUE #773: Create a custom Thali PouchDB constructor utility
## Problem: 
In order to use PouchDB against a local file (as opposed to a remote URL) in JXcore on mobile we need to set two options, one is called db and we have to set it to leveldown-mobile and the other is called prefix and it specifies what directory the DB's files are to be stored in. The prefix matters because on mobile there are only certain directories we can store things in.
   Everything was fine and good until PouchDB broke prefix badly in the last major release. In the old days prefix ONLY applied to local DBs. But now pouchDB applies prefix to both local DBs and to remote DBs. In other words if I set new PouchDB('foo', {prefix: '/bar'}) then PouchDB will try to create foo's files in the /bar directory. If I call new PouchDB('http://foo.com', {prefix: '/bar'}) then now it does something really bizarre like try to connect to the server /barhttp://foo.com or something like that. Bad. Bad. Bad.
  Now you may ask, who cares? Just make sure not to pass in prefix to PouchDB instances that use a URL. But it's not that simple. The problem is replication. When we replicate we MUST start by creating a var remotePouch = new PouchDB('http://foo.com', {stick in a bunch of ajax options to configure PSK and other stuff here}) in order to configure all of our network options correctly, set up PSK, etc. Then we take that object and we have to call remotePouch.replicate.to(foo). Where foo is the local dB name we are writing to. Now replicate.to() does take options but one of those options is not prefix. So how do we make sure that we read foo from the right directory?
 Historically we did this by telling people to use the defaults feature in PouchDB. This feature lets one set a number of default constructor options and when PouchDB creates a new instance of itself (for example, it will create a new instance of itself to represent foo above) it will use those defaults. So we would set prefix on the defaults and when PouchDB created a new instance of itself to represent foo it would automatically have the prefix. But if we use that feature now then it would break the HTTP URL since the default applies to all calls to new now, not just to local files.
  So how do we let people set the prefix just for local files and not for remote URLS?

## Solution: 
The solution is [here](https://github.com/thaliproject/Thali_CordovaPlugin/blob/8130305ed2a11afe9cb9acb8e80628870e4298cf/test/www/jxcore/lib/testUtils.js#L247). This is a function I wrote with the help of Nolan Lawson (it's really just stolen code from PouchDB) that wraps the constructor such that when the constructor creates a new version of itself if the version is for a local file then it will set the db to leveldown-mobile and the prefix to the right directory and if it is for a remote URL it does neither.
  But that function right now only exists in our test code. What we need to do is extract it from the test code and we need to update it so it takes two arguments. The first argument is the PouchDB object to be wrapped (no reason to make Thali itself dependent on any particular version of PouchDB) and the second argument is the defaultDirectory.
   I do not think we should build this into thaliManager. We should keep it as a stand alone utility because that way people can do other weird stuff with PouchDB and we won't break.

## Work Items: 
So please:
- [ ] Extract the function from the test file and move it into next generation as a utility function, please change the args as I discuss above
- [ ] Remove the test version completely and replace our call to it with the 'real' version (so our tests don't break)
- [ ] Add tests to our test infrastructure to make sure that the utility is working properly

andrew-aladev:
Fixes #773 

Added Generator, changed levelDownPouchDBGenerator to new one, added test


yaronyg:
Reviewed 7 of 9 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[doc/api/identityexchange.md, line 103 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNEPrh-hGcqld1loNLG:-KNEPrh-hGcqld1loNLH:479387882) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/doc/api/identityexchange.md#L103)):*
> ```Markdown
> var express = require('express');
> var ThaliReplicationManager = require('thali/thalireplicationmanager');
> var IdentityExchange = require('thali/identityexchange');
> ```

This file is hopelessly out of date, sorry, but there was really no way for you to know that. But I apologize for wasting your time. We need to clean up our docs, they are all from December. This work would happen as part of 621

For now please back out these changes as its best to leave the dead files dead.

---

*[test/www/jxcore/bv_tests/testPouchDBGenerator.js, line 12 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNESezn7bo-zgrcasbF:-KNESezn7bo-zgrcasbG:-199497859) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/test/www/jxcore/bv_tests/testPouchDBGenerator.js#L12)):*
> ```JavaScript
> var defaultDirectory = './db/';
> 
> if (! fs.existsSync(defaultDirectory)){
> ```

Please remove the space between the ! and the fs variable

---

*[test/www/jxcore/bv_tests/testPouchDBGenerator.js, line 12 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNESpDzS4Ba4PY9JiQx:-KNESpDzS4Ba4PY9JiQy:-1778039620) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/test/www/jxcore/bv_tests/testPouchDBGenerator.js#L12)):*
> ```JavaScript
> var defaultDirectory = './db/';
> 
> if (! fs.existsSync(defaultDirectory)){
> ```

Rather than leaving around a /db directory why not instead create it in setup and recursively delete it in teardown?

---

*[test/www/jxcore/bv_tests/testPouchDBGenerator.js, line 81 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNETfaijbLWnRZ8vQbg:-KNETfaijbLWnRZ8vQbh:-1812790834) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/test/www/jxcore/bv_tests/testPouchDBGenerator.js#L81)):*
> ```JavaScript
>   db = LocalPouchDB('dbname');
>   t.equals(db._adapter, 'leveldb');
>   t.ok(compareFunctions(db.__opts.db, leveldownMobile));
> ```

If memory serves node.js never requires the same object twice, instead future requires just get the same object as the first. In which case couldn't you just use db.__opts.db === leveldownMobile?

---

*[thali/package.json, line 25 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNERmlOZzLRzIUcsODB:-KNERmlOZzLRzIUcsODC:292064987) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/thali/package.json#L25)):*
> ```JSON
>   "devDependencies": {
>     "jsdoc": "3.3.3",
>     "leveldown-mobile": "1.1.1"
> ```

I can see why this would be needed in the package.json in the test directory but why would it be needed at the root?

---

*[thali/NextGeneration/utils/pouchDBGenerator.js, line 7 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNEQLjJYYdzbTqS_Wi7:-KNEQLjKQrPm3vUnad91:-923473810) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/thali/NextGeneration/utils/pouchDBGenerator.js#L7)):*
> ```JavaScript
> 
> 
> function PouchDBGenerator(PouchDB, defaultDirectory, options) {
> ```

We really need a JSDoc explaining why this exists and what the args definitions are. Obviously options would be Object but we should explain that they are the PouchDB options for its constructor.

---

*[thali/NextGeneration/utils/pouchDBGenerator.js, line 26 \[r2\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-KNERAby65VU5Aiit4nc:-KNERAbzEAGDIR81tH98:-137388915) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9b5d2c00d8cb77e0cb34c39ebf86e704e26c559e/thali/NextGeneration/utils/pouchDBGenerator.js#L26)):*
> ```JavaScript
>     opts = extend({}, opts);
> 
>     // workaround start
> ```

If you are reading this code and you see 'workaround start' what would you think it is a work around for? There is no context.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776)*

yaronyg:
:lgtm:

---

Reviewed 4 of 5 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/776#-:-KNJYjimtuISTuXdokXX:1417088721)*

PR #811, opened 2016-07-29
Almost to the top of the Thali stack

yaronyg:
@andrew-aladev - Your job on this PR is to review the actual code since it's all node code. Please DO NOT review any of the files that are in test/www/jxcore/bv_tests/disabled.

@evabishchevich - Your job on this PR is to build and run it on your local Android devices and see how it behaves itself. This code all passes desktop tests so if there is a problem on Android it will be because either Node's expectations of Android or wrong or vice versa. Please leverage @andrew-aladev for node issues. Also make sure you read the test/Readme.md on sinopia. Without that you won't be able to build and deploy. Take a look in particular at UnitTest_App.js line 18. I expect I screwed that if condition up. Normally I would handle this myself but it's a good test run for you and I really need to get Alex unblocked on iOS.

A note to both of you - no, PRs aren't supposed to be this ridiculously huge. But I was working on this code by myself for months part-time and there was no one around to do PRs. In the future PRs will be much smaller.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/811)


ThaliTester:
PR is added to the queue for testing as 1. task. (d1cf255)

ThaliTester:
Test (Fail) 79426650 build is completed (d1cf255)

See https://github.com/ThaliTester/TestResults/tree/79426650d1cf255_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

yaronyg:
@evabishchevich Note that this PR won't be able to run in CI until [this issue](https://github.com/thaliproject/CI/issues/49) is resolved. So please just run it on your local machine.

ThaliTester:
PR is added to the queue for testing as 1. task. (ec0c820)

ThaliTester:
Test (Fail) 79426650 build is completed (ec0c820)

See https://github.com/ThaliTester/TestResults/tree/79426650ec0c820_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (412a8a7)

ThaliTester:
Test (Fail) 79426650 build is completed (412a8a7)

See https://github.com/ThaliTester/TestResults/tree/79426650412a8a7_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (88517bb)

ThaliTester:
Test (Fail) 79426650 build is completed (88517bb)

See https://github.com/ThaliTester/TestResults/tree/7942665088517bb_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

test/www/jxcore/UnitTest_app.js
```diff
@@ -15,30 +15,28 @@ var ThaliMobile = require('thali/NextGeneration/thaliMobile');
 var Promise = require('lie');
 var utResult;
```
andrew-aladev: 'utResult' is not needed in this scope. Please move it to new if-else scope.

test/www/jxcore/lib/testUtils.js
```diff
@@ -437,3 +510,198 @@ module.exports.getSamePeerWithRetry = function (path, pskIdentity, pskKey,
       nonTCPAvailableHandler);
   });
 };
+
+module.exports.createPskPouchDBRemote = function (serverPort, dbName,
+                                                 pskId, pskKey, host) {
+  var serverUrl = 'https://' + (host ? host : '127.0.0.1') + ':' + serverPort +
+    '/db/' + dbName;
```
andrew-aladev: We can use BASE_DB_PATH from thaliConfig.

test/www/jxcore/lib/testUtils.js
```diff
@@ -437,3 +510,198 @@ module.exports.getSamePeerWithRetry = function (path, pskIdentity, pskKey,
...
+    return Promise.reject(new Error('Had a failure in ThaliMobile.start - ' +
+      JSON.stringify(combinedResult)));
+  }
+  return Promise.resolve();
+};
+
+module.exports.setUpServer = function (testBody, appConfig) {
+  var app = express();
+  appConfig && appConfig(app);
+  app.use('/db', expressPouchdb(LevelDownPouchDB, {mode: 'minimumForPouchDB'}));
```
andrew-aladev: We can use BASE_DB_PATH from thaliConfig instead of '/db'.

thali/NextGeneration/replication/thaliReplicationPeerAction.js
```diff
@@ -127,32 +204,106 @@ ThaliReplicationPeerAction.prototype.resultPromise = null;
...
+
+      We picked secureOptions because if we used cert (which is ignored when
+      we use PSK) we would also have to use key. We couldn't use PFX because
+      then request tries to load it! So secureOptions was our last choice and
+      it seems to work.
        */
-    }
-  }
+      var remoteUrl = 'https://' + self._peerAdvertisesDataForUs.hostAddress +
+        ':' + self._peerAdvertisesDataForUs.portNumber + '/db/' + self._dbName;
```
andrew-aladev: We can use BASE_DB_PATH from thaliConfig instead of '/db'.

test/www/jxcore/lib/httpTester.js
```diff
@@ -113,3 +113,67 @@ module.exports.getTestHttpsServer = function (expressApp,
...
+};
+
+module.exports.generateSeqDocPath = function (devicePublicKey) {
+  return '_local/' + thaliConfig.LOCAL_SEQ_POINT_PREFIX +
+    urlSafeBase64.encode(devicePublicKey);
+};
+
+module.exports.getSeqDoc = function(dbName, serverPort, pskId, pskKey,
+                                    devicePublicKey, host) {
+  var path = '/db/' + dbName + '/' +
```
andrew-aladev: We can use BASE_DB_PATH from thaliConfig instead of '/db'.

thali/NextGeneration/thaliPeerPool/thaliPeerPoolDefault.js
```diff
@@ -63,11 +64,18 @@ var ForeverAgent = require('forever-agent');
  */
 function ThaliPeerPoolDefault() {
   ThaliPeerPoolDefault.super_.call(this);
+  this.stopped = false;
```
andrew-aladev: There is another variable with the same sense in project/ Its name is '_started'. We can use '_stopped' here.

ThaliTester:
PR is added to the queue for testing as 1. task. (9ec99e9)

ThaliTester:
Test (Fail) 79426650 build is completed (9ec99e9)

See https://github.com/ThaliTester/TestResults/tree/794266509ec99e9_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (f41621e)

ThaliTester:
Test (Fail) 79426650 build is completed (f41621e)

See https://github.com/ThaliTester/TestResults/tree/79426650f41621e_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

czyzm:
After updating CI the cooperation with Sinopia is fine. Tasks related to custom pouchdb are executed correctly.
The build fails when executing jx runTests.js (/test/www/jxcore) with error:
"Error: Missing PFX or certificate + private key."
 
This comes from jxcore tls.js (line 1064) where options passed to createServer are checked:
if (!self.pfx && (!self.cert || !self.key)) {
    throw new Error('Missing PFX or certificate + private key.');
}

andrew-aladev:
`Error: Missing PFX or certificate + private key`

I had the same issue on Linux when I tried to build jxcore (master branch). Everything builded and installed, but 90% tests has failed with this error. Tests works perfect on Mac.

ThaliTester:
PR is added to the queue for testing as 1. task. (a9475e1)

yaronyg:
Review status: 0 of 58 files reviewed at latest revision, 6 unresolved discussions.

---

*[test/www/jxcore/UnitTest_app.js, line 16 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-KNsXglcgu5gDmbLdV3q-r4-16:-KOCa3fR5Ng-xRk0jeKo:-462935397) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/88517bb5e61caa455f1ecd1b1743973d1b864b2e/test/www/jxcore/UnitTest_app.js#L16)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> 'utResult' is not needed in this scope. Please move it to new if-else scope.
</details>
In theory I could move utResult inside of line 18's if statement but what would that accomplish? Thanks to hoisting (and we don't have ES6 support for let/const) isn't the variable defined in its 'true' context and thus more accurate?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811)*

ThaliTester:
Test (Fail) 79426650 build is completed (a9475e1)

See https://github.com/ThaliTester/TestResults/tree/79426650a9475e1_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

andrew-aladev:
> In theory I could move utResult inside of line 18's if statement but what would that accomplish? Thanks to hoisting (and we don't have ES6 support for let/const) isn't the variable defined in its 'true' context and thus more accurate?

This is just for cosmetics. 'utResult' is related only to "UT tests".

yaronyg:
Gotcha. I have to admit that I have very mixed feelings about this area and worse yet, I tend to be inconsistent. I think strictly speaking it's best to put all the variables in the same scope in the same place so there is no confusion. Especially for people who don't do Javascript all the time it's very easy to think that a variable is scoped when it's not (at least until we get let/const). On the other hand, the people reading this code are probably Javascript types and understand the issues. So putting the variable declaration in the right 'place' (even if that place doesn't define a new scope) is neater. For now, I'm feeling lazy so I'll leave it as it is. I just hope we move to Typescript fast and these kind of issues go away.

---

Review status: 0 of 58 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-:-KOIX1UQnuDgJuyXBdga:235188148)*

ThaliTester:
PR is added to the queue for testing as 1. task. (2283b5d)

ThaliTester:
Test (Success) 79426650 build is completed (2283b5d)

See https://github.com/ThaliTester/TestResults/tree/794266502283b5d_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
Test 794266502283b5d(2283b5d) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/794266502283b5d_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

yaronyg:
@czyzm @jareksl @mlesnic Can you guys explain to me how come CI thinks this test passed?!?!? I checked the logs and there are failures all over the place. What the heck is going on?

ThaliTester:
PR is added to the queue for testing as 1. task. (eeb77ae)

yaronyg:
Andrew said the following:

Let me say what I think about this pr:
1.	thali/install/setUpTests.sh and thali/install/setUpDesktop.sh have obtained a custom pouch db and android unittests dir.
2.	thali/NextGeneration/thaliWifiInfrastructure.js has jsdoc fixed and BOGUS_* deleted. I don't know the reason why this exists in master. Our working branch vNext_yarong_417_1 doesn't have this thing.
3.	We don't care about thali/NextGeneration/thaliReplicationManager.js because it will be replaced with new one.
4.	thali/NextGeneration/thaliPeerPool/thaliPeerPoolDefault.js has obtained an ability to stop its instance. It looks fine.
5.	thali/NextGeneration/thaliMobileNativeWrapper.js has BOGUS_* deleted.
6.	thali/NextGeneration/thaliConfig.js had BOGUS_* deleted and some options added. We are already using these options in vNext_yarong_417_1.
7.	We don't care about thali/NextGeneration/security/thaliACLLayer.js because it will be deleted.
8.	thali/NextGeneration/replication/thaliSendNotificationBasedOnReplication.js obtained functions getPskIdToSecret and getPskIdToPublicKey that we are already using.
9.	thali/NextGeneration/replication/thaliReplicationPeerAction.js that's the file I am working with every day. This file gave me a 'race condition' that we were talking about several days ago. We can implement more presize errorArray handling and verbose printing in future. We should definitely write many tests for this code in future. But the code itself looks fine.
10.	thali/NextGeneration/replication/thaliPullReplicationFromNotification.js has obtained an ability to stop its client. It creates ReplicationPeerAction and it is going to enqueue it to PeerPoolInterface. It uses hash to kill existing ReplicationPeerAction with the same connection type and id.
11.	thali/NextGeneration/replication/localSeqManager.js is going to sync last replication version between peers. I don't like this code, but I don't know how to make refactoring. It is not readable: timers are jumping around promises with strange self variables and overcomplicated scopes. I can investigate this in future.
12.	thali/NextGeneration/notification/thaliNotificationServer.js has fixed getPskIdToSecret, that we are already using.
13.	thali/NextGeneration/notification/thaliNotificationClient.js obtained a way to ignore empty beacon data.
14.	thali/NextGeneration/notification/thaliNotificationAction.js has added superclass kill on complete.
15.	thali/NextGeneration/makeIntoCloseAllServer.js has fixed connections array. Splice removes closed connection.
16.	test/www/jxcore/lib/testUtils.js has added test server with participants that we are using today. The code is ok.
17.	test/www/jxcore/lib/httpTester.js is a helper for testing thali/NextGeneration/replication/localSeqManager.js 
18.	test/www/jxcore/bv_tests/testthaliPeerPoolDefault.js has BOGUS_* deleted and tested stop of PeerPoolDefault.
19.	test/www/jxcore/bv_tests/testThaliReplicationPeerActionCoordinated.js and test/www/jxcore/bv_tests/testThaliReplicationPeerAction.js tested a ReplicationPeerAction. The code seems fine.
20.	test/www/jxcore/bv_tests/testThaliPullReplicationFromNotificationCoordinated.js and test/www/jxcore/bv_tests/testThaliPullReplicationFromNotification.js tested a ReplicationPeerAction. The code seems fine.


So I'll walk through these 1 by 1

1. Sounds correct

2. I forked this branch from 417 just before your first commit. Because your commit wasn't finished I didn't want to block on the issues there and so wanted to focus on just getting the existing code checked in. That's why 417 is different. Once 811 goes in to vNext then we will merge vNext_yarong_417_1 with vNext and inherit the 417 changes. But I fixed the lint issues anyway (maybe for the second time). I'm not sure what you mean by BOGUS_* deleted?

3. Correct

4. Yeah :)

5. This goes back to point 2 above, what does BOGUS_* deleted mean?

6. What does BOGUS_* deleted mean? And as mentioned above vNext_Yarong_417_1 is actually several commits ahead of this PR. See point 2 for why.

7. Agreed, once you check in thaliManager with Salti. But that won't happen here.

8. Good.

9.  We actually have a whole file of tests for thaliReplicationPeerAction. The race condition you showed me a few days ago wasn't from this file, it was because of a problem in the test sample. Do you have anything that fails that you can show me so I can investigate the possible race condition?

10. Yup

11.  This code is complex because what it is doing is complex although that doesn't mean we couldn't find a better way to write it. There are a couple of different problems that this code is handling. First, it is handling getting a constant series of update calls that it has to translate into a smaller set of network events. We are only allowed (in order to not waste bandwidth and cpu) to send a seq update to a remote peer every X milliseconds. So if we get notified that the sequence number has changed more frequently than X milliseconds (which is likely) then we have to store up the last updated seq value until X milliseconds has passed and then we can send it. But we have another problem, there is a race condition we have to prevent.

Time 0 - We issue an update to the remote peer that seq is no 10.

Time X - This is x milliseconds later when we can send a new seq update. Now we send an update that the seq is 50.

Time 2X - This is another x milliseconds later and we set the seq to 100.

Because we use multiple TCP/IP connections over HTTP it is possible that all of these requests could be processed in any random possible order. In other words at time 2X it is possible that the requests from 0, X and 2X have not yet been processed by the remote peer. In the naive case all would have the same change rev so which ever one the remote peer happens to pick first would succeed and the rest would fail because they have the wrong rev.

To prevent this mess I have a promise queue where I do something slightly sleezy. For example, let's say it is Time X and I detect that the request from Time 0 hasn't been responded to yet. In that case I put a request onto a promise queue behind the request from time 0. But let's say that the request from time 0 takes a really long time. In that case when Time 2X happens I don't add a new entry to the promise queue. Instead I leave the current one because when it executes (which can only happen after the request from Time 0 returns) it will pick up whatever is the latest seq and send it. In other words rather than queuing up two requests (the first of which would be useless since it's just going to be overwritten by the second), I only queue up one request which only sets the seq value to update at the time it actually executes (e.g. right after the previous network request finishes).

As I said, it's complicated.

12. - 17. Good?

18. What is BOGUS_* again?

19. - 21 Good

22.  Hopefully the previous explanation helped to clarify things

---

Review status: 0 of 58 files reviewed at latest revision, 11 unresolved discussions.

---

*[thali/NextGeneration/notification/thaliNotificationClient.js, line 110 \[r8\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-KOMM94VRn6f9N7YQVjE:-KOMM94WX4V7upB26sLo:1498645334) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2283b5d11607444578a66fca513febfd15691000/thali/NextGeneration/notification/thaliNotificationClient.js#L110)):*
> ```JavaScript
>  * @type {PeerAdvertisesDataForUs}
>  */
> ThaliNotificationClient.prototype.Events = {
> ```

Andrew asked if we can make this static

This used to be static. However in commit 73924f50 I changed it to be a prototype value. The reason I did this is because of a problem with testing. Thankfully I actually left an explanation in the check in comments otherwise I wouldn't have remembered. :)

The issue is that when testing we use a mock of thaliMobile. That mock does not actually call thaliNotificationClient. If we did require thaliNotificationClient (which I would have to do if I wanted to get to the events value) then that would trigger some set up behavior built into thaliNotificationClient that would  screw up the tests. So I could either proxyquire thaliNotificationClient and try to fix up the mess OR because there was literally one value (events) causing a problem I could just move events onto the prototype and not require anything. I chose the later.

I have added a comment explaining the issue.

---

*[thali/NextGeneration/replication/localSeqManager.js, line 66 \[r8\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-KOMOo0LocTd5KRZZKC0:-KOMOo0LocTd5KRZZKC1:-672516341) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2283b5d11607444578a66fca513febfd15691000/thali/NextGeneration/replication/localSeqManager.js#L66)):*
> ```JavaScript
>   self._lastUpdateTime = Date.now();
> 
>   function getRevOrNull() {
> ```

Andrew said: We can create a private method without arguments _getRevision here.

But why? This function is only used here. Or do you just not like functions defined in functions? 

---

*[thali/NextGeneration/replication/localSeqManager.js, line 79 \[r8\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-KOMPrJRqLlFfDu-sTGz:-KOMPrJRqLlFfDu-sTH-:184765635) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2283b5d11607444578a66fca513febfd15691000/thali/NextGeneration/replication/localSeqManager.js#L79)):*
> ```JavaScript
>   }
> 
>   return (self._seqDocRev ? Promise.resolve(self._seqDocRev) : getRevOrNull())
> ```

Andrew said: We can move this line into _getRevision method.

Agreed

---

*[thali/NextGeneration/replication/thaliPullReplicationFromNotification.js, line 80 \[r8\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-KOMQzINn4GL5fDL4BJs:-KOMQzINn4GL5fDL4BJt:-687674051) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2283b5d11607444578a66fca513febfd15691000/thali/NextGeneration/replication/thaliPullReplicationFromNotification.js#L80)):*
> ```JavaScript
> ThaliPullReplicationFromNotification.prototype._peerDictionary = null;
> 
> ThaliPullReplicationFromNotification.prototype._peerDictionaryKey =
> ```

Andrew said: This function asks to be static.

Agreed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/811#-:-KOMS1DflmYnL1E2McfL:-1059271301)*

ThaliTester:
Test (Success) 79426650 build is completed (eeb77ae)

See https://github.com/ThaliTester/TestResults/tree/79426650eeb77ae_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
Test 79426650eeb77ae(eeb77ae) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/79426650eeb77ae_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (9fa1f7f)

ThaliTester:
Test (Success) 79426650 build is completed (9fa1f7f)

See https://github.com/ThaliTester/TestResults/tree/794266509fa1f7f_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
Test 794266509fa1f7f(9fa1f7f) has failed

See https://github.com/ThaliTester/TestResults/tree/794266509fa1f7f_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 2. task. (45a82cf)

ThaliTester:
Test (Fail) 79426650 build is completed (45a82cf)

See https://github.com/ThaliTester/TestResults/tree/7942665045a82cf_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (616b042)

ThaliTester:
Test (Success) 79426650 build is completed (616b042)

See https://github.com/ThaliTester/TestResults/tree/79426650616b042_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
Test 79426650616b042(616b042) has failed

See https://github.com/ThaliTester/TestResults/tree/79426650616b042_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the fail logs

andrew-aladev:
> We actually have a whole file of tests for thaliReplicationPeerAction. The race condition you showed me a few days ago wasn't from this file, it was because of a problem in the test sample. Do you have anything that fails that you can show me so I can investigate the possible race condition?

Yes. This issue has been fixed, the test was broken. But when I was trying to investigate this issue I've read PeerAction many times. It triggers the replication to the pouchdb.

> Andrew said: We can create a private method without arguments _getRevision here.
But why? This function is only used here. Or do you just not like functions defined in functions?

This function needs only `self` from the parent scope. Private class method will be more strict and I think it will be better.

yaronyg:
"peerAvailabilityChange is called" is now passing.We are now up to line 768 of testThaliMobileNative.

yaronyg:
@andrew-aladev In regards to _getRevision having the function internal means no one else can call it. Reducing scope means fewer mistakes, doesn't it?

andrew-aladev:
> @andrew-aladev In regards to _getRevision having the function internal means no one else can call it. Reducing scope means fewer mistakes, doesn't it?

Yes, underscore means private function should not be called directly from outside this class or its inheritors.
There is a chance to redefine variable from the parent scope. Another problem: people can start silently use many variables from the parent scope. This creates problems when you want to change this scope.

yaronyg:
Working with Andrew on some nasty bugs. We are getting a crazy hanging bug.

yaronyg:
@andrew-aladev Actually not in this case because the function is scoped inside of the parent function so nobody can get to it.

yaronyg:
The good news is that all the tests passed today twice! The bad news is that the tests also randomly fail. :( There are problems with the server sockets getting closed before anyone can use them.

yaronyg:
Today was a bust because Genia had to set up a new machine and started work a bit late.

ThaliTester:
PR is added to the queue for testing as 2. task. (471aed9)

ThaliTester:
Test (Success) 79426650 build is completed (471aed9)

See https://github.com/ThaliTester/TestResults/tree/79426650471aed9_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the logs

ThaliTester:
Test 79426650471aed9(471aed9) has failed

See https://github.com/ThaliTester/TestResults/tree/79426650471aed9_Almost_to_the_top_of_the_Thali_stack_yaronyg/ for the fail logs

yaronyg:
Closed in favor of vNext_yarong_417_812

Issue #819, opened 2016-08-03
Handle basic functions


baydet:
Part of #769 
implement the following funcitons

`appEnteringBackground`
`appEnteredForeground`
`didRegisterToNative`

yaronyg:
There is a PR but it needs to be updated to test didRegisterToNative with a hybrid node/native test.

baydet:
`killConnections` moved to #824

Issue #49, opened 2015-07-24
Need to agree on logging solution for iOS


yaronyg:
We need to be able to configure our logging on iOS using some framework that supports:
* Setting different levels of logging
* Being able to set different levels of logging for different components
* Being able to log to rotating local files where we can delete old files
* Being able to log to a remote server for when we run tests in the testing lab

tobybrad:
Lumerjack seems to fit the bill: https://github.com/CocoaLumberjack/CocoaLumberjack

larryonoff:
I've written larger commend, but Huboard lost it. So please see just a list of the libraries that I think are interesting:
- https://github.com/Nike-Inc/Willow. Pure Swift. Doesn't support, but allows extending it.
  - Being able to log to rotating local files where we can delete old files
  - Being able to log to a remote server for when we run tests in the testing lab
- https://github.com/CocoaLumberjack/CocoaLumberjack. Objective-C with Swift support. Doesn't support, but allows extending it.
  - Being able to log to a remote server for when we run tests in the testing lab
- https://github.com/SwiftyBeaver/SwiftyBeaver. Pure Swift, but has paid plans.

yaronyg:
Make a proposal! :)

PR #826, opened 2016-08-03
#819 added basic AppContext functions

baydet:
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/826)


ThaliTester:
PR is added to the queue for testing as 1. task. (2bc7db9)

baydet:
#819

yaronyg:
Reviewed 51 of 51 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[lib/ios/ThaliCore/ThaliCore.xcodeproj/xcshareddata/xcschemes/ThaliCore.xcscheme, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOITHZyrU8rsHsyjr6r:-KOITHZzk4YGbfeja0nv:407736234) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore.xcodeproj/xcshareddata/xcschemes/ThaliCore.xcscheme#L1)):*
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> ```

Per http://stackoverflow.com/questions/16638322/what-is-xcscheme-file-what-is-it-purpose I think we should remove the xcscheme files

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOITmNtMyliWf1hmtL8:-KOITmNtMyliWf1hmtL9:-1975034818) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L2)):*
> ```Swift
> //
> //  The MIT License (MIT)
> ```

I would recommend using an IDE like AppCode that is able to spell check even camel cased function and variable names.  It will also spell check comments as well.

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 35 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOITWkJkhi_Gv2JndFV:-KOITWkJkhi_Gv2JndFW:-637140650) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L35)):*
> ```Swift
> @objc public protocol AppContextDelegate: class, NSObjectProtocol {
>     /**
>      Notifies about context's peer chagnges
> ```

NB: /chagnges/changes

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 40 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOITiRmJdYP43xHpyIz:-KOITiRmJdYP43xHpyJ-:-729656262) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L40)):*
> ```Swift
>      - parameter context: related AppContext
>      */
>     func peerAviabilityChanged(peers: Array<[String : AnyObject]>, inContext context: AppContext)
> ```

/peerAviabilityChanged/peerAvailabilityChanged

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 48 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIUDvuZhYwsbJiP0k4:-KOIUDvvCwGMOy5QxyB-:-859744585) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L48)):*
> ```Swift
>      - parameter context: related AppContext
>      */
>     func networkStatusChanged(status: [String : AnyObject], inContext context: AppContext)
> ```

Instead of passing in String : AnyObject since eventually all of these objects have expected types, can we define the Swift file to expect the expected types for each API directly and then let those types be marshalled in the JXCoreExtension.m file?

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 118 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIUqjIkuWxKyNSavmY:-KOIUqjIkuWxKyNSavmZ:-1992039223) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L118)):*
> ```Swift
>      Start the client components
> 
>      - returns: true if succeed
> ```

NB: /succeed/successful 

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 174 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIV4ngfkyKIMjuHc1J:-KOIV4ngfkyKIMjuHc1K:1079958767) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L174)):*
> ```Swift
> 
>     /**
>      Ask context to update its network status variables
> ```

Does this function have a purpose anymore?

---

*[lib/ios/ThaliCore/ThaliCoreTests/AppContextTests.swift, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIVIligzSBCEs7-x3X:-KOIVIljCdaJDaYW_7L4:-2012741275) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCoreTests/AppContextTests.swift#L2)):*
> ```Swift
> //
> //  The MIT License (MIT)
> ```

Can you please use the two line version of the copyright? Check with @larryonoff for the right text

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

larryonoff:
@yaronyg thanks for the feedback. I'll work today on resolving the issues.

larryonoff:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[lib/ios/ThaliCore/ThaliCore.xcodeproj/xcshareddata/xcschemes/ThaliCore.xcscheme, line 1 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOITHZyrU8rsHsyjr6r:-KOJOQnRxeiNz01TlZ2R:-2094126016) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore.xcodeproj/xcshareddata/xcschemes/ThaliCore.xcscheme#L1)):*
> Per http://stackoverflow.com/questions/16638322/what-is-xcscheme-file-what-is-it-purpose I think we should remove the xcscheme files
</details>
I assume it's better to leave ```xcscheme``` under version control because this file is required to build framework / apps with tools like Carthage, see the link https://github.com/Carthage/Carthage#share-your-xcode-schemes. ```xcscheme``` is also required by Xcode Server to automate the integration process of building, analyzing, testing, and archiving your app, see https://developer.apple.com/library/ios/recipes/xcode_help-scheme_editor/Articles/SchemeShare.html.

PS. I really didn't try automated compilation without using schemes, but I can try if it's needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

baydet:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 48 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIUDvuZhYwsbJiP0k4:-KOJcfSHCF3XDGS0pToK:-934271718) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L48)):*
> Instead of passing in String : AnyObject since eventually all of these objects have expected types, can we define the Swift file to expect the expected types for each API directly and then let those types be marshalled in the JXCoreExtension.m file?
</details>
I would like to replace [String  : AnyObject] with some kind of [NetworkStatusOption : Status] where NetworkStatusOption and Status are enums with String raw value. But the problem is that these enums will not be visible in objective-c. And that is the problem if we want to keep JXcoreExtension in objC. 
Also I think that we shouldn't make a lot of complex logic in JXCoreExtension extension because this logic sits outside of our Core library. It means it will not be covered with unit testing.
To make this code more type constrained I can replace [String  : AnyObject] with [String  : String].

---

*[lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift, line 174 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIV4ngfkyKIMjuHc1J:-KOJaw1h_673wCsgKHT4:-96045042) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCore/Core/AppContext.swift#L174)):*
> Does this function have a purpose anymore?
</details>
In the current version this function is used once on didRegisterToNative. It fired `AppContextDelegate`'s `networkStatusChanged` call. We could resolve this by replacing this function with readonly variable `networkStatus` and and call `networkChanged` implicitly  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

larryonoff:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[lib/ios/ThaliCore/ThaliCoreTests/AppContextTests.swift, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIVIligzSBCEs7-x3X:-KOJk5TVyZCnRUGWhE3c:718937661) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCoreTests/AppContextTests.swift#L2)):*
> Can you please use the two line version of the copyright? Check with @larryonoff for the right text
</details>
@baydet   Please see an example below

```
//
//  Thali CordovaPlugin
//  JXcoreExtension.m
//
//  Copyright (C) Microsoft. All rights reserved.
//  Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
//
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
PR is added to the queue for testing as 2. task. (3c78d34)

ThaliTester:
Test (Fail) 79896569 build is completed (3c78d34)

See https://github.com/ThaliTester/TestResults/tree/798965693c78d34__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (cf5135c)

ThaliTester:
Test (Fail) 79896569 build is completed (cf5135c)

See https://github.com/ThaliTester/TestResults/tree/79896569cf5135c__819_added_basic_AppContext_functions_baydet/ for the logs

yaronyg:
Reviewed 6 of 6 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/ios/AppContext.swift, line 48 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIUDvuZhYwsbJiP0k4:-KOS-oG8OamoRdwtZLzf:bn3dp3c) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/src/ios/AppContext.swift#L48)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> I would like to replace [String  : AnyObject] with some kind of [NetworkStatusOption : Status] where NetworkStatusOption and Status are enums with String raw value. But the problem is that these enums will not be visible in objective-c. And that is the problem if we want to keep JXcoreExtension in objC. 
> Also I think that we shouldn't make a lot of complex logic in JXCoreExtension extension because this logic sits outside of our Core library. It means it will not be covered with unit testing.
> To make this code more type constrained I can replace [String  : AnyObject] with [String  : String].
</details>
I thought we agreed that we would have a framework file that exposes the framework interface and it expects fully parsed objects. And then in Thali_CordovaPlugin we would have a swift file whose only job is to get the raw arguments from Objective-C and then parse them and call the framework with the appropriate args?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
PR is added to the queue for testing as 3. task. (8c72891)

ThaliTester:
Test (Fail) 79896569 build is completed (8c72891)

See https://github.com/ThaliTester/TestResults/tree/798965698c72891__819_added_basic_AppContext_functions_baydet/ for the logs

yaronyg:
@baydet I don't understand why you assigned this back to me. I thought we still need to resolve the issue with how to call JXcoreExtension.m?

---

Reviewed 1 of 52 files at r1, 1 of 6 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-:-KOgURotAq-8sSlf1aQA:bbbu29e)*

baydet:
@yaronyg sorry, just apparently lost a scope of what needs to be done

---

Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-:-KOiVnUrngFK2i11Z7Gr:b-zj5z0)*

ThaliTester:
PR is added to the queue for testing as 2. task. (c3873d1)

ThaliTester:
Test (Fail) 79896569 build is completed (c3873d1)

See https://github.com/ThaliTester/TestResults/tree/79896569c3873d1__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (996c122)

baydet:
@yaronyg, i've updated JXcoreExtension.m file and AppContext.swift. These files are part of cordova plugin. In AppContext object will communicate with ThaliCore and map all messages from iOS framework into consumable data structures for JXcoreExtension.m

---

Review status: 50 of 58 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-:-KOjZRqcvG1pVW0YU0tv:bnmqoea)*

ThaliTester:
Test (Fail) 79896569 build is completed (996c122)

See https://github.com/ThaliTester/TestResults/tree/79896569996c122__819_added_basic_AppContext_functions_baydet/ for the logs

baydet:
Review status: 50 of 58 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/ios/AppContext.swift, line 48 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIUDvuZhYwsbJiP0k4:-KOjbpU1paQfa778H1S6:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/src/ios/AppContext.swift#L48)):*
> I thought we agreed that we would have a framework file that exposes the framework interface and it expects fully parsed objects. And then in Thali_CordovaPlugin we would have a swift file whose only job is to get the raw arguments from Objective-C and then parse them and call the framework with the appropriate args?
</details>
Done.

---

*[lib/ios/ThaliCore/ThaliCoreTests/AppContextTests.swift, line 2 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KOIVIligzSBCEs7-x3X:-KOjbqvjteKGJJ1p49Ib:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/2bc7db9d715fb974ee7a199e839ea36f6945ec1b/lib/ios/ThaliCore/ThaliCoreTests/AppContextTests.swift#L2)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> @baydet   Please see an example below
> 
> ```
> //
> //  Thali CordovaPlugin
> //  JXcoreExtension.m
> //
> //  Copyright (C) Microsoft. All rights reserved.
> //  Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
> //
> ```
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
PR is added to the queue for testing as 1. task. (85857bc)

ThaliTester:
Test (Fail) 79896569 build is completed (85857bc)

See https://github.com/ThaliTester/TestResults/tree/7989656985857bc__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (28fb46a)

ThaliTester:
Test (Fail) 79896569 build is completed (28fb46a)

See https://github.com/ThaliTester/TestResults/tree/7989656928fb46a__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (4ed50c3)

ThaliTester:
PR is added to the queue for testing as 2. task. (abe2e03)

ThaliTester:
Test (Fail) 79896569 build is completed (4ed50c3)

See https://github.com/ThaliTester/TestResults/tree/798965694ed50c3__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
Test (Fail) 79896569 build is completed (abe2e03)

See https://github.com/ThaliTester/TestResults/tree/79896569abe2e03__819_added_basic_AppContext_functions_baydet/ for the logs

yaronyg:
Code and tests are done but blocked on iOS CI issues.

ThaliTester:
PR is added to the queue for testing as 1. task. (992243f)

ThaliTester:
Test (Fail) 79896569 build is completed (992243f)

See https://github.com/ThaliTester/TestResults/tree/79896569992243f__819_added_basic_AppContext_functions_baydet/ for the logs

yaronyg:
Reviewed 2 of 6 files at r3, 4 of 7 files at r5, 6 of 13 files at r7, 1 of 1 files at r8, 1 of 3 files at r9, 1 of 1 files at r10, 2 of 2 files at r11.
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[src/ios/AppContext.swift, line 29 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV5l8tXSIjFE_M7I8f:-KPV5l8tXSIjFE_M7I8g:b-x0rkbj) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L29)):*
> ```Swift
>      - parameter context: related AppContext
>      */
>     func context(context: AppContext, didChangeNetworkStatus status: [String : AnyObject])
> ```

My understanding of what is going on is that we need a way for the swift code to call into Objective-C so that we can get access to JXcoreExtension.m. But why use a protocol? It really confused me in trying to read it.

What I expected was to find a method in this file called sendNetworkChanged. I expect that object to take a typed structure with member names like bluetoothLowEnergy, bluetooth, wifi, cellular and bssidName. Each would be typed as a bool.

I would then expect the sendNetworkChanged method in this file to call a sendNetworkChanged method in JXcoreExtension.m whose argument is a JSON string. Said string having been created by the sendNetworkChanged method in this file. I would expect this to happen  by having a bridging header that exposes JXcoreExtension.m to swift.

No protocols. Just normal APIs with normal straight forward names.

What did I miss?

---

*[src/ios/AppContext.swift, line 143 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV6qAVySq2uVSWiX3U:-KPV6qAWRVoyXru8tWT4:b9lwuff) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L143)):*
> ```Swift
>      - parameter callback: callback with connection results.
>      */
>     public func connectToPeer(peer: String, callback:ClientConnectCallback) {
> ```

please call this multiConnect per the new spec

---

*[src/ios/AppContext.swift, line 182 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV7GojAT8eTetQXL5R:-KPV7GojAT8eTetQXL5S:b-bg7wg9) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L182)):*
> ```Swift
> /// Node functions names
> extension AppContext {
>     @objc public class func networkChanged() -> String {
> ```

There is an old joke in CS, there are only 3 numbers in CS, 0, 1 and infinity. If none of these functions is called more than once is it really worth creating this whole class and adding this level of indirection? Wouldn't it be easier to either stick in the strings directly into JXcoreExtension.m in the method blocks where they are used or if that offends too much (I still think it makes the most sense) then how about defining a bunch of CONSTS in JXcoreExtension.m and just using them there. They don't seem to really have any value in this file.

In general we should look to reduce ceremony. That is, don't create abstractions just to create abstractions. It's like devs who will do things like define int fourOhFour = 404; instead of just typing in an int value of 404. I think it takes indirection and abstraction too far.

---

*[src/ios/AppContextTests.swift, line 14 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVA1egrGY3hY358y0w:-KPVA1ehDIyqmy8RALoQ:bgxpqoy) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContextTests.swift#L14)):*
> ```Swift
>     func testUpdateNetworkStatus() {
>         
>         class AppContextDelegateMock: NSObject, AppContextDelegate {
> ```

Rather than using a protocol for mocking why not just use inheritance and override? By using a protocol we essentially bringing back header files in all but name and increasing our ceremony and making our code harder to understand. If we just stick with classes then we have less ceremony and a more direct connection between the code as implementing and the code as tested.

---

*[src/ios/JXcoreExtension.m, line 14 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVAarGoSUqqB6hBiOm:-KPVAarGoSUqqB6hBiOn:brit1om) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L14)):*
> ```objc
> 
> // JXcoreExtension implementation.
> @interface JXcoreExtension (AppContextDelegate) <AppContextDelegate>
> ```

Per my comments in AppContext.swift I don't expect to see a protocol here. Rather I expect to see methods like sendPeerAvailabilityChanged that takes a string as an argument. This would be defined in this file and made available to be called in Swift.

---

*[src/ios/JXcoreExtension.m, line 27 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVB21W0IxpXkWWnehR:-KPVB21W0IxpXkWWnehS:b12tfbn) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L27)):*
> ```objc
> }
> 
> + (NSString *)objectToJSON:(NSObject *)object {
> ```

Why even have this logic here? In the end we are just sending a string. So why not just put all the logic to get to that string in the AppContext.swift file and leave here just the very stripped down ability to push a string into Node? We should aim to push as much of our logic as possible into Swift and do the absolute minimum here.

---

*[src/ios/JXcoreExtension.m, line 64 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVCP-RLkVpTEr1GZhe:-KPVCP-RLkVpTEr1GZhf:b2gfdcs) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L64)):*
> ```objc
>     [self defineDidRegisterToNative:appContext];
>     [self defineGetOSVersion:appContext];
>     [self defineExecuteNativeTests:appContext];
> ```

This MUST be wrapped in an ifdef. It should never be in our production code.

---

*[src/ios/JXcoreExtension.m, line 71 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVBNG9LePMKLo3GDQU:-KPVBNG9LePMKLo3GDQV:bp7r5at) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L71)):*
> ```objc
>          NSLog(@"jxcore: startListeningForAdvertisements");
> 
>          if ([appContext startListeningForAdvertisements]) {
> ```

Why not simplify this further such that the call to [appContext startListeningForAdvertisements] just returns whatever is supposed to go into 'withParams' and leave the logging and such to Swift? Again, we should strip this Objective-C down to the absolute bare minimum. I would pull out all the logging (do that in Swift) and any logic like this one. The functions should return what needs to go back to node and no more.

---

*[src/ios/JXcoreExtension.m, line 85 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV356G3W5mMpY1kaMd:-KPV356G3W5mMpY1kaMe:b-h533yy) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L85)):*
> ```objc
>              }
>          }
>      } withName:[AppContext startListeningForAdvertisements]];
> ```

Asking the dev to notice that appContext != AppContext and that therefore [appContext startListeningForAdvertisements] does something completely different than [AppContext startListeninigForAdvertisements] seems unnecessarily error prone. Please pick a visually more distinct way to distinguish between executing the method and getting the method's name. Or just stick the string here directly per my comment in AppContext.swift.

---

*[src/ios/JXcoreExtension.m, line 230 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVCrLYiMreraYLPP7r:-KPVCrLYiMreraYLPP7s:b-jg8740) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L230)):*
> ```objc
>     [JXcore addNativeBlock:^(NSArray * params, NSString *callbackId)
>      {
>          if (params.count != 2 || ![params[0] isKindOfClass:[NSString class]]) {
> ```

We should also move the param validation to swift if we can. Again, let's strip this file down to the minimum and do everything that can be done in Swift, in Swift. That way we centralize in a single language, with our logic all in one place.

---

*[src/ios/JXcoreExtension.m, line 249 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVD3qz1aXdbRhELpEy:-KPVD3qz1aXdbRhELpEz:b-k8ocj) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L249)):*
> ```objc
> }
> 
> - (void)defineExecuteNativeTests:(AppContext *)appContext
> ```

This needs to be wrapped in an IFDEF

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

baydet:
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[src/ios/AppContext.swift, line 29 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV5l8tXSIjFE_M7I8f:-KPWwm8bF7OLt3rA0qNr:b-jvbz4w) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L29)):*
> My understanding of what is going on is that we need a way for the swift code to call into Objective-C so that we can get access to JXcoreExtension.m. But why use a protocol? It really confused me in trying to read it.
> 
> What I expected was to find a method in this file called sendNetworkChanged. I expect that object to take a typed structure with member names like bluetoothLowEnergy, bluetooth, wifi, cellular and bssidName. Each would be typed as a bool.
> 
> I would then expect the sendNetworkChanged method in this file to call a sendNetworkChanged method in JXcoreExtension.m whose argument is a JSON string. Said string having been created by the sendNetworkChanged method in this file. I would expect this to happen  by having a bridging header that exposes JXcoreExtension.m to swift.
> 
> No protocols. Just normal APIs with normal straight forward names.
> 
> What did I miss?
</details>
One of the goals of using protocol and delegate pattern in this case is avoidance of usage JXcoreExtension.m in swift. 

There is another good reason for delegate pattern in this case. If we will see sendPeerAvailabilityChanged function in JXcoreExtension it will be harder to understand when and where it should be called by owner of JXcoreExtension object. Delegate pattern in this case tells us that:  "Hi, I'm App context. I know when network has changed, when peer availability have changed etc. If you want, you may listen for these changes"

 If we will move this functions to JXcoreExtension we are declaring something like that: "Hi, I'm JXcoreExtension. You may tell me anytime you like about peer availability changes, network changes etc. Or I can call it by myself anytime I like what may cause my internal state changes which you will not know about but what you may care about". And this makes JXcoreExtension class more complex what we should avoid.

---

*[src/ios/AppContext.swift, line 143 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV6qAVySq2uVSWiX3U:-KPWro6cJus89jtaBHzA:b-svxuw4) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L143)):*
> please call this multiConnect per the new spec
</details>
wouldn't you mind to keep it as it is for now. Because it already renamed to multiConnect in branch linked with #880. Or should I anyway cherrypick this changes to this branch? 

---

*[src/ios/AppContext.swift, line 182 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV7GojAT8eTetQXL5R:-KPWsihzrdj3sxfs5nEB:btg1ht2) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L182)):*
> There is an old joke in CS, there are only 3 numbers in CS, 0, 1 and infinity. If none of these functions is called more than once is it really worth creating this whole class and adding this level of indirection? Wouldn't it be easier to either stick in the strings directly into JXcoreExtension.m in the method blocks where they are used or if that offends too much (I still think it makes the most sense) then how about defining a bunch of CONSTS in JXcoreExtension.m and just using them there. They don't seem to really have any value in this file.
> 
> In general we should look to reduce ceremony. That is, don't create abstractions just to create abstractions. It's like devs who will do things like define int fourOhFour = 404; instead of just typing in an int value of 404. I think it takes indirection and abstraction too far.
</details>
I'm not big fan of code kind of this too. But for sure we should keep it AppContext.swift. Mostly because of `networkChanged` because it used in `didRegisterToNative` function of AppContext. And for consistency reasons this `networkChanged` const grab all other constant strings to AppContext.swift. 

I would like to make this code simpler but it's the only way to make const strings which will be visible in obj-c. Also I didn't want to increase complexity of AppContext.swift and create dependency between it and obj-c code with bridging headers.

---

*[src/ios/AppContextTests.swift, line 14 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVA1egrGY3hY358y0w:-KPWykllYjAlWKvS64qO:bu3dtza) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContextTests.swift#L14)):*
> Rather than using a protocol for mocking why not just use inheritance and override? By using a protocol we essentially bringing back header files in all but name and increasing our ceremony and making our code harder to understand. If we just stick with classes then we have less ceremony and a more direct connection between the code as implementing and the code as tested.
</details>
In this case maybe you're right. 

But in general protocol it's not the same as header files. For example, it's much easier to mock protocol than create inherit mock class and override some methods for tests because result of these methods could affect on original class behavior. So we need to dig down into this class guts to understand what it's doing. It's definitely increase complexity of mock class. In conclusion it's all about protocol oriented programming (POP). There is article with fancy images about why POP better than OOP - http://krakendev.io/blog/subclassing-can-suck-and-heres-why (kind of yellow title but anyway)

---

*[src/ios/JXcoreExtension.m, line 14 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVAarGoSUqqB6hBiOm:-KPX3TSErHzUAgulsZPr:b-z9g5ok) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L14)):*
> Per my comments in AppContext.swift I don't expect to see a protocol here. Rather I expect to see methods like sendPeerAvailabilityChanged that takes a string as an argument. This would be defined in this file and made available to be called in Swift.
</details>
detailed explanation in AppContext.swift

---

*[src/ios/JXcoreExtension.m, line 85 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV356G3W5mMpY1kaMd:-KPX72aNR6XSqf-qaDDk:b-bsczbu) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L85)):*
> Asking the dev to notice that appContext != AppContext and that therefore [appContext startListeningForAdvertisements] does something completely different than [AppContext startListeninigForAdvertisements] seems unnecessarily error prone. Please pick a visually more distinct way to distinguish between executing the method and getting the method's name. Or just stick the string here directly per my comment in AppContext.swift.
</details>
I'm going to move all these strings to separate class of AppContext.swift file if you don't mind

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

baydet:
Review status: all files reviewed at latest revision, 13 unresolved discussions.

---

*[src/ios/JXcoreExtension.m, line 64 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVCP-RLkVpTEr1GZhe:-KPY1rB8CggdRqMk8wok:b-7rl0jd) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L64)):*
> This MUST be wrapped in an ifdef. It should never be in our production code.
</details>
done.

---

*[src/ios/JXcoreExtension.m, line 85 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV356G3W5mMpY1kaMd:-KPY28-srascUwlT2nN7:b-35oifn) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L85)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> I'm going to move all these strings to separate class of AppContext.swift file if you don't mind
</details>
@yaronyg Are you OK with it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
PR is added to the queue for testing as 2. task. (5eb909b)

ThaliTester:
Test (Fail) 79896569 build is completed (5eb909b)

See https://github.com/ThaliTester/TestResults/tree/798965695eb909b__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (2103005)

baydet:
Review status: 59 of 62 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/ios/AppContext.swift, line 143 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV6qAVySq2uVSWiX3U:-KPYZ1yIAlKnX_6shYy7:b-nic59l) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L143)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> wouldn't you mind to keep it as it is for now. Because it already renamed to multiConnect in branch linked with #880. Or should I anyway cherrypick this changes to this branch? 
</details>
Apparently done.

---

*[src/ios/JXcoreExtension.m, line 27 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVB21W0IxpXkWWnehR:-KPYZ9yZt1yoBuynRJvn:bb7f2mf) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L27)):*
> Why even have this logic here? In the end we are just sending a string. So why not just put all the logic to get to that string in the AppContext.swift file and leave here just the very stripped down ability to push a string into Node? We should aim to push as much of our logic as possible into Swift and do the absolute minimum here.
</details>
done

---

*[src/ios/JXcoreExtension.m, line 71 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVBNG9LePMKLo3GDQU:-KPY3G2XQVJS_NNE_w7x:bizl1e7) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L71)):*
> Why not simplify this further such that the call to [appContext startListeningForAdvertisements] just returns whatever is supposed to go into 'withParams' and leave the logging and such to Swift? Again, we should strip this Objective-C down to the absolute bare minimum. I would pull out all the logging (do that in Swift) and any logic like this one. The functions should return what needs to go back to node and no more.
</details>
for now I'm going to remove all logging from JXcoreExtension.m unless we don't have any agreement on what we need to be logged and what logging library we will use - #49. 

---

*[src/ios/JXcoreExtension.m, line 230 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVCrLYiMreraYLPP7r:-KPYYu9GUqd1ok2qLoU1:b-7rl0jd) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L230)):*
> We should also move the param validation to swift if we can. Again, let's strip this file down to the minimum and do everything that can be done in Swift, in Swift. That way we centralize in a single language, with our logic all in one place.
</details>
done.

---

*[src/ios/JXcoreExtension.m, line 249 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVD3qz1aXdbRhELpEy:-KPY2AlHD-vNTG2ZMU7d:bb7f2mf) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L249)):*
> This needs to be wrapped in an IFDEF
</details>
done

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
Test (Fail) 79896569 build is completed (2103005)

See https://github.com/ThaliTester/TestResults/tree/798965692103005__819_added_basic_AppContext_functions_baydet/ for the logs

larryonoff:
Review status: 59 of 62 files reviewed at latest revision, 16 unresolved discussions.

---

*[src/ios/AppContext.swift, line 79 \[r13\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPmPgmOFKLXfkEu4pWK:-KPmPgmOFKLXfkEu4pWL:bv5z2a2) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/21030057b84f01bffec174e4e511ded2f6295d25/src/ios/AppContext.swift#L79)):*
> ```Swift
>     private var networkChangedRegistered: Bool = false
> 
>     private func updateNetworkStatus() {
> ```

All these methods just for raising events, so I recommend rename them into something like ```notifyOnDidUpdateNetworkStatus```.

---

*[src/ios/AppContext.swift, line 83 \[r13\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPmMpTZ9aCajJ5XxNcz:-KPmMpTZ9aCajJ5XxNd-:b-ltrj5y) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/21030057b84f01bffec174e4e511ded2f6295d25/src/ios/AppContext.swift#L83)):*
> ```Swift
>         do {
>             delegate?.context(self, didChangeNetworkStatus: try jsonValue([:])!)
>         } catch _ {
> ```

I assume that we should put assert in the catch if the case assumed to be unreachable.

---

*[src/ios/AppContext.swift, line 153 \[r13\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPmOuAmYZKVOz46btbd:-KPmOuAmYZKVOz46btbe:bc4ks71) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/21030057b84f01bffec174e4e511ded2f6295d25/src/ios/AppContext.swift#L153)):*
> ```Swift
> }
> 
> /// Node functions names
> ```

I think we need to expose node functions names in a calls with the name like `AppContextJSEvent`, but not the same as it is, because the current names are confusing.

I also recommend trying to replace these constants as class variables and not as methods if bridged into obj-c.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
PR is added to the queue for testing as 1. task. (f9189c1)

baydet:
Review status: 59 of 62 files reviewed at latest revision, 16 unresolved discussions.

---

*[src/ios/AppContext.swift, line 79 \[r13\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPmPgmOFKLXfkEu4pWK:-KPmcrOhUrjpb-ib0q0I:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/21030057b84f01bffec174e4e511ded2f6295d25/src/ios/AppContext.swift#L79)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> All these methods just for raising events, so I recommend rename them into something like ```notifyOnDidUpdateNetworkStatus```.
</details>
Done.

---

*[src/ios/AppContext.swift, line 83 \[r13\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPmMpTZ9aCajJ5XxNcz:-KPmd4iOj2AjacBQugD-:b-7rl0jd) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/21030057b84f01bffec174e4e511ded2f6295d25/src/ios/AppContext.swift#L83)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> I assume that we should put assert in the catch if the case assumed to be unreachable.
</details>
done.

---

*[src/ios/AppContext.swift, line 153 \[r13\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPmOuAmYZKVOz46btbd:-KPmcuvYAN0mmdaCA_AO:b-vqvdqs) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/21030057b84f01bffec174e4e511ded2f6295d25/src/ios/AppContext.swift#L153)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> I think we need to expose node functions names in a calls with the name like `AppContextJSEvent`, but not the same as it is, because the current names are confusing.
> 
> I also recommend trying to replace these constants as class variables and not as methods if bridged into obj-c.
</details>
Good point about class variables. Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

ThaliTester:
Test (Fail) 79896569 build is completed (f9189c1)

See https://github.com/ThaliTester/TestResults/tree/79896569f9189c1__819_added_basic_AppContext_functions_baydet/ for the logs

yaronyg:
Reviewed 1 of 6 files at r3, 3 of 3 files at r14.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/ios/AppContext.swift, line 29 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV5l8tXSIjFE_M7I8f:-KPoNAYSoSvrUnUB-f9w:beio8px) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L29)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> One of the goals of using protocol and delegate pattern in this case is avoidance of usage JXcoreExtension.m in swift. 
> 
> There is another good reason for delegate pattern in this case. If we will see sendPeerAvailabilityChanged function in JXcoreExtension it will be harder to understand when and where it should be called by owner of JXcoreExtension object. Delegate pattern in this case tells us that:  "Hi, I'm App context. I know when network has changed, when peer availability have changed etc. If you want, you may listen for these changes"
> 
>  If we will move this functions to JXcoreExtension we are declaring something like that: "Hi, I'm JXcoreExtension. You may tell me anytime you like about peer availability changes, network changes etc. Or I can call it by myself anytime I like what may cause my internal state changes which you will not know about but what you may care about". And this makes JXcoreExtension class more complex what we should avoid.
</details>
I believe you are describing a pub/sub model. But interestingly enough that isn't what we are building. We are building (in Swift) a publish only model. There is what's known as a sink (which is JXCoreExtension.m) where all events eventually have to go to be published. Over that we wrap AppContext.swift but for events it's just a translation layer that turns structure swift objects into JSON to be pushed to JXCoreExtension.m. So there shouldn't be any confusion on the part of the developer. In fact I would argue that the protocol causes confusion because it obscures the fact that there is one and only one place to publish an event, that's JXCoreExtension.m.

Now, if you want to subscribe to an event you don't do it in Swift. There literally should never be a reason for an event to be subscribed in Swift. Events only get subscribed in Node.

That's why I am suggesting the sendNetworkChanged method on both classes. They clearly show the flow of control for events. Swift pushes to Objective-C who publishes to node.

What do you think?

---

*[src/ios/AppContext.swift, line 182 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV7GojAT8eTetQXL5R:-KPoOIpkbBgFufyjU_iD:bqogs2c) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L182)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> I'm not big fan of code kind of this too. But for sure we should keep it AppContext.swift. Mostly because of `networkChanged` because it used in `didRegisterToNative` function of AppContext. And for consistency reasons this `networkChanged` const grab all other constant strings to AppContext.swift. 
> 
> I would like to make this code simpler but it's the only way to make const strings which will be visible in obj-c. Also I didn't want to increase complexity of AppContext.swift and create dependency between it and obj-c code with bridging headers.
</details>
Why would any of these strings ever be used anywhere but Objective-C?

---

*[src/ios/AppContextTests.swift, line 14 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVA1egrGY3hY358y0w:-KPoOykCCwIk_3oqpXT6:bx183td) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContextTests.swift#L14)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> In this case maybe you're right. 
> 
> But in general protocol it's not the same as header files. For example, it's much easier to mock protocol than create inherit mock class and override some methods for tests because result of these methods could affect on original class behavior. So we need to dig down into this class guts to understand what it's doing. It's definitely increase complexity of mock class. In conclusion it's all about protocol oriented programming (POP). There is article with fancy images about why POP better than OOP - http://krakendev.io/blog/subclassing-can-suck-and-heres-why (kind of yellow title but anyway)
</details>
Everybody gets their opinion, including the guy who wrote that article. But it doesn't make them right. :) (And yes, the same applies to me)

Personally I hate OO because I find I spend more time on my classes than my code and I don't get much for it. That's why I've always been more of a functional programmer. 

And in a sense I agree with the author of that article to the extent that I don't think we should enter an OO fest. But I also think that protocol oriented programming is just another name for header file based programming. It's about abstractions for the sake of abstractions. In functional programmer we just pass around the values and call it a day. We don't spend a ton of time defining interfaces.

In this case my suggestion to use subclassing isn't as part of production coding. In other words I'm not suggesting you jump into the OO pit. Rather my suggestion is based on a trick to do mocking.

In terms of your concern regarding mocking. In general I come from the school that mocking should be all or nothing. And if you are mocking all the methods in a class than you just don't call super and you don't care about the internal implementation. :)

And yes, I did notice that you changed to subclassing. :)

---

*[src/ios/JXcoreExtension.m, line 71 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVBNG9LePMKLo3GDQU:-KPoQlxoAp156MLk46VL:bhqu9mm) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L71)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> for now I'm going to remove all logging from JXcoreExtension.m unless we don't have any agreement on what we need to be logged and what logging library we will use - #49. 
</details>
You should still follow your instincts. If there is an unusual event or problem that we need to log then you should log it and we can clean things up when we get to #49. But trying to add logging later is a bad idea because so much context is lost that we are likely to miss the thing that need to be logged. But I'd still generally be looking to do the logging in swift.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

baydet:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/ios/AppContext.swift, line 29 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV5l8tXSIjFE_M7I8f:-KPqZN82vBKnWZpA9VrT:b9pe4ii) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L29)):*
>In fact I would argue that the protocol causes confusion because it obscures the fact that there is one and only one place to publish an event, that's JXCoreExtension.m.
Right, I understand that we will have only one place to publish event. Because we have no need to have multiple listeners for AppContext's changes.

>Now, if you want to subscribe to an event you don't do it in Swift. There literally should never be a reason for an event to be subscribed in Swift. Events only get subscribed in Node.
I'm sorry. I didn't get what you mean by that.

And you know, now I'm a little bit confused because if we will look into possible implementation of your approach we will have almost the same interfaces. Anyway we will have to write analogue of AppContextDelegate because we have no direct access to interface of  JXcoreExtension (as it's part of Cordova generated project). In that case we will have to write category on JXcoreExtension: create additional header file, create bridging header. Now we have access to this class in swift file. So we will have a lot of additional files that adds complexity and increases coupling between JXcoreExtension and AppContext.

---

*[src/ios/AppContext.swift, line 182 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV7GojAT8eTetQXL5R:-KPq9XMGE1Yt4wmd2T4Y:b-sw0bsf) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L182)):*
> Why would any of these strings ever be used anywhere but Objective-C?
</details>
We are waiting for `networkChanged` registration and after that firing `networkChanged` event

---

*[src/ios/JXcoreExtension.m, line 71 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPVBNG9LePMKLo3GDQU:-KPqBp_ty8sDkMAEmAt0:b-haj4f4) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/JXcoreExtension.m#L71)):*
> You should still follow your instincts. If there is an unusual event or problem that we need to log then you should log it and we can clean things up when we get to #49. But trying to add logging later is a bad idea because so much context is lost that we are likely to miss the thing that need to be logged. But I'd still generally be looking to do the logging in swift.
</details>
I agree on only-swift logging. I think I've simplified JXcoreExtension to the lowest level.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

baydet:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[src/ios/AppContext.swift, line 29 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV5l8tXSIjFE_M7I8f:-KPrxkTsl0AdqO27oBA9:bu0cafi) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L29)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> >In fact I would argue that the protocol causes confusion because it obscures the fact that there is one and only one place to publish an event, that's JXCoreExtension.m.
> Right, I understand that we will have only one place to publish event. Because we have no need to have multiple listeners for AppContext's changes.
> 
> >Now, if you want to subscribe to an event you don't do it in Swift. There literally should never be a reason for an event to be subscribed in Swift. Events only get subscribed in Node.
> I'm sorry. I didn't get what you mean by that.
> 
> And you know, now I'm a little bit confused because if we will look into possible implementation of your approach we will have almost the same interfaces. Anyway we will have to write analogue of AppContextDelegate because we have no direct access to interface of  JXcoreExtension (as it's part of Cordova generated project). In that case we will have to write category on JXcoreExtension: create additional header file, create bridging header. Now we have access to this class in swift file. So we will have a lot of additional files that adds complexity and increases coupling between JXcoreExtension and AppContext.
</details>
@yaronyg, sorry for this quoting hell. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

yaronyg:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[src/ios/AppContext.swift, line 29 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV5l8tXSIjFE_M7I8f:-KPsfUSFkUfQFGuTN6hb:b-9ot6ok) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L29)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> @yaronyg, sorry for this quoting hell. 
</details>
No problem with quoting hell and we should put this on the stack tomorrow so we can drive it to close. But the way we handled this in Android is that everything related to JXCoreExtension lives in ThaliCordova plugin. This would include appcontext. What is in the framework is generic functionality that has nothing to do with the Node apis per se. So I am arguing for the bridging header but I'm arguing that lives in Thali_CordovaPlugin along with appcontext and that appcontext is the one who calls the framework

---

*[src/ios/AppContext.swift, line 182 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-KPV7GojAT8eTetQXL5R:-KPsghO_YDiuevmt-7pE:b2dy4br) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/992243f3c52ecbb94499ebb285f913b9e1298fff/src/ios/AppContext.swift#L182)):*
<details><summary><i>Previously, baydet (Alexander Evsyuchenya) wrote…</i></summary>
> We are waiting for `networkChanged` registration and after that firing `networkChanged` event
</details>
But how does that involve the string above? You don't need the string to catch the event. The event is a callback. I'm missing something. Again, let's put this on the stack.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826)*

yaronyg:
:lgtm:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/826#-:-KPxCz_CSca_TjaLOZvS:bnfp4nl)*

ThaliTester:
PR is added to the queue for testing as 1. task. (6ccafc0)

ThaliTester:
Test (Fail) 79896569 build is completed (6ccafc0)

See https://github.com/ThaliTester/TestResults/tree/798965696ccafc0__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 2. task. (3cf79be)

ThaliTester:
Test (Fail) 79896569 build is completed (3cf79be)

See https://github.com/ThaliTester/TestResults/tree/798965693cf79be__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 4. task. (23d4b17)

ThaliTester:
Test (Success) 79896569 build is completed (23d4b17)

See https://github.com/ThaliTester/TestResults/tree/7989656923d4b17__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
Test 7989656923d4b17(23d4b17) has successfully finished without an error

See https://github.com/ThaliTester/TestResults/tree/7989656923d4b17__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 5. task. (b60016a)

ThaliTester:
Test (Fail) 79896569 build is completed (b60016a)

See https://github.com/ThaliTester/TestResults/tree/79896569b60016a__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 5. task. (f437b90)

ThaliTester:
Test (Fail) 79896569 build is completed (f437b90)

See https://github.com/ThaliTester/TestResults/tree/79896569f437b90__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 3. task. (f32c83a)

ThaliTester:
Test (Fail) 79896569 build is completed (f32c83a)

See https://github.com/ThaliTester/TestResults/tree/79896569f32c83a__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 6. task. (22b1f5d)

ThaliTester:
Test (Fail) 79896569 build is completed (22b1f5d)

See https://github.com/ThaliTester/TestResults/tree/7989656922b1f5d__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 3. task. (ef47422)

yaronyg:
Ilya is working on this bug because we aren't getting it passing in CI.

larryonoff:
I just made a mad thing and committed into the branch vNext_aevsyuchenya_819. So I believe we'll have fix with the next CI run.

ThaliTester:
Test (Success) 79896569 build is completed (ef47422)

See https://github.com/ThaliTester/TestResults/tree/79896569ef47422__819_added_basic_AppContext_functions_baydet/ for the logs

ThaliTester:
PR is added to the queue for testing as 6. task. (fbe8035)

ThaliTester:
Test 79896569ef47422(ef47422) has failed

See https://github.com/ThaliTester/TestResults/tree/79896569ef47422__819_added_basic_AppContext_functions_baydet/ for the fail logs

ThaliTester:
Test (Success) 79896569 build is completed (fbe8035)

See https://github.com/ThaliTester/TestResults/tree/79896569fbe8035__819_added_basic_AppContext_functions_baydet/ for the logs

Issue #838, opened 2016-08-07
Thali Manager is designed to be stopped anytime.


andrew-aladev:
There is a method to force empty replications: bump `update_seq` of Pouch DB. This is just a hack but it doesn't matter. We will obtain many requests to:

1. `/NotificationBeacons`
2. `db/_changes`
3. `db/_bulk_docs`
4. `db/_local/something`

Calls can looks like `1, 3, 1, 4, 2, 2, 1, 4, 3, 4, 3, 2`. I was just wondering what will be if we will try to stop these requests when each request has been started at least once. For example:

`1, 3, 1, 4, 2, <stop here!> 2, 1, 4, 3, 4, 3, 2`

Please checkout my branch `vNext_aaladev_507`. There is a test.

1. `cd test/www/jxcore`
2. `DEBUG=* jx runCoordinatedTests.js --filter=bv_tests/disabled/testThaliManagerStopCoordinated.js`

You will obtain something like:

```
DEBUG thaliReplicationPeerAction: Got error on replication -  500 connect ECONNREFUSED

Uncaught Promise Rejection: Error: connect ECONNREFUSED

Trace: { [Error: connect ECONNREFUSED]
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  status: 500,
  result: 
   { ok: false,
     start_time: Sun Aug 07 2016 22:06:48 GMT+0300 (MSK),
     docs_read: 0,
     docs_written: 0,
     doc_write_failures: 0,
     errors: [ [Circular] ],
     status: 'aborting',
     end_time: Sun Aug 07 2016 22:06:49 GMT+0300 (MSK),
     last_seq: 0 } }
    at process.tests (/Users/hiqouser/workspace/Thali_CordovaPlugin/test/www/jxcore/lib/thaliTape.js:38:11)
    at process.emit (events.js:85:17)
    at Array.1 (/Users/hiqouser/workspace/Thali_CordovaPlugin/test/www/jxcore/node_modules/lie/lib/index.js:128:19)
    at nextTick (/Users/hiqouser/workspace/Thali_CordovaPlugin/test/www/jxcore/node_modules/immediate/lib/index.js:61:18)
    at process._tickCallback (node.js:917:13)

Error: connect ECONNREFUSED
    at errnoException (net.js:837:11)
    at Object.afterConnect [as oncomplete] (net.js:829:19)
[...]

PR #840, opened 2016-08-08
Implemented thali manager

### LINKED ISSUE #507: Implement thaliManager
- [ ] Clone the vNext_yarong_417_1 branch then 'git checkout -B vNext_andrew-aladev_507'
- [ ] Implement #773, you'll need it
- [ ] Update comment on PouchDB argument to ThaliManager to point at the new utility function from #773 and let the user know that unless they have a good reason not to, they need to use it to wrap the version of PouchDB they are using.
- [ ] Change the code (and the comments) so that if the user doesn't submit their own thailPeerPoolInterface we fail.
- [ ] In the constructor make sure that we set up salti as the first router configured on our express router object so we know all requests will go through it first. Then set up the expressPouchDB object that was passed in for the BASE_DB_PATH with 'minimumForPouchDB'. The second step is already in the code.
- [ ] The constructor also needs to setup thaliSendNotificationBasedOnReplication and thaliPullReplicationFromNotification. Both are already in the constructor. thaliPullReplicationFromNotification will handle setting up thaliNotificationClient. And thaliSendNotificationBasedOnReplication will handle setting up thaliNotificationServer. That is why we only need the two top level replication classes. They handle everything else.
- [ ] The start function has a bunch of jobs to do. First it will start up pull replication from notification. This doesn't do anything because the underlying network isn't started yet but we don't want a race condition so we start it first. Then we can start the network layer by calling ThaliMobile.start followed by starting listening for Advertisements (e.g. people announcing themselves) and startUpdateAdvertisingAndListening (where we advertise ourselves) finally followed by starting thaliSendNotificationBasedOnReplication which depends on things like advertising ourselves being turned on (since it's whole purpose is to update stuff like that).
- [ ] Stop is easier because we just shut down the two replication objects and we have a single stop that shuts everything down for ThaliMobile.
- [ ] We can do a local test with a bogus PSK request to make sure Salti is on the job.
- [ ] Do a test with a legal PSK request (e.g. a PSK from an actual authorized peer) but for a URL they shouldn't have access to just to make sure we are using the right config of salti.
- [ ] If you feel like it you can use proxyquire to mock out the underlying objects and write unit tests to make sure we call the right objects for start and stop but I'm not sure that is really worth doing.
- [ ] For this class the much more interesting tests are the coordinated ones. The first test is the repeat write test. In this test during setup we will share our public key via the data option (see line 29 of testThaliPullReplicationFromNotificationCoordinated). In the test itself create a local DB with some well known name and write to a record of the form { _id: (our public key), test1: true }. Then register for changes on the DB and make sure to include the doc. Then call start on the thaliManager and pass in the public keys of all the other peers using the testUtils.turnParticipantsIntoBufferArray (line 120 of testThaliPullReplicationFromNotificationCoordinated). What you are looking for is that you see changes with docs with the IDs of all your peers with test1: true. Once that happens then you start the second stage of the test. In the second stage you will update your local record in the DB with { _id: (out public key), test1: true, test2: true }. Now create a new changes tracker (make sure to cancel the previous one!!!) and this one is looking for records of all the peers again (make sure to set your new changes tracker with the PouchDB option "since: 0") but you are looking to see test2: true. Once you see that then you are done with the test and can exit. In teardown make sure we call stop.
- [ ] For the next test you are actually going to continue the previous test. You MUST use the exact same thaliManager object you used from the first test. That is the whole point of this test. You are going to use the same DB and you are going to update our favorite record with { _id: (out public key), test1: true, test2: true, test3: true }. The rest of the test looks like the first one except you are just waiting to find a record for each of the other peers with test3: true. What this test shows is that we can stop the whole system (which happened in teardown of the first test) and then restart the whole system and everything still works with the exactly same thaliManager object. This shows that the object properly cleans itself up and restarts itself.
- [ ] Make sure both of the coordinated tests have a timer that will kill the test if it runs too long.

Note that the previous two tests are just variants on the testThaliPullReplicationFromNotificationCoordinated. There are lots of subtleties I don't really go into above such as how the changes tracker works, about including docs, live changes tracking, since 0 (which I at least mention), etc. All of that is in the testThaliPullReplicationFromNotificationCoordinated test.

andrew-aladev:
Closes #507

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/840)


ThaliTester:
PR is added to the queue for testing as 2. task. (b4a17a0)

ThaliTester:
Test (Fail) 80473891 build is completed (b4a17a0)

See https://github.com/ThaliTester/TestResults/tree/80473891b4a17a0_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
Reviewed 20 of 20 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[test/www/jxcore/package.json, line 54 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgZlWjNeHc29cYIFR3:-KOgZGwCUzBl2UB3NtZ4:b-9etkk8) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/package.json#L54)):*
> ```JSON
>     "urlsafe-base64": "1.0.0",
>     "uuid": "2.0.0",
>     "salti": "https://github.com/thaliproject/salti.git#vNext_aaladev_10"
> ```

Why not just publish Salti to NPM and get it from there? Please send me your account name on npmjs.com and I'll give you permission to publish.

---

*[test/www/jxcore/runCoordinatedTests.js, line 113 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOg_DJYHVenefe0916n:-KOg_DJYHVenefe0916o:bon7vhv) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/runCoordinatedTests.js#L113)):*
> ```JavaScript
>   if (argv.filter) {
>     instanceArgs.push(argv.filter);
>   } if (argv._) {
> ```

We already have that ability. Just stick '.only' on a test. I'm a bit nervous about us introducing it twice.

---

*[test/www/jxcore/runCoordinatedTests.js, line 116 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOg_8Huu6o71nSc245w:-KOg_8Huu6o71nSc245x:bfsivsa) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/runCoordinatedTests.js#L116)):*
> ```JavaScript
>     // Any argument that didn't have an option
>     // associated with it can be bypassed.
>     // This adds an ability to run single test.
> ```

NB:/adds an ability/adds the ability/

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 154 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgbzdONvu_93_2KSrg:-KOgbzdONvu_93_2KSrh:b-f3knwk) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L154)):*
> ```JavaScript
> }
> 
> function checkExpressPouchDB(t, mocks) {
> ```

It seems like we could use a utility function of the form checkArgs([[argvalue, description]]). It would check calledOnce, check the number of args and then check that the listed args were passed in.

Also in the code below we don't seem to care about the order in which the args are passed in. That seems wrong. Shouldn't we know which arg value belongs in which position?

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 155 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgb3Y2n-I7miLocDKo:-KOgb3Y3eSjeaoOx7IRZ:b-jexuqz) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L155)):*
> ```JavaScript
> 
> function checkExpressPouchDB(t, mocks) {
>   // Testing that 'expressPouchDB' has called properly.
> ```

NB: /has/was/ (this applies both here and below)

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 157 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgbRcp_IwPW03izu1S:-KOgb8abygiU51lveYYu:b-764af1) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L157)):*
> ```JavaScript
>   // Testing that 'expressPouchDB' has called properly.
>   t.ok(mocks.expressPouchDB.called, 'expressPouchDB has called');
>   t.ok(mocks.expressPouchDB.calledOnce, 'expressPouchDB has called once');
> ```

NB: /has/was/Why do you need both called and calledOnce? If calledOnce is false then we are in a bad place and should fail so why do we need the first check for called?

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 447 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgciGyEiYOBjUwjPxU:-KOgciGyEiYOBjUwjPxV:b727zw6) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L447)):*
> ```JavaScript
>   );
> 
>   checkExpressPouchDB(t, mocks);
> ```

Since we use this group of functions several times wouldn't it be easier to just create a single 'check' method that calls all of these sub-methods?

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 563 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgdc1hjoDgZHDhR00Q:-KOgdc1hjoDgZHDhR00R:b-bnhuu6) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L563)):*
> ```JavaScript
>       thaliManager.stop(),
>       thaliManager.stop(),
>       thaliManager.stop()
> ```

Shouldn't there be checks in here some place to make sure that these start and stops didn't cause damage?

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 64 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgfuQqBYfs5e9JlFKw:-KOgfuQrqUdnhJvMbc6q:b-cayuy2) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L64)):*
> ```JavaScript
>     // We are not removing defaultDirectory.
>     // Tests should resolve it by itself.
>     // fs.removeSync(defaultDirectory);
> ```

The point of having the two tests run one after each other is that we would call stop after the first test and then start on the second test. This would then test our behavior after a full stop and start. But here we only call stop at the very end.

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 81 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgdu2tBZa8te_qdB_W:-KOgdu2tBZa8te_qdB_X:b-lo8c95) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L81)):*
> ```JavaScript
> });
> 
> function validateDocs(pouchDB, docs, keys) {
> ```

Shouldn't we also check that we only have the exact number of docs we expect? Otherwise if we have some strange bug that introduces extraneous docs we wouldn't detect it since the docs we are looking for are there.

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 113 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgeNqr6uFDjxPC1kM-:-KOgeNqr6uFDjxPC1kM0:b-xan7i8) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L113)):*
> ```JavaScript
>       if (findDoc(change.doc)) {
>         if (allDocsFound) {
>           // We should turn 'changesFeed' off and call 'resolve' now.
> ```

NB: I suspect this comment is out of date and you are now waiting for 'complete' to call resolve. I would suggest removing the comment.

---

*[thali/NextGeneration/thaliConfig.js, line 15 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgiQhixDgkzuvcB6Nf:-KOgiQhixDgkzuvcB6Ng:b-wju72c) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/thali/NextGeneration/thaliConfig.js#L15)):*
> ```JavaScript
>   BEACON_MILLISECONDS_TO_EXPIRE: 60 * 60 * 1000,
>   BASE_DB_PATH: '/db',
>   BASE_DB_PREFIX: './db/',
> ```

If i'm reading the code correctly then this value is only needed for testing, not for production. So we should keep it somewhere other than thaliConfig. thaliConfig is just for production values.

---

*[thali/NextGeneration/thaliManager.js, line 94 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgisEmElOXE-ODEXoi:-KOgisEmElOXE-ODEXoj:bav8byb) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/thali/NextGeneration/thaliManager.js#L94)):*
> ```JavaScript
>     function (thaliId, req) {
>       /*
>        * In the case of an id that begins with te string [LOCAL_SEQ_POINT_PREFIX] we MUST enforce
> ```

NB: /te/the

---

*[thali/NextGeneration/thaliManager.js, line 98 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgivkQVbqycT_HvYFi:-KOgivkQVbqycT_HvYFj:b-lfhcsd) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/thali/NextGeneration/thaliManager.js#L98)):*
> ```JavaScript
>        * caller as generated by thaliNotificationBeacons.createPublicKeyHash.
>        */
>       return thaliId == self._getPskIdToPublicKey(req.connection.pskIdentity);
> ```

Shouldn't this be ===? Otherwise there are some potential attacks using type coercion that I'd rather not think about. Or maybe not, I'm not 100% sure.

---

*[thali/NextGeneration/thaliManager.js, line 134 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgjGcxDz9Af-qE0l3o:-KOgjGcxDz9Af-qE0l3p:bwjto3m) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/thali/NextGeneration/thaliManager.js#L134)):*
> ```JavaScript
> }
> 
> ThaliManager.STATES = {
> ```

Do anyone debugging this thing a favor and don't use numbers, use strings instead.

---

*[thali/NextGeneration/thaliManager.js, line 144 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgjNnZBbm0Oqt343xv:-KOgjNnZBbm0Oqt343xw:b-b610fh) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/thali/NextGeneration/thaliManager.js#L144)):*
> ```JavaScript
> /**
>  * Starts up everything including listening for advertisements, sending out
>  * own advertisements, generating beacons to notify peers that we have data
> ```

NB: /own/our

---

*[thali/NextGeneration/thaliManager.js, line 175 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgjbcPlh5oD5bHo-25:-KOgjbcPlh5oD5bHo-26:b-6s7u41) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/thali/NextGeneration/thaliManager.js#L175)):*
> ```JavaScript
>     }
>   }
>   this.state = ThaliManager.STATES.STARTING;
> ```

This seems like a good place for an assert saying we have to be in CREATED or STOPPED to have reached here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840)*

ThaliTester:
PR is added to the queue for testing as 2. task. (1509f0e)

ThaliTester:
Test (Fail) 80473891 build is completed (1509f0e)

See https://github.com/ThaliTester/TestResults/tree/804738911509f0e_Implemented_thali_manager_andrew-aladev/ for the logs

andrew-aladev:
> Why not just publish Salti to NPM and get it from there? Please send me your account name on npmjs.com and I'll give you permission to publish.

https://www.npmjs.com/~andrew-aladev

ThaliTester:
PR is added to the queue for testing as 3. task. (03d5d83)

andrew-aladev:
I've merged a small commit from vNext for CI.

ThaliTester:
Test (Fail) 80473891 build is completed (03d5d83)

See https://github.com/ThaliTester/TestResults/tree/8047389103d5d83_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
Reviewed 2 of 2 files at r2, 7 of 7 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[test/www/jxcore/package.json, line 54 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgZlWjNeHc29cYIFR3:-KOkcVcKDjMJ3B7v-vSY:b-g1gyyi) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/package.json#L54)):*
> Why not just publish Salti to NPM and get it from there? Please send me your account name on npmjs.com and I'll give you permission to publish.
</details>
I still think we should just publish this to NPM instead of using a Git URL but it's not worth worrying about now.

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 563 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgdc1hjoDgZHDhR00Q:-KOkdBG1sMQzRmBBQPkE:bl28bix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L563)):*
> Shouldn't there be checks in here some place to make sure that these start and stops didn't cause damage?
</details>
I'm still not clear on this one. Stop gets called a bunch of times but then what? If it doesn't throw that's o.k.? Shouldn't we check that the system is still functioning after all those stops?

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 64 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgfuQqBYfs5e9JlFKw:-KOkdRG1HTIblK5zUHOV:b-4ujgb4) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L64)):*
> The point of having the two tests run one after each other is that we would call stop after the first test and then start on the second test. This would then test our behavior after a full stop and start. But here we only call stop at the very end.
</details>
I take it that you don't agree with my comment because the code will still only call stop after the last tests runs instead of after the first. Could you please share your reasoning?

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 81 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgdu2tBZa8te_qdB_W:-KOkm6EJX_3b-iYdb45L:b3kre) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L81)):*
> Shouldn't we also check that we only have the exact number of docs we expect? Otherwise if we have some strange bug that introduces extraneous docs we wouldn't detect it since the docs we are looking for are there.
</details>
?

---

*[test/www/jxcore/lib/testUtils.js, line 697 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOkmKDdpIl2xQd5Z1dA:-KOkmKDdpIl2xQd5Z1dB:b-4fag8h) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/03d5d8355c82b054f8c40f1e189b74c882375ed3/test/www/jxcore/lib/testUtils.js#L697)):*
> ```JavaScript
> 
>   args.forEach(function (arg, index) {
>     // If arg is not defined we wont check it.
> ```

The APIs we are checking have a fixed number of arguments, typically mandatory. So shouldn't we validate that we have exactly the number, type and position of args we are expecting?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840)*

ThaliTester:
PR is added to the queue for testing as 1. task. (68c6a44)

andrew-aladev:
> I still think we should just publish this to NPM instead of using a Git URL but it's not worth worrying about now.

I've increased the salti version but npm publish said: 'you do not have permission to publish'. My profile is https://www.npmjs.com/~andrew-aladev

> I'm still not clear on this one. Stop gets called a bunch of times but then what? If it doesn't throw that's o.k.? Shouldn't we check that the system is still functioning after all those stops?

fixed

> I take it that you don't agree with my comment because the code will still only call stop after the last tests runs instead of after the first. Could you please share your reasoning?

This was written in docs:

> For the next test you are actually going to continue the previous test. You MUST use the exact same thaliManager object you used from the first test.

But I can't stop thali manager after each test because of issue #838. I've added a comment. Please uncomment related if: you will not able to start second test, it will fail after first one. This looks magic.

> Shouldn't we also check that we only have the exact number of docs we expect? Otherwise if we have some strange bug that introduces extraneous docs we wouldn't detect it since the docs we are looking for are there.

fixed

> The APIs we are checking have a fixed number of arguments, typically mandatory. So shouldn't we validate that we have exactly the number, type and position of args we are expecting?

fixed

ThaliTester:
PR is added to the queue for testing as 2. task. (998245f)

ThaliTester:
Test (Fail) 80473891 build is completed (68c6a44)

See https://github.com/ThaliTester/TestResults/tree/8047389168c6a44_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (998245f)

See https://github.com/ThaliTester/TestResults/tree/80473891998245f_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
Andrew it makes it really hard to do a review when comments get thrown all over the place and so we lose code context, we lose history, etc. Please respond to comments inline with the code and not here in the main thread. Reviewable.io makes this super easy. I'm going to copy your comments into context and I would appreciate it if from now on you can respond to them in the same context.

---

Reviewed 2 of 3 files at r5, 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/www/jxcore/bv_tests/testThaliManager.js, line 563 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgdc1hjoDgZHDhR00Q:-KOquRM2MY8Aa1X-W9XB:bjtirwk) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManager.js#L563)):*
> I'm still not clear on this one. Stop gets called a bunch of times but then what? If it doesn't throw that's o.k.? Shouldn't we check that the system is still functioning after all those stops?
</details>
Andrew: fixed
Me: Agreed

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 64 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgfuQqBYfs5e9JlFKw:-KOquj4VT8oW2wPlYCry:buds0hm) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L64)):*
> I take it that you don't agree with my comment because the code will still only call stop after the last tests runs instead of after the first. Could you please share your reasoning?
</details>
Andrew said: 

This was written in docs:

For the next test you are actually going to continue the previous test. You MUST use the exact same thaliManager object you used from the first test.

But I can't stop thali manager after each test because of issue #838. I've added a comment. Please uncomment related if: you will not able to start second test, it will fail after first one. This looks magic.

Yaron said:
There is no magic. We should be able to use the same instance of thaliManager for both tests. This means that #838 is a show stopper and this PR can't go in until it is fixed. This is actually a really big deal because in production we will constantly be starting and stopping the thaliManager and expecting it to continue functioning. It is literally impossible to run Thali on iOS without this functionality. This test's failure when we re-use the object means that we can't use thaliManager in its most important use case. :(

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 81 \[r1\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOgdu2tBZa8te_qdB_W:-KOr3Um8p-Ze4MY2N8Ix:b-lonyh5) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/b4a17a0694a29d06cf1104c62e55b6e8b5fc5d0e/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L81)):*
> ?
</details>
Andrew said: fixed

---

*[test/www/jxcore/bv_tests/testThaliManagerCoordinated.js, line 143 \[r6\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOr4Cro9upZZ7wsnIFC:-KOr4Cro9upZZ7wsnIFD:b-rxgmyl) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/998245fc3e113db8cc50d24a0003dfe15aeee8ed/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js#L143)):*
> ```JavaScript
>       } else {
>         changesFeed.cancel();
>         reject('inalid doc');
> ```

NB: /inalid/invalid

---

*[test/www/jxcore/lib/testUtils.js, line 697 \[r4\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KOkmKDdpIl2xQd5Z1dA:-KOr3bDJGQMjmBeOCNO0:b-lonyh5) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/03d5d8355c82b054f8c40f1e189b74c882375ed3/test/www/jxcore/lib/testUtils.js#L697)):*
> The APIs we are checking have a fixed number of arguments, typically mandatory. So shouldn't we validate that we have exactly the number, type and position of args we are expecting?
</details>
Andrew said: fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-:-KOqta4UPt_2jBOTbI3m:b-8mu2zj)*

ThaliTester:
PR is added to the queue for testing as 2. task. (8663c98)

ThaliTester:
PR is added to the queue for testing as 3. task. (4534d0f)

ThaliTester:
Test (Fail) 80473891 build is completed (8663c98)

See https://github.com/ThaliTester/TestResults/tree/804738918663c98_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (4534d0f)

See https://github.com/ThaliTester/TestResults/tree/804738914534d0f_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
It should be done today.

ThaliTester:
PR is added to the queue for testing as 1. task. (ee5bed5)

ThaliTester:
Test (Fail) 80473891 build is completed (ee5bed5)

See https://github.com/ThaliTester/TestResults/tree/80473891ee5bed5_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (78b34bc)

ThaliTester:
PR is added to the queue for testing as 2. task. (af9a2a4)

ThaliTester:
Test (Fail) 80473891 build is completed (78b34bc)

See https://github.com/ThaliTester/TestResults/tree/8047389178b34bc_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (af9a2a4)

See https://github.com/ThaliTester/TestResults/tree/80473891af9a2a4_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (b7a58df)

ThaliTester:
Test (Fail) 80473891 build is completed (b7a58df)

See https://github.com/ThaliTester/TestResults/tree/80473891b7a58df_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (58cbebb)

ThaliTester:
Test (Fail) 80473891 build is completed (58cbebb)

See https://github.com/ThaliTester/TestResults/tree/8047389158cbebb_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (18c69b6)

ThaliTester:
PR is added to the queue for testing as 2. task. (712e90f)

ThaliTester:
Test (Fail) 80473891 build is completed (18c69b6)

See https://github.com/ThaliTester/TestResults/tree/8047389118c69b6_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (712e90f)

See https://github.com/ThaliTester/TestResults/tree/80473891712e90f_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (b9a1fd0)

ThaliTester:
PR is added to the queue for testing as 2. task. (fd8ce66)

ThaliTester:
Test (Fail) 80473891 build is completed (b9a1fd0)

See https://github.com/ThaliTester/TestResults/tree/80473891b9a1fd0_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (fd8ce66)

See https://github.com/ThaliTester/TestResults/tree/80473891fd8ce66_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (d327c14)

ThaliTester:
Test (Fail) 80473891 build is completed (d327c14)

See https://github.com/ThaliTester/TestResults/tree/80473891d327c14_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
We have a problem with how we stop PouchDB. We need to both stop the replication and listen for the error and complete events or we will get an unhandled error.

ThaliTester:
PR is added to the queue for testing as 2. task. (c4ac6df)

ThaliTester:
Test (Fail) 80473891 build is completed (c4ac6df)

See https://github.com/ThaliTester/TestResults/tree/80473891c4ac6df_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (d6262b1)

ThaliTester:
Test (Fail) 80473891 build is completed (d6262b1)

See https://github.com/ThaliTester/TestResults/tree/80473891d6262b1_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (a212f99)

ThaliTester:
PR is added to the queue for testing as 2. task. (b8e09da)

ThaliTester:
Test (Fail) 80473891 build is completed (a212f99)

See https://github.com/ThaliTester/TestResults/tree/80473891a212f99_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (b8e09da)

See https://github.com/ThaliTester/TestResults/tree/80473891b8e09da_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (27fc5ae)

ThaliTester:
Test (Fail) 80473891 build is completed (27fc5ae)

See https://github.com/ThaliTester/TestResults/tree/8047389127fc5ae_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
Reviewed 23 of 23 files at r7.

---

*[test/www/jxcore/bv_tests/testThaliPullReplicationFromNotificationCoordinated.js, line 95 \[r7\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KPoZVteuy-FU5u2rcha:-KPoZVteuy-FU5u2rchb:ba5ku89) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/27fc5ae157c7899e0a73e0d43ac1a6beacca98a3/test/www/jxcore/bv_tests/testThaliPullReplicationFromNotificationCoordinated.js#L95)):*
> ```JavaScript
>     changes && changes.cancel();
>     cancelTimer && clearTimeout(cancelTimer);
>     thaliPeerPoolDefault.stop()
> ```

Are we guaranteed that stop can never call reject()?

---

*[thali/NextGeneration/replication/thaliPullReplicationFromNotification.js, line 269 \[r7\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KPob2uuItWNURtpBBLK:-KPob2uuItWNURtpBBLL:baxwkug) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/27fc5ae157c7899e0a73e0d43ac1a6beacca98a3/thali/NextGeneration/replication/thaliPullReplicationFromNotification.js#L269)):*
> ```JavaScript
>   // We can just make peerDictionary empty.
>   this._peerDictionary = {};
>   this._stoppingPromise = this._thaliPeerPoolInterface.stop()
> ```

Just because we stopped thaliPullReplicationFromNotification does that necessarily mean we have to stop the thaliPeerPoolInterface? For example, someone may disable replication but want to still do identity exchange (which required the thaliPeerPoolInterface to not be stopped).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840)*

ThaliTester:
PR is added to the queue for testing as 2. task. (c88e3bc)

ThaliTester:
Test (Fail) 80473891 build is completed (c88e3bc)

See https://github.com/ThaliTester/TestResults/tree/80473891c88e3bc_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
:lgtm:

---

Reviewed 5 of 5 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-:-KPxHNTDGQCNjiOt-Hbh:bnfp4nl)*

ThaliTester:
PR is added to the queue for testing as 1. task. (d6c913d)

ThaliTester:
Test (Fail) 80473891 build is completed (d6c913d)

See https://github.com/ThaliTester/TestResults/tree/80473891d6c913d_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (66c9066)

ThaliTester:
PR is added to the queue for testing as 3. task. (c2373f6)

ThaliTester:
Test (Fail) 80473891 build is completed (66c9066)

See https://github.com/ThaliTester/TestResults/tree/8047389166c9066_Implemented_thali_manager_andrew-aladev/ for the logs

ThaliTester:
Test (Fail) 80473891 build is completed (c2373f6)

See https://github.com/ThaliTester/TestResults/tree/80473891c2373f6_Implemented_thali_manager_andrew-aladev/ for the logs

yaronyg:
@andrew-aladev This wasn't supposed to be checked in until it passed build in CI and it never did. So why did you merge it into vNext_yarong_417_812?

Also you made more changes to it that I never reviewed! So the lgtm is lost when you make any changes!

In any case please review my comments and fix as appropriate.

---

Reviewed 16 of 19 files at r9, 4 of 5 files at r10, 2 of 2 files at r11.

---

*[test/www/jxcore/lib/testUtils.js, line 672 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KQ2m9IWAvRhDYTmT66f:-KQ2m9IWAvRhDYTmT66g:bm8mxf7) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c2373f6d97d4e8e101ad1de1b80dab48d9164abf/test/www/jxcore/lib/testUtils.js#L672)):*
> ```JavaScript
> };
> 
> // We doesn't want our test to run infinite time.
> ```

NB: /doesn't/don't

---

*[test/www/jxcore/lib/testUtils.js, line 680 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KQ2mH-yFa7F28lMcNd0:-KQ2mH-yFa7F28lMcNd1:b-ywmskj) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c2373f6d97d4e8e101ad1de1b80dab48d9164abf/test/www/jxcore/lib/testUtils.js#L680)):*
> ```JavaScript
>   }, timeout);
>   var oldEnd = t.end;
>   t.end = function () {
> ```

Somebody reading the code sees in a test a call to testTimeout and then later they just see a call to t.end(). How are they supposed to intuit that testTimeout redefines t.end? I think it's surprising and makes the tests harder to follow. A way to keep this design without surprising devs too much is to have the function return 't'. That way when it is called the dev sees 't = testUtils.testTimeout(t, foo)'. This gives them a strong hint that testTimeout is doing something to t.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-:-KQ2lKOPTGqmvXcX3aiO:b-yart12)*

yaronyg:

---

*[test/www/jxcore/package.json, line 25 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KQ37ldxFeJzJZyZk2f8:-KQ37ldxFeJzJZyZk2f9:b-7xpry7) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c2373f6d97d4e8e101ad1de1b80dab48d9164abf/test/www/jxcore/package.json#L25)):*
> ```JSON
>     "concat-map": "0.0.1",
>     "express": "4.13.3",
>     "express-pouchdb": "1.0.4",
> ```

So I have no idea how this all got past me but we need to use the magic version of express-pouchdb not the released version!

---

*[test/www/jxcore/package.json, line 39 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KQ37tMI8CdyOU3vMUu-:-KQ37tMI8CdyOU3vMUu0:b50azoy) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c2373f6d97d4e8e101ad1de1b80dab48d9164abf/test/www/jxcore/package.json#L39)):*
> ```JSON
>     "node-ssdp": "thaliproject/node-ssdp#fcc5d403f6cc5e667572fba059332908c3cc822e",
>     "node-uuid": "1.4.7",
>     "pouchdb-node": "5.4.4",
> ```

We changed from pouchdb-node for a reason! I don't understand why the system didn't show me these changes. :(

---

*[test/www/jxcore/package.json, line 54 \[r11\]](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840#-KQ37rppFQarfdzklZDa:-KQ37rppFQarfdzklZDb:bia5uy) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c2373f6d97d4e8e101ad1de1b80dab48d9164abf/test/www/jxcore/package.json#L54)):*
> ```JSON
>     "urlsafe-base64": "1.0.0",
>     "uuid": "2.0.2",
>     "salti": "https://github.com/thaliproject/salti.git#master",
> ```

These should have been alphabetized

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/840)*

PR #1025, opened 2016-09-13
Move off JXcore where we can

yaronyg:
build.sh - The change is just to jsdoc comments that aren't even active

README.md - Just updating to npm and removing instructions about perf
tests since they don't currently work

installCustomPouchDB.js - We can install packages using npm instead of jx

install.js - Fixing JCSC compliance issues and launching JXC with node
instead of jx. Moved cordova plugin install to npm and added in production
switch so we don't get dev dependencies

setUpDesktop.sh - Turns out everything seems to work with node so switched
there. Didn't use --production because we might want to debug some of this
stuff but I admit that is a pretty weak argument.

setUpTests.sh - Moved test server install to node as well as address
generation. Moved others to node/npm. Appears to work.

installCordovaPlugin.js - Installs Thali using node

package.json - Just updating main

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1025)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -63,8 +63,8 @@ jx npm run test-coordinated;ERROR_ABORT
 
 # Verify that docs can be generated
 #cd $PROJECT_ROOT/thali/;ERROR_ABORT
-#jx npm run createPublicDocs;ERROR_ABORT
-#jx npm run createInternalDocs;ERROR_ABORT
+#npm run createPublicDocs;ERROR_ABORT
+#npm run createInternalDocs;ERROR_ABORT
 
 # Make sure we are back in the project root folder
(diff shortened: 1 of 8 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (c426454)

ThaliTester:
Test (Fail) 84966889 build is completed (c426454)

See https://github.com/ThaliTester/TestResults/tree/84966889c426454_Move_off_JXcore_where_we_can_yaronyg/ for the logs

yaronyg:
Closing in favor of #1092

PR #1092, opened 2016-09-20
Move off JXcore where we can

larryonoff:
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1092)

PR #1100, opened 2016-09-21
Validate versions and key configurations

yaronyg:
@andrew-aladev Please do a code review
@larryonoff I suspect this has to be submitted along with #1092 since the new file depends on being run from Node not JXcore.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1100)


ThaliTester:
PR is added to the queue for testing as 7. task. (9bb6f58)

ThaliTester:
Test (Fail) 86080965 build is completed (9bb6f58)

See https://github.com/ThaliTester/TestResults/tree/860809659bb6f58_Validate_versions_and_key_configurations_yaronyg/ for the logs

andrew-aladev:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[thali/install/setUpDesktop.sh, line 4 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSDp3Pyrz6xqpxV9F30:-KSDp3Pyrz6xqpxV9F31:b-yntp4y) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/setUpDesktop.sh#L4)):*
> ```Shell
> 
> # Exit immediately if a command exits with a non-zero status.
> set -e
> ```

Maybe we can add 'set -o pipefail' for bash and forgot about 'ERROR_ABORT'?

---

*[thali/install/validateBuildEnvironment.js, line 12 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSDx_bPxfEpjixTqEPq:-KSDx_bPxfEpjixTqEPr:b-lywpng) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L12)):*
> ```JavaScript
> const versions =
> {
>   xcode: '7.3.1',
> ```

Why we are checking minor and patch parts of version? Maybe we can use some pattern here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

andrew-aladev:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[thali/install/validateBuildEnvironment.js, line 14 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGI51c-2iufBuxcFet:-KSGI51c-2iufBuxcFeu:bc8s1l8) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L14)):*
> ```JavaScript
>   xcode: '7.3.1',
>   xcodeCommandLineTools: ' ',
>   osX: '10.11.6',
> ```

It looks like 30% of this file will work only on osx. Maybe we can ignore these checks on other platform.

---

*[thali/install/validateBuildEnvironment.js, line 183 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGNJ-IimEYBNQSuQSp:-KSGNJ-JEjh4aRpnh-cP:b-82wk2c) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L183)):*
> ```JavaScript
>       (result, version) => version === result
>   },
>   sinopiaNode: {
> ```

We can forget to run sinopia.  In this case build.sh will fail with unclear error. Maybe we can make a http query to registry url and verify that title equals "Sinopia".

---

*[thali/install/validateBuildEnvironment.js, line 258 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGLVrzlCKKGmU4LoXl:-KSGLVs-JXvJwjTcFIS-:b-m4dyna) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L258)):*
> ```JavaScript
>   let errorMessage = '';
>   let runningPromise = Promise.resolve();
>   Object.getOwnPropertyNames(commandsAndResults).forEach(function (name) {
> ```

We can use array.reduce(function (runningPromise, name) {
  return runningPromise.then(...);
} , Promise.resolve());

---

*[thali/install/validateBuildEnvironment.js, line 276 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGJyhWqTzkjUwuloNc:-KSGJyhWqTzkjUwuloNd:b-jlf8zl) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L276)):*
> ```JavaScript
> // Detects if we were called from the command line
> if (require.main === module) {
>   const versionsCount = Object.getOwnPropertyNames(versions).length;
> ```

It looks like we are checking that array of `versions` keys equals array of `commandsAndResults` keys. Maybe we can replace this with assert arrayEquals.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

ThaliTester:
PR is added to the queue for testing as 1. task. (6d2a719)

ThaliTester:
Test (Fail) 86080965 build is completed (6d2a719)

See https://github.com/ThaliTester/TestResults/tree/860809656d2a719_Validate_versions_and_key_configurations_yaronyg/ for the logs

yaronyg:
Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[thali/install/setUpDesktop.sh, line 4 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSDp3Pyrz6xqpxV9F30:-KSELPkzOL68ngl9j9n6:b-hfc8th) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/setUpDesktop.sh#L4)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> Maybe we can add 'set -o pipefail' for bash and forgot about 'ERROR_ABORT'?
</details>
I think the answer is no. But the reasons are really interesting (to me at least) and I think you found a major bug in our script.

The whole point of ERROR_ABORT is to give a loud and easy to spot error if something goes wrong with the script. This is really important while we are developing on a machine and running setUpDesktop.sh manually and even when we are debugging failed builds on CI.

However it turns out that because we are using set -e, ERROR_ABORT will NEVER get called. For example, if 'cd 'dirname $0' fails then set -e will immediately exit and ERROR_ABORT won't be called!

So the current set up makes no sense.

My preference is to remove set -e. The reason is that set -e (or, apparently what the cool kids use according to http://redsymbol.net/articles/unofficial-bash-strict-mode/ is set -euo pipefail). I think instead we should just use set -uo pipefail. But even with set -uo pipefail (or just set -o pipefail) we still need ERROR_ABORT. set -o pipefail just makes a piped set of commands fail if any of them fail but it doesn't cause the script to exit.

So, for example, I have a script called test.sh that contains 'exit -1'. So I then run in another script file:

bash ./test.sh | echo;ERROR_ABORT
echo "still here"

Now if I run this without any set values I will output "still here".

If I add 'set -e' then I still output "still here" because the error is swallowed by the pipe.

If I add 'set -uo pipefail' then I get the ERROR_ABORT message because the pipe failed.

But if I add 'set -euo pipefail' then I get nothing. The script exits with a -1 but I don't get an error message because pipefail made the pipe fail and e made us immediately exit.

But wait, you forced me into bash hell (how dare you ask a good question!!!!) so now you get to go farther with me. Because I think there is a clean way to make us all happy. Which is:

set -euo pipefail
trap "LOG '${RED_COLOR}' 'SetUp Desktop Failed\n'" ERR

It turns out that if a script exits  due to -e then it fires an ERR signal that we can catch and then log.

So with the two lines above we can remove ERROR_ABORT from all of our commands!

This has been your tutorial in Bash and I'd like to say again what I've said before, I HATE BASH!

---

*[thali/install/validateBuildEnvironment.js, line 12 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSDx_bPxfEpjixTqEPq:-KSJJyCgTfCMbam13nDl:bs5wue0) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L12)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> Why we are checking minor and patch parts of version? Maybe we can use some pattern here.
</details>
This is intentional. Our goal is that any two devs building on any two machines will get as close to identical builds as we can reasonably accomplish. And unfortunately even minor version differences in tools can cause differences in the output (even if they aren't supposed to). So to keep our sanity we are locking everything down. Think of this as the equivalent of using shrinkwrap.

---

*[thali/install/validateBuildEnvironment.js, line 14 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGI51c-2iufBuxcFet:-KSJKCoLR3OpiBzAQsuE:b-886gsx) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L14)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> It looks like 30% of this file will work only on osx. Maybe we can ignore these checks on other platform.
</details>
Fixed

---

*[thali/install/validateBuildEnvironment.js, line 183 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGNJ-IimEYBNQSuQSp:-KSJTAj9FZa8HUJBS8la:brnzjo5) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L183)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> We can forget to run sinopia.  In this case build.sh will fail with unclear error. Maybe we can make a http query to registry url and verify that title equals "Sinopia".
</details>
I found a super easy fix for this (see #1150) but it's just not worth the time right now to fix.

---

*[thali/install/validateBuildEnvironment.js, line 258 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGLVrzlCKKGmU4LoXl:-KSJfZoaYNmhRiId3LIh:b-886gsx) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L258)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> We can use array.reduce(function (runningPromise, name) {
>   return runningPromise.then(...);
> } , Promise.resolve());
</details>
Fixed

---

*[thali/install/validateBuildEnvironment.js, line 276 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSGJyhWqTzkjUwuloNc:-KSJfewc66Kte6zsskFe:b-886gsx) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L276)):*
<details><summary><i>Previously, andrew-aladev (Andrew Aladjev) wrote…</i></summary>
> It looks like we are checking that array of `versions` keys equals array of `commandsAndResults` keys. Maybe we can replace this with assert arrayEquals.
</details>
Fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

mlesnic:
Versions installed in CI that are different than in script:
brew 1.0.4.9
jxcore 0.3.1.2
androidSupportLibraries - script says it is not installed, when in fact it is, v38
gradle is installed during every build, so commands according to gradle versions failed, 
git 2.7.4
switflit 0.12.0

JXCore will be updated when #1092 will be checked into vNext and according to the rest of tools, should I downgrade them? I have doubts, because all of this tools are newer than in script. And what about gradle?

ThaliTester:
PR is added to the queue for testing as 2. task. (06a2b2e)

yaronyg:
@mlesnic Thanks for checking this. Please see the points below. Some things I fixed on my end but others require changes to CI. Please work with @larryonoff to get us to the point where everything is passing but the JXcore version. @larryonoff can make changes to this PR if necessary to fix things, just so we don't end up having to spend another day on this.
* Please update brew to 1.0.4-17. If you run 'brew update' that should update CI to the same version.
* JXcore - that is the bug we have to fix using @larryonoff's #1092 
* androidSupportLibraries - I'm an idiot, I'll go fix it. Sorry. :(
* gradle - Are you talking about cordovaAndroidSet*? If so then those have to be manually set in ~/.bashrc and ~/.bash_profile. And they have to be set before we run build.sh. So this really means we need to set them, take a snapshot of the VM and then see if the script passes on the new snapshot. The values that are needed in both files are:

export ORG_GRADLE_PROJECT_cdvMinSdkVersion=22
export ORG_GRADLE_PROJECT_cdvBuildToolsVersion=23.0.3
export ORG_GRADLE_PROJECT_cdvCompileSdkVersion=android-23
* git - Please update the image to 2.10.0 and then take a new snapshot. 2.10.0 has fixes we need to work with worktree which is ridiculously useful.
* swiftlint - Fixed

---

Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-:-KSbqvWGgcftAmJI66mC:b41c7yg)*

ThaliTester:
Test (Fail) 86080965 build is completed (06a2b2e)

See https://github.com/ThaliTester/TestResults/tree/8608096506a2b2e_Validate_versions_and_key_configurations_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (5fb062a)

ThaliTester:
Test (Fail) 86080965 build is completed (5fb062a)

See https://github.com/ThaliTester/TestResults/tree/860809655fb062a_Validate_versions_and_key_configurations_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (759fc99)

ThaliTester:
Test (Fail) 86080965 build is completed (759fc99)

See https://github.com/ThaliTester/TestResults/tree/86080965759fc99_Validate_versions_and_key_configurations_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (1df36b7)

ThaliTester:
Test (Fail) 86080965 build is completed (1df36b7)

See https://github.com/ThaliTester/TestResults/tree/860809651df36b7_Validate_versions_and_key_configurations_yaronyg/ for the logs

yaronyg:
@mlesnic / @larryonoff - I looked at the latest build failure and if [this](https://github.com/direnv/direnv/issues/210) is correct then it's a problem with the version of RVM in the CI environment.

larryonoff:
@yaronyg [The issue](https://github.com/direnv/direnv/issues/210) refers on ZSH, but our CI uses bash. Also macOS doesn't have `rvm` installed by default.

larryonoff:
Review status: 0 of 7 files reviewed at latest revision, 10 unresolved discussions.

---

*[thali/install/validateBuildEnvironment.js, line 14 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCQpW3E4ZCvuetFiI9:-KSCQpW3E4ZCvuetFiIA:b-opvzhz) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L14)):*
> ```JavaScript
>   xcode: '7.3.1',
>   xcodeCommandLineTools: ' ',
>   osX: '10.11.6',
> ```

NB: rename into macOS. This really new name :)

---

*[thali/install/validateBuildEnvironment.js, line 22 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCQvdeAYD3T1gruvxZ:-KSCQvdeAYD3T1gruvx_:b-747ugk) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L22)):*
> ```JavaScript
>   jxcore: '0.3.1.4',
>   androidBuildTools: '23.0.3',
>   androidPlatform: 'android-23',
> ```

I'm not sure about this one is correct. In install script I make the following substitutions:
- android-${versions.androidPlatform}
- build-tools-${versions.androidBuildTools}

---

*[thali/install/validateBuildEnvironment.js, line 38 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCRN7ql4wHfMwJX6NV:-KSCRN7ql4wHfMwJX6NW:b7ngca1) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L38)):*
> ```JavaScript
>   java: '1.8.0_102',
>   git: '2.7.4',
>   swiftLint: '0.1.1',
> ```

1. please rename into `swiftlint` (just lowercase all letters)
2. please change the version into `0.12.0`. `0.1.1` is too old one.

---

*[thali/install/validateBuildEnvironment.js, line 21 at r6](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSeiFeUSO3IMIMMuwY4:-KSeiFeUSO3IMIMMuwY5:bvsfqz6) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df36b74ee93f1ece85579857ed90a0e05a0cdd1/thali/install/validateBuildEnvironment.js#L21)):*
> ```JavaScript
>   node: '6.6.0',
>   npm: '3.10.3',
>   brew: '1.0.4-17',
> ```

I thought a little about fixing brew version. As result I don't like the idea of fixing brew,  because it updates very often (from 1.0 into 1.0.4 in about a week). So the latest version of brew will be installed if someone installs brew the first time, as result script will fail. I suggest using it as minimum version.

My command line tells me that brew version is 1.0.4, not 1.0.4-17.
```
brew --version
Homebrew 1.0.4
Homebrew/homebrew-core (git revision cc9ac; last commit 2016-09-26)
```

And it's better having semantic version check. It can be easily added using https://github.com/npm/node-semver.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

mlesnic:
@yaronyg 
Build fails because validate environment script fails.
I upgraded git to 2.10.0, added exports to bashrc and bash_profile, brew to 1.0.5 - newer than in script. So right now the script fails only on brew and jxcore versions.

larryonoff:
@yaronyg I think we shouldn't fix minor versions, but fix major versions.

ThaliTester:
PR is added to the queue for testing as 1. task. (b8cdf67)

ThaliTester:
Test (Fail) 86080965 build is completed (b8cdf67)

See https://github.com/ThaliTester/TestResults/tree/86080965b8cdf67_Validate_versions_and_key_configurations_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (c953861)

ThaliTester:
Test (Fail) 86080965 build is completed (c953861)

See https://github.com/ThaliTester/TestResults/tree/86080965c953861_Validate_versions_and_key_configurations_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (193feb9)

yaronyg:
@larryonoff @andrew-aladev I don't want this PR to rot so if you two can review I'd appreciate it.

ThaliTester:
Test (Fail) 86080965 build is completed (193feb9)

See https://github.com/ThaliTester/TestResults/tree/86080965193feb9_Validate_versions_and_key_configurations_yaronyg/ for the logs

larryonoff:
Review status: 0 of 6 files reviewed at latest revision, 11 unresolved discussions.

---

*[thali/install/validateBuildEnvironment.js, line 337 at r9](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KUS5ck5co5jhTqP__NE:-KUS5ck5co5jhTqP__NF:bvt5gmg) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/193feb9e49fa46e29e0826523895c287ff1e9d32/thali/install/validateBuildEnvironment.js#L337)):*
> ```JavaScript
>       const jx = path.join(home, '.jx');
>       const jxc = path.join(home, '.jxc');
>       const nodeGyp = path.join(home, 'node-gyp');
> ```

`node-gyp` folder name should start from symbol `.`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

ThaliTester:
PR is added to the queue for testing as 1. task. (27264c9)

yaronyg:
@larryonoff @andrew-aladev I think I successfully resolved all of your comments. If there are any unresolved issues please let me know. Otherwise please give me a LGTM.

---

Review status: 0 of 6 files reviewed at latest revision, 11 unresolved discussions.

---

*[thali/install/validateBuildEnvironment.js, line 14 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCQpW3E4ZCvuetFiI9:-KUK11ksn0tJl8AliVke:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L14)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> NB: rename into macOS. This really new name :)
</details>
Done.

---

*[thali/install/validateBuildEnvironment.js, line 22 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCQvdeAYD3T1gruvxZ:-KUK480oMWqtaPmHk-Jo:broncoh) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L22)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> I'm not sure about this one is correct. In install script I make the following substitutions:
> - android-${versions.androidPlatform}
> - build-tools-${versions.androidBuildTools}
</details>
I don't understand. The validation logic checks the environment variables which is what we are trying to set.

---

*[thali/install/validateBuildEnvironment.js, line 38 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCRN7ql4wHfMwJX6NV:-KUK4K_xBfMvHDRt9tkb:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L38)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> 1. please rename into `swiftlint` (just lowercase all letters)
> 2. please change the version into `0.12.0`. `0.1.1` is too old one.
</details>
Done.

---

*[thali/install/validateBuildEnvironment.js, line 21 at r6](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSeiFeUSO3IMIMMuwY4:-KUK1N7ABNdkOYqqpy86:b-q3yjp0) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/1df36b74ee93f1ece85579857ed90a0e05a0cdd1/thali/install/validateBuildEnvironment.js#L21)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> I thought a little about fixing brew version. As result I don't like the idea of fixing brew,  because it updates very often (from 1.0 into 1.0.4 in about a week). So the latest version of brew will be installed if someone installs brew the first time, as result script will fail. I suggest using it as minimum version.
> 
> My command line tells me that brew version is 1.0.4, not 1.0.4-17.
> ```
> brew --version
> Homebrew 1.0.4
> Homebrew/homebrew-core (git revision cc9ac; last commit 2016-09-26)
> ```
> 
> And it's better having semantic version check. It can be easily added using https://github.com/npm/node-semver.
</details>
Semver won't work because the actual version number returned includes extra data and besides who says that Brew actually follows semver? In any case it's so easy to accidentally update brew that I changed the logic to just see that the version starts with 1.0.

---

*[thali/install/validateBuildEnvironment.js, line 337 at r9](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KUS5ck5co5jhTqP__NE:-KUTfYVM-RO2o6JPiiEu:b2e5any) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/193feb9e49fa46e29e0826523895c287ff1e9d32/thali/install/validateBuildEnvironment.js#L337)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> `node-gyp` folder name should start from symbol `.`.
</details>
Eek! Good catch!. Fixed

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-:-KUThVmG-jrZZxRaUhao:bfgricc)*

ThaliTester:
Test (Fail) 86080965 build is completed (27264c9)

See https://github.com/ThaliTester/TestResults/tree/8608096527264c9_Validate_versions_and_key_configurations_yaronyg/ for the logs

larryonoff:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[thali/install/validateBuildEnvironment.js, line 22 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCQvdeAYD3T1gruvxZ:-KUW8qhJDB5ED8JWE7Oh:b-pakgv4) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L22)):*
> I don't understand. The validation logic checks the environment variables which is what we are trying to set.
</details>
I meant that I'm not sure that `androidPlatform`  value `android-23` is valid, just `23` may be better. Because e.g. for `androidBuildTools` there can be the same semantics 'build-tools-23.0.3', but we use just version. So I think that we should use the same approach for both variable and I think this should be versions. 

Install script uses this variables, for example for all android tools it's

```
echo "y" | android update sdk --no-ui --all --filter \
tools,\
platform-tools,\
extra-android-m2repository,\
extra-google-m2repository,\
build-tools-${versions.androidBuildTools},\
android-${versions.androidPlatform}
```
As result after substitution it will be 
```
echo "y" | android update sdk --no-ui --all --filter \
tools,\
platform-tools,\
extra-android-m2repository,\
extra-google-m2repository,\
build-tools-23.0.3
android-android-23
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

andrew-aladev:
:lgtm:

---

Reviewed 5 of 9 files at r8, 1 of 1 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-:-KUYeBGP7pOYMXznAuz6:bnfp4nl)*

yaronyg:

---

*[thali/install/validateBuildEnvironment.js, line 22 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100#-KSCQvdeAYD3T1gruvxZ:-KUsSuOBQLy71fkg7eb1:b-3ydbvc) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/9bb6f589fe1b3c0cf65378cd023c9e746bb6127d/thali/install/validateBuildEnvironment.js#L22)):*
<details><summary><i>Previously, larryonoff (larryonoff) wrote…</i></summary>
> I meant that I'm not sure that `androidPlatform`  value `android-23` is valid, just `23` may be better. Because e.g. for `androidBuildTools` there can be the same semantics 'build-tools-23.0.3', but we use just version. So I think that we should use the same approach for both variable and I think this should be versions. 
> 
> Install script uses this variables, for example for all android tools it's
> 
> ```
> echo "y" | android update sdk --no-ui --all --filter \
> tools,\
> platform-tools,\
> extra-android-m2repository,\
> extra-google-m2repository,\
> build-tools-${versions.androidBuildTools},\
> android-${versions.androidPlatform}
> ```
> As result after substitution it will be 
> ```
> echo "y" | android update sdk --no-ui --all --filter \
> tools,\
> platform-tools,\
> extra-android-m2repository,\
> extra-google-m2repository,\
> build-tools-23.0.3
> android-android-23
> ```
> 
</details>
I completely agree with this comment but right now we need to get the code checked in and the release out so I'm filing this as #1411 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1100)*

ThaliTester:
PR is added to the queue for testing as 1. task. (d038c64)

ThaliTester:
Test (Fail) 86080965 build is completed (d038c64)

See https://github.com/ThaliTester/TestResults/tree/86080965d038c64_Validate_versions_and_key_configurations_yaronyg/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (7e6d887)

ThaliTester:
Test (Fail) 86080965 build is completed (7e6d887)

See https://github.com/ThaliTester/TestResults/tree/860809657e6d887_Validate_versions_and_key_configurations_yaronyg/ for the logs

PR #1148, opened 2016-09-24
SSDP updater for runTests and runCoordinated

### LINKED ISSUE #1072: Please change tests to use random SSDP_NT when we are running on desktop
Please see thaliproject/Thali_CordovaPlugin#1066 for context

### Requirements:

When we run 'jx npm test' or 'jx npm run test-meta' we have to make sure that we use a randomly generated SSDP_NT value that will override the value from thaliConfig.js. We want to make sure however that the randomly generated value isn't put into thaliConfig.js since the test is run in our repo clone and we could accidentally check it in!

When we run 'jx npm run test-coordinated' on desktop (meaning inside of the repo clone) we need the node processes spawned by runCoordinatedTests.js to all have the same randomly generated SSDP_NT value or they won't find each other. To make things more complex if we are running both iOS and Android tests (e.g. when @artemjackson fixes thaliproject/Thali_CordovaPlugin#902) simultaneously then all the node processes using the iOS mock layer have to have the same randomly generated SSDP_NT value and all the node processes using the Android mock layer have to also all have the same SSDP_NT value but a different one than the iOS processes so the tests don't interact.

### Ways to meet the requirements:

For 'jx npm test' and 'jx npm run test-meta' I suspect the original instructions in this bug probably made the most sense, that is, edit runTests.js to:
* Check to see if we are running under the Thali_CordovaPlugin directory, if so then require thaliConfig.js, replace the value of SSDP_NT in memory with a new randomly generated string (use npm randomstring) of the same size as the current string.
* If we aren't under Thali_CordovaPlugin then loadFile should just require(filePath) and be done. But if we are under Thali_CordovaPlugin then use proxyquire with @global in the form proxyquire(filePath, { 'thaliConfig': { thaliConfig, '@global': true }}) so that we will replace thaliConfig in all files all the way down the line with our updated one.

But this won't work for 'jx npm run test-coordinated' because it would result in all the spawned Node processes having different SSDP_NT values and so not finding each other. I can see two ways to fix this:

* Pass an argument to UnitTest_app.js - We could have runCoordinatedTests, when it calls spawn in spawnTestInstance pass in an argument with the SSDP_NT value that the spawned instance is to use. This would be passed to UnitTest_app.js who would then have to pass it to runTests.js who would then use it. We could then have @artemjackson fix thaliproject/Thali_CordovaPlugin#902 by teaching runCoordinatedTests how to pass one SSDP_NT value to Android mock images and a different one to the iOS mock images.
* Introduce a thaliConfigOveride.json file - This would be a file that we would put into .gitignore. The file would have to have two sections in its json value, iOS and Android. We would then put in the SSDP_NT values for iOS and a different for Android in the appropriate sections and we would then test runTests.js to read in the file and use it to override thaliConfig.js based on what platform it is. If we go this route then we should obviously make 'jx npm test' and 'jx npm run test-meta' work the same way. Which is the main advantage of this approach. We have one file that won't ever be checked in that handles everything.

### Which do you prefer?

Personally I like passing the argument. It just seems simpler and it doesn't create any artifacts we could ever check in. But that's just my two cents. Pick which ever one you want and please implement! Or do you have a third option?

andrew-aladev:
Closes #1072

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1148)

### DIFF
--- a/test/TestServer/HttpServer.js
+++ b/test/TestServer/HttpServer.js
@@ -72,7 +72,7 @@ Server.prototype._disconnect = function (socket, reason) {
 Server.prototype._error = function (socket, error) {
   logger.error(
     'unexpected server error: \'%s\'',
-    error
+    error.toString()
   );
   throw new Error(error);
 }
--- a/test/www/jxcore/runCoordinatedTests.js
+++ b/test/www/jxcore/runCoordinatedTests.js
@@ -1,6 +1,9 @@
 'use strict';
 
-var spawn = require('child_process').spawn;
+var spawn        = require('child_process').spawn;
+var randomString = require('randomstring');
+var objectAssign = require('object-assign');
+
 
 var DEFAULT_INSTANCE_COUNT = 3;
 
@@ -114,13 +117,20 @@ var testServerInstance = spawn('jx', ['../../TestServer/index.js',
   JSON.stringify(testServerConfiguration)]);
 setListeners(testServerInstance, 0);
 
+
+// We want to provide same random SSDP_NT for each test instance in group.
+var localEnv = objectAssign({}, process.env);
+localEnv.SSDP_NT = randomString.generate({
+  length: 'http://www.thaliproject.org/ssdp'.length
+});
+
 var testInstances = {};
 var spawnTestInstance = function (instanceId) {
   var instanceArgs = [argv.test];
   if (argv.filter) {
     instanceArgs.push(argv.filter);
   }
-  var testInstance = spawn('jx', instanceArgs);
+  var testInstance = spawn('jx', instanceArgs, { env: localEnv });
   setListeners(testInstance, instanceId);
   testInstances[instanceId] = testInstance;
 };
--- a/test/www/jxcore/runTests.js
+++ b/test/www/jxcore/runTests.js
@@ -1,10 +1,23 @@
 'use strict';
 
-var fs = require('fs-extra-promise');
-var path = require('path');
+var fs           = require('fs-extra-promise');
+var path         = require('path');
+var randomString = require('randomstring');
+
+
+// Before including anything serious from thali we want to ensure
+// that we have SSDP_NT env defined.
+if (!process.env.SSDP_NT) {
+  // We want to provide a new random value.
+  process.env.SSDP_NT = randomString.generate({
+    length: 'http://www.thaliproject.org/ssdp'.length
+  });
+}
+
 var thaliTape = require('./lib/thaliTape');
 var testUtils = require('./lib/testUtils');
 var logger    = require('./lib/testLogger')('runTests');
+
 var platform = require('thali/NextGeneration/utils/platform');
 
 // The global.Mobile object is replaced here after thaliTape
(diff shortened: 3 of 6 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (a7376dd)

yaronyg:
Please resolve the one comment and then :lgtm: 

---

Reviewed 5 of 5 files at r1.

---

*[test/www/jxcore/runTests.js, line 10 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1148#-KSOY_JSSSdZUtmiJvfZ:-KSOYSW0JowVtnAIOR6S:b84vt8a) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/a7376dd99dac8c849d84962f6afb0e8f92d51332/test/www/jxcore/runTests.js#L10)):*
> ```JavaScript
> // Before including anything serious from thali we want to ensure
> // that we have SSDP_NT env defined.
> if (!process.env.SSDP_NT) {
> ```

If we don't have the environment variable defined then shouldn't we just use whatever value is in ThaliConfig?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1148#-:-KSOYtd9Vsk6A3T25arV:b-lwk48n)*

ThaliTester:
Test (Success) 86599259 build is completed (a7376dd)

See https://github.com/ThaliTester/TestResults/tree/86599259a7376dd_SSDP_updater_for_runTests_and_runCoordinated_andrew-aladev/ for the logs

ThaliTester:
Test 86599259a7376dd(a7376dd) has failed

See https://github.com/ThaliTester/TestResults/tree/86599259a7376dd_SSDP_updater_for_runTests_and_runCoordinated_andrew-aladev/ for the fail logs

ThaliTester:
PR is added to the queue for testing as 2. task. (2488df7)

andrew-aladev:
Review status: 4 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[test/www/jxcore/runTests.js, line 10 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1148#-KSOY_JSSSdZUtmiJvfZ:-KSb9nvyo9W5FnwVBKXn:b-k7vkzi) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/a7376dd99dac8c849d84962f6afb0e8f92d51332/test/www/jxcore/runTests.js#L10)):*
> If we don't have the environment variable defined then shouldn't we just use whatever value is in ThaliConfig?
</details>
When we are doing 'jx npm test' or 'jx npm run test-meta'  we want to have a random value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1148)*

ThaliTester:
Test (Success) 86599259 build is completed (2488df7)

See https://github.com/ThaliTester/TestResults/tree/865992592488df7_SSDP_updater_for_runTests_and_runCoordinated_andrew-aladev/ for the logs

ThaliTester:
Test 865992592488df7(2488df7) has failed

See https://github.com/ThaliTester/TestResults/tree/865992592488df7_SSDP_updater_for_runTests_and_runCoordinated_andrew-aladev/ for the fail logs

PR #1279, opened 2016-10-07
Spawning test in separate process

### LINKED ISSUE #1228: Please change the test runner to run each test in its own child process of JXcore
The way we run tests right now we load all the tests at the start which is a huge number of files and is stressing JXcore way beyond what it would have to deal with during run time. So we need to change how we run tests so that we run each test in its own JXcore instance. So we would probably take runTests and instead of having it load each test file into the environment, instead we would have it spawn a JXcore child process and have it run the test, exit, clean up and then create a new child process for the next test. The obvious question is how do we make the coordination server work properly? It depends on getting a test count which comes from reading all of the files but that should be easy to fix. .Only will also be tricky but let that be a lower priority feature.

andrew-aladev:
We have both `require` (old variant) and `spawn` working: `require` for ios (for now) and `spawn` for desktop and android.
Closes #1228.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1279)

### DIFF
--- a/test/TestServer/HttpServer.js
+++ b/test/TestServer/HttpServer.js
@@ -90,6 +90,13 @@ Server.prototype._present = function (socket, deviceInfo) {
   this.emit('present', device);
 }
 
+Server.prototype.disconnectAll = function () {
+  var sockets = this._io.sockets.sockets;
+  Object.keys(sockets).forEach(function(socketName) {
+    sockets[socketName].disconnect(true);
+  });
+}
+
 Server.prototype._exit = function () {
   this._io.close();
 }
--- a/test/TestServer/TestFramework.js
+++ b/test/TestServer/TestFramework.js
@@ -21,35 +21,7 @@ function TestFramework(options) {
   TestFramework.super_.call(this);
 
   this._setOptions(options);
-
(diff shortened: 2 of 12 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (368e99e)

ThaliTester:
Test (Success) 88523129 build is completed (368e99e)

See https://github.com/ThaliTester/TestResults/tree/88523129368e99e_Spawning_test_in_separate_process_andrew-aladev/ for the logs

ThaliTester:
Test 88523129368e99e(368e99e) has failed

See https://github.com/ThaliTester/TestResults/tree/88523129368e99e_Spawning_test_in_separate_process_andrew-aladev/ for the fail logs

yaronyg:
:lgtm:

---

Reviewed 12 of 12 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1279#-:-KTgo4Q8ZN2INM3OAbXW:bnfp4nl)*

PR #1430, opened 2016-10-26
Enable coordination server to run multiple networkTypes

### LINKED ISSUE #1172: Enable coordination server to run multiple networkTypes
This is a relative of thaliproject/Thali_CordovaPlugin#915. 

The goal is to be able to specify separately for iOS and for Android which networkTypes to run. In other words we might run WiFi only mode in iOS but Wifi only and Native in Android.

I think the simplest way we can do this is to just edit UnitTest_app.js. We can specify what network types to run for iOS vs Android. Then when UnitTest_app.js is running it can detect if it's Android or iOS and then just run the tests N times once for each entry in the networkTypes array for that platform.

To the coordination server it will look like it is just running Android and iOS once. The coordination server won't know or care that in fact the tests are being run 3 times.

If I remember correctly at some point the code reports back to the coordination server how many tests are going to be run so if we are going to run the tests multiple times because of multiple network types then we need to report a multiple of a single run to make sure the numbers match.

chapko:
Closes #1172.

I have to admit it is very ugly solution. Partially because we have multiple entry points for running tests (there is `jx runCoordinated.js`, `jx runTests.js`, and `jx UnitTest_app.js` which are depend on each other) and side effects in the core modules.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1430)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -182,6 +182,26 @@ starting the coordinated test from the command line but passing in the parameter
 it runs with the assumption that one will then run `jx UnitTest_app.js` from
 inside an IDE to debug.
 
+By default tests are running with mocked android native API and with WiFi
+network type. It is possible to change it via `--networkType=<wifi|both|native>`
+and `--platform=<ios|android>` parameters, for example:
+
+```
+$ jx runCoordinatedTests.js --platform=ios --networkType=both
+```
+
+Also it is possible to run multiple coordinated tests simultaneously on the same
+machine. Use `COORDINATED_PORT` environment variable to set different port for
+each test runner, for example:
+
+```
+COORDINATED_PORT=12345 jx runCoordinatedTests.js --platform=ios &> ios.log & \
+COORDINATED_PORT=54321 jx runCoordinatedTests.js --platform=android &> android.log &
+```
+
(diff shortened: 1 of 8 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (87a8d91)

ThaliTester:
Test (Success) 91033152 build is completed (87a8d91)

See https://github.com/ThaliTester/TestResults/tree/9103315287a8d91_Enable_coordination_server_to_run_multiple_networkTypes_chapko/ for the logs

ThaliTester:
Test 9103315287a8d91(87a8d91) has failed

See https://github.com/ThaliTester/TestResults/tree/9103315287a8d91_Enable_coordination_server_to_run_multiple_networkTypes_chapko/ for the fail logs

yaronyg:
:lgtm:

---

Reviewed 8 of 8 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1430#-:-KV1cgK5H0aUFRY3L7LE:bnfp4nl)*

PR #1539, opened 2016-11-18
1382 thaliNotificationClient needs different logic in iOS branch

### LINKED ISSUE #1382: thaliNotificationClient needs different logic in iOS branch
**TL;DR - I'm guessing you should do the sleezy choice below and then once the sleezy choice is committed, update this bug to state that and put the bug into icebox changing the label from 'bug' to 'enhancement' so we eventually implement the better choice.**

The original design for thaliNotificationClient was based on the assumption that every time a device updated its generation it would generate a brand new peerID and that there was no way for a peer to know "Hey, this is the same peer I saw before, but with a new peerID!"

Obviously that isn't the case with the new design where we explicitly have a semi-persistent peerID (e.g. we only change it when we call stop) and a changing generation.

So this argues for changing how thaliNotificationClient spawns thaliNotificationActions and what those thaliNotificationActions do.

Now under the theory of 'do less' we shouldn't care but we are going to run into an interesting issue. When we index entries in the peerDictionary using peerID this effectively means that the peerDictionary now has a single global state for a peer no matter what. But this is going to get really confusing if we get multiple peerAvailabilityChanged events for the same peerID (but obviously different generations).

All the existing code is going to break. Because we will keep overwriting the status of the thaliNotificationActions in the same entry in peerDictionary! This is because peerDictionary was intended originally to track what we would now think of as a combination of peerID and generation.

So we have two choices:

Sleezy but probably best choice - Change the key for the peerDictionary so it combines peerID and generation. Note however that peerAvailabilityChanged events will only be per peerID and not per generation so we will need new code that searches the dictionary for all entries related to the same peerID and removes them.

Better but probably have to wait until later choice - Change thaliNotificationClient so that it actually manages a single peer (not generation) in peerDictionary. This has a variety of interesting but scary implications. For example, let's say we get a notification of 0:0 (peer ID 0, generation 0) and we fire off a thaliNotificationAction. Let's say that the thaliNotificationAction is still in created state when we get a second peerAvailabilityChanged notification of 0:1. In theory we don't have to do anything. After all when the thaliNotificationAction fires it will connect to the peer using the peerID, not the generation. So the fact that the generation changed really doesn't matter. But what if, for example, when we get the 0:1 notification the thaliNotificationAction is in the started state? In theory we could ignore the 0:1 notification since we already have an active action trying to retrieve the beacons but in fact this would lead to a race condition:

time 0 - the action retrieves the beacons
time 1 - the beacons on the remote peer are changed
time 2 - the remote peer advertises a new generation
time 3 - we receive the peerAvailabilityChanged event for 0:1
time 4 - we check and see that the outstanding thaliNotifcationAction for the peer is 'started' and so ignore the 0:1 notification
time 5 - The thaliNotificationAction for the peer returns saying 'nothing has changed'

In this case we missed the beacon update all together! So it means our logic has to be that if we get a peerAvailabilityChanged notification while an existing thaliNotificationAction is running then we have to actually queue a second thaliNotificationAction for the same peer. Although we can be reasonably smart and not queue the second one until the first one has returned. That way at any one time we have exactly one thaliNotificationAction for any peer outstanding.

Things can also get a little tricky when peerAvailable = false events. In theory if we get a peerAvailabilityChanged event with peerAvailable = false then we should just kill any actions even if they are started. But especially on Android there are a ton of good reasons why the native layer thinks a peer is gone when they aren't. But we should probably handle that down in thaliMobile and not here. Here we should just a peerAvailable = false as 'authoritative' and kill any action and kill any plans to enqueue a second action (per the race condition above).

chapko:
Fixes #1382.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1539)


ThaliTester:
PR is added to the queue for testing as 1. task. (3e9c5d2)

ThaliTester:
Test 94407748 (3e9c5d2) build started.

ThaliTester:
Test (Success) 94407748 build is completed (3e9c5d2)

See https://github.com/ThaliTester/TestResults/tree/944077483e9c5d2_1382_thaliNotificationClient_needs_different_logic_in_iOS_branch_chapko/ for the logs

ThaliTester:
Test 944077483e9c5d2(3e9c5d2) has failed

See https://github.com/ThaliTester/TestResults/tree/944077483e9c5d2_1382_thaliNotificationClient_needs_different_logic_in_iOS_branch_chapko/ for the fail logs

yaronyg:
I think we can get rid of thaliPeerConnectionInformation.js but otherwise :lgtm: 

---

Reviewed 11 of 11 files at r1.

---

*[thali/NextGeneration/notification/thaliPeerConnectionInformation.js, line 9 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539#-KWu78S6ccQXUSZlp87z:-KWu78S6ccQXUSZlp88-:b162z5h) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/3e9c5d25aeecd8eec1a0ad80335456ad530c7de4/thali/NextGeneration/notification/thaliPeerConnectionInformation.js#L9)):*
> ```JavaScript
>  *
>  * Defines a PeerConnectionInformation class for use by
>  * {@link module:thaliPeerDictionary}.
> ```

Looking at the code we never actually use this in thaliPeerDictionary (it's references in NotificationPeerDictionaryEntry but in fact it isn't passed in not is it needed to be passed in).

In fact as far as I can tell the only place this is ever used is in thaliNotificationAction and the only purpose it is ever used for is to get the host address and port number.

Wouldn't it be easier to just delete this entire file and change thaliNotificationAction on line 133 to not create _peerConnection but instead create self._peerHostInfo and call it a day?

---

*[thali/NextGeneration/notification/thaliPeerDictionary.js, line 169 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539#-KWu8e03r59m2Lc6ZGz1:-KWu8e03r59m2Lc6ZGz2:b-r69z8h) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/3e9c5d25aeecd8eec1a0ad80335456ad530c7de4/thali/NextGeneration/notification/thaliPeerDictionary.js#L169)):*
> ```JavaScript
>  *
>  * @public
>  * @param {object} peer
> ```

NB: Please change to 'Object' (linting)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539#-:-KWuB32SXhgOHWEUOzq2:b-xnnj2u)*

ThaliTester:
PR is added to the queue for testing as 1. task. (f58d03e)

ThaliTester:
Test 94407748 (f58d03e) build started.

ThaliTester:
Test (Success) 94407748 build is completed (f58d03e)

See https://github.com/ThaliTester/TestResults/tree/94407748f58d03e_1382_thaliNotificationClient_needs_different_logic_in_iOS_branch_chapko/ for the logs

chapko:
Review status: 5 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[thali/NextGeneration/notification/thaliPeerDictionary.js, line 169 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539#-KWu8e03r59m2Lc6ZGz1:-KX6IBsfUG2tXhUSmauo:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/3e9c5d25aeecd8eec1a0ad80335456ad530c7de4/thali/NextGeneration/notification/thaliPeerDictionary.js#L169)):*
> NB: Please change to 'Object' (linting)
</details>
Done.

---

*[thali/NextGeneration/notification/thaliPeerConnectionInformation.js, line 9 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539#-KWu78S6ccQXUSZlp87z:-KX6Ia5DVj5ST9f32Lnq:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/3e9c5d25aeecd8eec1a0ad80335456ad530c7de4/thali/NextGeneration/notification/thaliPeerConnectionInformation.js#L9)):*
> Looking at the code we never actually use this in thaliPeerDictionary (it's references in NotificationPeerDictionaryEntry but in fact it isn't passed in not is it needed to be passed in).
> 
> In fact as far as I can tell the only place this is ever used is in thaliNotificationAction and the only purpose it is ever used for is to get the host address and port number.
> 
> Wouldn't it be easier to just delete this entire file and change thaliNotificationAction on line 133 to not create _peerConnection but instead create self._peerHostInfo and call it a day?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539)*

ThaliTester:
Test 94407748f58d03e(f58d03e) has failed

See https://github.com/ThaliTester/TestResults/tree/94407748f58d03e_1382_thaliNotificationClient_needs_different_logic_in_iOS_branch_chapko/ for the fail logs

yaronyg:
:lgtm:

---

Reviewed 7 of 8 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1539#-:-KX7K8jlnSp-X7Z2zZJN:bnfp4nl)*

Issue #1632, opened 2016-12-13
Android build fails "uses-sdk:minSdkVersion 16 cannot be smaller than version 21" error


chapko:
<details>
<summary>Environment (click to expand)</summary>

- Android SDK Tools: 25.2.3
- Android SDK Platform-tools: 25.0.1
- Android SDK Build-tool: 25.0.1
- Installed Platforms:
  - android-21 (rev 2)
  - android-23 (rev 3)
  - android-24 (rev 2)
  - android-25 (rev 2)

```
$ env
LANG=en_US.utf8
DISPLAY=:0
SHLVL=1
LOGNAME=eugene
XDG_VTNR=2
PWD=/home/tmp/ThaliTest
HG=/usr/bin/hg
MOZ_PLUGIN_PATH=/usr/lib/mozilla/plugins
XAUTHORITY=/run/user/1000/gdm/Xauthority
DESKTOP_AUTOSTART_ID=10d97ecd65c3f016dc148156572896421200000015690001
QT_QPA_PLATFORMTHEME=qt5ct
JOURNAL_STREAM=8:18832
COLORTERM=truecolor
XDG_SESSION_ID=c2
DESKTOP_SESSION=cinnamon
XDG_SESSION_DESKTOP=cinnamon
GDMSESSION=cinnamon
GNOME_DESKTOP_SESSION_ID=this-is-deprecated
USERNAME=eugene
WINDOWPATH=2
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
VTE_VERSION=4601
MAIL=/var/spool/mail/eugene
GTK_OVERLAY_SCROLLING=1
GDM_LANG=en_US.utf8
SHELL=/usr/bin/zsh
XDG_SESSION_TYPE=x11
WINDOWID=44040279
TERM=xterm-256color
GTK_MODULES=canberra-gtk-module
SESSION_MANAGER=local/chapko:@/tmp/.ICE-unix/1569,unix/chapko:/tmp/.ICE-unix/1569
XDG_CURRENT_DESKTOP=X-Cinnamon
[...]

PR #1636, opened 2016-12-13
#1632 - Fixing broken build

yaronyg:
README - Updating to the actual validateBuildEnvironment script that we now run

install - Turns out I used the syntax for build-extras.gradle instead of gradle.properties and didn't realize it because I still had the environment variables defined.

setUpDesktop - The logic we were using which would use the global variable if already defined otherwise it would set it doesn't work in the case of NVM_NODEJS_ORG_MIRROR if someone is using NVM which also sets this variable. Since NVM is more common than needing to change this location when doing our build I decided to revert that change and just overwrite the variable when we are doing a build.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1636)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -51,10 +51,10 @@ In terms of operating system we assume you are running on macOS. In theory we ca
 run successfully on Linux or Windows but we really don't test there very often
 so things are likely to break.
 
-From there you need to install in macOS a fairly large mountain of software. For
-now you can see the list [here](https://github.com/thaliproject/Thali_CordovaPlugin/blob/1df36b74ee93f1ece85579857ed90a0e05a0cdd1/thali/install/validateBuildEnvironment.js#L16).
-You need to install the listed software at the listed version. Two of the entries
-are about Sinopia. Details on that is in the next section.
+From there you need to install in macOS a fairly large mountain of software. We
+ have a script, validateBuildEnvironment.js, that checks for all the required
+ software. You need to install all that software at the listed version for the
+ build to work. Two of the entries are about Sinopia. Details on that is in the next section.
 
 ### Running your own NPM registry
 From time to time we run into bugs in PouchDB and Express-PouchDB that we can't
--- a/thali/install/install.js
+++ b/thali/install/install.js
@@ -306,9 +306,9 @@ function generateGradlePropertiesFile(thaliDepotName, thaliBranchName,
     projectDir, 'src', 'android', 'gradle.properties');
   var fileContents = util.format(
     'btconnectorlib2Version = %s\n' +
-    'ext.cdvMinSdkVersion = %s\n' +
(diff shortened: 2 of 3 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (3f1da2e)

ThaliTester:
Test 97851490 (3f1da2e) build started.

ThaliTester:
Test (Success) 97851490 build is completed (3f1da2e)

See https://github.com/ThaliTester/TestResults/tree/978514903f1da2e__1632_-_Fixing_broken_build_yaronyg/ for the logs

ThaliTester:
Test 978514903f1da2e(3f1da2e) has failed

See https://github.com/ThaliTester/TestResults/tree/978514903f1da2e__1632_-_Fixing_broken_build_yaronyg/ for the fail logs

chapko:
Reviewed 3 of 3 files at r1.

---

*[thali/install/setUpDesktop.sh, line 15 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1636#-KYxRvTmcAaHRMPlTxA8:-KYxRvTmcAaHRMPlTxA9:bmddyal) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/aaf35e740ed7aeb7661258028b64a4bdfb27d4b7/thali/install/setUpDesktop.sh#L15)):*
> ```Shell
> # These variables are set since a bug in jxcore
> # `jx npm install` and `jx install` have different behaviours
> NVM_NODEJS_ORG_MIRROR="${NVM_NODEJS_ORG_MIRROR-https://jxcore.azureedge.net}"
> ```

Why it is removed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1636)*

yaronyg:

---

*[thali/install/setUpDesktop.sh, line 15 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1636#-KYxRvTmcAaHRMPlTxA8:-KYyM8dt5Rtr1EzHzmNq:bdpve8r) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/aaf35e740ed7aeb7661258028b64a4bdfb27d4b7/thali/install/setUpDesktop.sh#L15)):*
> setUpDesktop - The logic we were using which would use the global variable if already defined otherwise it would set it doesn't work in the case of NVM_NODEJS_ORG_MIRROR if someone is using NVM which also sets this variable. Since NVM is more common than needing to change this location when doing our build I decided to revert that change and just overwrite the variable when we are doing a build.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1636)*

Issue #1234, opened 2016-10-03
express-pouchdb middleware doesn't close file descriptors after server is closed


chapko:
I added a workaround for this in `master_chapko_pouchdb-closing-fix` to make tests pass. Currently I just destroy database before closing server which removes db files and frees opened file descriptors.

PR #1664, opened 2016-12-20
#1629 Fix Android BLE Zombie Problem

### LINKED ISSUE #1629: Fix Android BLE Zombie Problem
## The Problem
Thali uses BLE to advertise the presence of a Thali device. In that advertisement is a Bluetooth MAC address and a one byte generation counter. In theory we should get a nice stream of events where each event has a higher generation (accounting for roll over) than the previous one.

However we are getting repeated situations where we get BLE discovery announcements for the same BLE MAC + Bluetooth MAC in an impossible generation order. 

For example, we'll get Generation 0 and then 3 and then 2 and then 4 and so on. We also get repeated generations. For example, we'll get Generation 0 and then Generation 1 and then another Generation 0.

From the tests we have run the problem seems to be in Android itself. There is something screwy in the queue that discovery announcements are put into that causes them to come out in random orders and with repeats. We call these outdated announcements 'zombies'. Since we can't figure out how to fix the Zombies in Android, we have to deal with it.

## Why we care

In theory we really could ignore this problem. If we treat generations not as integers but rather as a 'I have new data' flag then having generations go backwards and forwards or repeat won't matter. At worst it will just create spurious flags saying "I have new data" when in fact the remote peer doesn't. In that case we will just waste a Bluetooth connection talking to a peer who doesn't actually have anything for us.

But even that isn't a minor issue. It is really expensive to create a Bluetooth connection and while we are doing that we are largely unavailable for replication with other peers and even BLE discovery goes wobbly because of contention for the 2.4 GHz antenna. So fake "I have new data" flags are actually kind of expensive.

So we are motivated to try and filter out zombies.

## How do we filter out zombies?

Ignoring roll over for just a second we can do some fairly simple things. For example, if our first discovery of a peer was on generation 0 and then 2 seconds later we get generation 4 we know something is wrong. The reason is that currently we only send generation updates no faster than once a second. So if we got generation 0 2 seconds ago and then get generation 4 we know something is very wrong.

Unfortunately there are three possibilities and there is no clear way to know which possibility it is:

Possibility 1 - The generation 0 was the 'real' value and generation 4 is a zombie from the very distant past (there would have to be a wrap around to account for it)

Possibility 2 - The generation 0 was the zombie and generation 4 is the real value.

Possibility 3 - Both values are zombies and we still haven't gotten the current real value yet

So in thinking about this problem we have two basic approaches.

Approach 1 - Limited Zombie Time Frame - If we know that Zombies couldn't be older than some reasonably small fixed value then we could reason in the face of Zombies. But we actually don't know how long a value could get stuck in the Bluetooth queue before it shakes lose and is sent to us.

Approach 2 - Unknown Zombie Time Frame - In this scenario we assume that a Zombie could be stuck in the Bluetooth queue for an unbounded time so we can't trust anything.

Now in thinking through this keep in mind that eventually we are going to switch to using a BLE GATT server and that will (as far as we know) make the Zombie issue go completely away. But until then we need to decide how to handle this mess.

### Unknown Zombie Time Frame

I generally like to make as few assumptions as possible so the first place I want to start with is the unknown zombie time frame. What can we do to handle Zombies if we never know if any value is a Zombie or not?

One possibility is that we introduce our one 1 second rate limiter. Right now Android will forward us notification updates as soon as it gets them. So if we get updates faster than 1 a second then we know that those announcements must be zombies because announcements are only sent out at the rate of 1/second. So we can put in a filter that ignores any peerAvailabilityChanged events from the Android native layer that arrive more frequently than once a second. 

The problem is that it's not clear how often zombies show up. If they all show up at once then this is a reasonable approach. But if they are going to trickle out then this will force us to create many unnecessary bluetooth connections which is bad. It really hurts our perf.

But what if we didn't use 1/second? What if we used say 1/5 seconds or 1 event every 10 seconds? This isn't as loony as it might sound. The main downside to this approach is that it might cause us to wait too long before doing an update but if we are in the background that isn't too bad. And if we are in the foreground and the app wants us to keep a replication going then it should just do that by extending the replication timeout rather than depending on notifications of changes and expensive bluetooth connection establishment.

#### Limited Zombie Time Frame

One suspects (but we have not proven) that the Zombie problem has a limit. That we aren't going to get Zombies from an hour ago if only because the queue that this data is stored in by Android (or the hardware) probably isn't that big. So what if there is a time limit to how old a Zombie could be? Let's call that limit L seconds.

In that case whenever we get a generation we know it could be 'off' by up to L seconds. In other words if L is say 60 seconds and we get a generation 0 then we know that the 'real' generation must be a value between 0 and 60 (remember that we already have logic that guarantees we won't publish generation updates to BLE faster than 1/second).

So once we get an advertisement with generation 0 we know:
* It MUST NOT be possible for us to immediately get a generation with 61-195 because if we do then L is wrong and we have problems. Note that this window will obviously shift as time passes.
* If we see future announcements with a generation less than our 'current value' (e.g. 0 + however many seconds have passed) then we can ignore it as a Zombie. Keep in mind roll over. This means that if 2 seconds pass and we get generation 199 then we can ignore this because even with the L error not enough time has passed for a roll over.
* If we see a generation higher than ours that is within the legal window (e.g. L + time that has passed + generation, account for roll over) then that is our new base and we have to jump to it.

In essence we create a ratchet. So long as L is >= to the real Zombie time limit we will only move forward and will be able to ignore a lot of (but not all of) the noise.

```
function canIgnoreNewGeneration(lastGeneration, newGeneration) {
   assert(newGeneration is integer and >= 0 and <= 255);
   timeSinceLastGeneration = Time in Seconds that have elapsed since the last generation value was received
   // So much time has passed that any generation is possibly real
   if (L + timeSinceLastGeneration >= 256) return false;
   leftBorder = lastGeneration + 1;
   rightBorder = (lastGeneration + L + timeSinceLastGeneration) mod 256;
   if (leftBorder < rightBorder) {
      return !(newGeneration >= leftBorder && newGeneration <= rightBorder);
   }
   return !(newGeneration >= leftBorder || newGeneration <= rightBorder);
}

if (canIgnoreNewGeneration(lastGeneration, newGeneration) == true) {
  return
}
```

## Fixing the generation counter

Android's behavior is a pain. We have to deal with Zombies. We have to deal with roll over. It makes life for those up the stack really hard. To simplify we need to not let the generation on the wire be sent up the stack from thaliMobile. Instead we should generate our own generation counter and every time we update the cached entry we should update that generation counter. It is the 'fake' generation counter that we should send up the stack. That way those up the stack only have to deal with a linearly increasing counter.

## The Plan

For no particularly good reason I'm picking the limited Zombie Time Frame approach. And while we are there we also need to fix the generation counter.

### thaliConfig
- [ ] Add a value 'ANDROID_ZOMBIE_TIME_IN_SECONDS' and set it to 60, this will be the L value mentioned in canIgnoreNewGeneration

### thaliMobile.handlePeer
- [ ] If we get peerAvailable === false on the Android platform then immediately call changedCachedPeerUnavailable(peer). It doesn't matter what generation is being sent because !peerAvailable is generated via a heuristic in Android land (we didn't hear any announcements from the peer for a minute) not as part of an explicit message like 'byebye' or an advertiser being shut down in iOS. So if it says a particular peerID is gone then that peerID is gone regardless of generation. Note that the iOS logic doesn't change. It should work exactly as it currently does.
- [ ] changed thaliMobile.handlePeer so that on Android if there is a cachedPeer and if peer.peerAvailable === true then we call something like canIgnoreNewGeneration(cachedPeer.generation, peer.Generation) and if it's false then we just return without taking any action since we assume we got a zombie. Don't even update peer.availableSince as this is a zombie. Basically, if canIgnoreNewGeneration is false then we should just pretend we never got called.
- [ ] On the Android platform we should generate our own generation counter in thaliMobile. So the first time we add an entry on Android for cachedPeer we MUST set a value 'fakeGeneration' to 0 and record it on the peer object so it will be placed into the cache by changeCachedPeerAvailable. And in peerStatus (the value we will emit as a peerAvailabilityChanged event) we have to set generation to peer.fakeGeneration. Note that this doesn't affect iOS or Wifi so those platforms must set generation to peer.generation.
- [ ] Any time we get a peerAvailablityChanged event on Android that has an existing cachedPeer entry then we have to add a peer.fakeGeneration = cachedPeer.fakeGeneration + 1;

chapko:
Fixes #1629

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1664)


ThaliTester:
PR is added to the queue for testing as 1. task. (7d9c273)

ThaliTester:
Test 98761056 (7d9c273) build started.

ThaliTester:
Test (Fail) 98761056 build is completed (7d9c273)

See https://github.com/ThaliTester/TestResults/tree/987610567d9c273__1629_Fix_Android_BLE_Zombie_Problem_chapko/ for the logs

chapko:
Build fails because of #1665. All other tests (without 'testThaliMobileNativeDiscoveryCoordinated.js') are running fine on three Nexuses

yaronyg:
This is quite awesome! But I really think that the zombie filter and throttle/throttled need to be changed into classes using constructor/new to make the code much easier to understand. I really struggled to understand the code because it did things that look like class behavior but it didn't follow the Javascript idioms for that like using 'new' and prototype.

---

Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[thali/NextGeneration/utils/common.js, line 11 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSpqiDNKxeuJw41LOe:-KZSpqiDNKxeuJw41LOf:b-ysxt23) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/common.js#L11)):*
> ```JavaScript
> };
> 
> module.exports.makeAsync = function (fn) {
> ```

It would probably be a good idea to put in a note about why this function is necessary at all. I have some guesses but honestly I'm not 100% sure. This is exactly the sort of esoteric piece of code that will cause future devs to be scared because they won't know what edge case drove the need for it.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 12 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSEuaShTDSvKbmH0A1:-KZSEuaShTDSvKbmH0A2:b-u51mhx) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L12)):*
> ```JavaScript
> 
> /**
>  * Throttled function.
> ```

Isn't this what localSeqManager does with update calls? Shouldn't we re-write localSeqManager to use this function? We don't want identical functionality in two places do we?

---

*[thali/NextGeneration/utils/zombieFilter.js, line 67 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSlMGS9iQZjckGcUnO:-KZSlMGS9iQZjckGcUnP:b6sk3hh) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L67)):*
> ```JavaScript
>  * @returns {ThrottledFunction} throttled version of `fn`
>  */
> function throttle(fn, options) {
> ```

I found the structure of this function difficult to follow. It uses the generator/factory pattern but that is pretty rare in Javascript land. Typically we use the new pattern. So I would have expected throttle to be 'new Throttle'. The reason folks favor 'new' is because Javascript has a class like capability that makes it much easier to understand the scope of variables via 'this' or 'self' as created on the constructor. In a generator/factory one has to kind of figure out which variables are the moral equivalent of a constructor variable and which ones are local to a method.

So please change this to a constructor/new pattern so we can much more easily understand the code.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 80 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSltMFZog6l8DRUZBn:-KZSltMFZog6l8DRUZBo:b-54lfre) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L80)):*
> ```JavaScript
>     fn.apply(context, args);
> 
>     if (!timeout) {
> ```

Timeout was set to null above so isn't this check always true and thus unnecessary?

---

*[thali/NextGeneration/utils/zombieFilter.js, line 85 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSmnh0GakvaE9MFCoj:-KZSmnh16FWbb6NrHUgG:b-kyub7x) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L85)):*
> ```JavaScript
>   }
> 
>   function throttled () {
> ```

I've reviewed this file and I'm completely unclear on when throttled is actually called. Part of the reason for this I suspect is because throttle() outputs throttled who then gets called under a different name. I'm sure I'll eventually figure it out (and I eventually did) but my point is that it shouldn't have been so hard to understand the flow. The way this function is written makes it really difficult for me to understand how its used by the cache filter. Please, fix by switching to 'new' and using explicit method calls.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 96 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSkuKUjRIcVhHypHpY:-KZSkuKUjRIcVhHypHpZ:b-2z876l) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L96)):*
> ```JavaScript
> 
>     if (!timeout) {
>       timeout = global.setTimeout(invoke, minDelay);
> ```

We really need a comment here explaining why global is necessary. A reference to the relevant bug should do it.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 138 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSojtpJRNLORiLr1bV:-KZSojtqSGnhpDn6SKhY:b-2660q3) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L138)):*
> ```JavaScript
>   var newCachedPeer = {
>     nativeGeneration: nativeGeneration,
>     fakeGeneration: oldCachedPeer ? oldCachedPeer.fakeGeneration : 0,
> ```

I find the way that fakeGeneration was used here to be surprising. When I read 'fakeGeneration' that means to me 'The fake value that represents the current generation'. But that doesn't appear to be true. It seems to really be 'the fake generation value that should be used for the next event'. Which as I said, I found surprising. I would suggest that we remove ++fakeGeneration from fixedPeerHandler and instead move it here where we increment oldCachedPeer.fakeGeneration before returning it here.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 267 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSpGnh2M7xkT_ont1j:-KZSpGnh2M7xkT_ont1k:b8cfcmc) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L267)):*
> ```JavaScript
>     cachePeer(cache, peerIdentifier, nativePeer.generation, handler);
> 
>     handler(nativePeer);
> ```

This is what I mean by confusing. To understand what the heck this function is one has to realize that it's actually a call to throttle/throttled. That is not what I could call obvious. But with a constructor pattern it would be.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 270 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSnh3NMBmrYeNX5L7M:-KZSnh3NMBmrYeNX5L7N:bw8x7nr) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L270)):*
> ```JavaScript
>   };
> 
>   filteredNonTCPPeerHandler.clearCache = clearCache.bind(null, cache);
> ```

Again, as with throttled, I think this approach is very surprising. Javascript has a well defined way to handle the pattern of adding a method to an object, which is using constructors and prototypes. By avoiding those mechanisms we create surprising code because the patterns defined by Javascript to handle exactly this scenario aren't being used so the developer can't make any assumptions about what is going on. Please turn all of this into a proper constructor with prototypes so we have code whose behavior will be clear to the reader.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-:-KZSsjoFm1Bu_yGMKqFl:b17mvet)*

yaronyg:
thaliproject/Thali_CordovaPlugin#1234

ThaliTester:
PR is added to the queue for testing as 1. task. (debde49)

ThaliTester:
Test 98761056 (debde49) build started.

chapko:
Review status: 4 of 8 files reviewed at latest revision, 9 unresolved discussions.

---

*[thali/NextGeneration/utils/common.js, line 11 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSpqiDNKxeuJw41LOe:-KZW9uZU_NhJFPzFKYn-:b-7eaejz) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/common.js#L11)):*

It would probably be a good idea to put in a note about why this function is necessary at all. I have some guesses but honestly I'm not 100% sure. This is exactly the sort of esoteric piece of code that will cause future devs to be scared because they won't know what edge case drove the need for it.
</blockquote></details>

Yes, I agree that it looks a bit esoteric so I added an example to make it clearer what it does. And I actually had comments in thaliMobile explaining why I use it but accidentally lost them after small refactoring.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 12 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSEuaShTDSvKbmH0A1:-KZWE8HuCaEADWwJ-nyQ:bo6rtlz) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L12)):*

Isn't this what localSeqManager does with update calls? Shouldn't we re-write localSeqManager to use this function? We don't want identical functionality in two places do we?
</blockquote></details>

It is similar but not quite the same. My `throttle` actually should be called `debounce`. It looks like LocalSeqManager uses actual throttle functionality

---

*[thali/NextGeneration/utils/zombieFilter.js, line 67 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSlMGS9iQZjckGcUnO:-KZWLOGVvL_mZlmb9Q2Z:b-2p3zhp) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L67)):*

I found the structure of this function difficult to follow. It uses the generator/factory pattern but that is pretty rare in Javascript land. Typically we use the new pattern. So I would have expected throttle to be 'new Throttle'. The reason folks favor 'new' is because Javascript has a class like capability that makes it much easier to understand the scope of variables via 'this' or 'self' as created on the constructor. In a generator/factory one has to kind of figure out which variables are the moral equivalent of a constructor variable and which ones are local to a method.

So please change this to a constructor/new pattern so we can much more easily understand the code.
</blockquote></details>

Well, I disagree with it. I used [`throttle`](http://underscorejs.org/docs/underscore.html#section-82) implementation from underscore.js for the reference. But final result looks more like highly simplified [`debounce`](https://github.com/lodash/lodash/blob/15983648a2cdb52f688aa9fe15111cc854251f49/lodash.js#L10321-L10441) from lodash. They both have quite the same structure.

The purpose of the `throttle` is to take some function and return "throttled" version of this function. Using `new Throttle` feels unnatural for me in this case because it is not clear what kind of object it produces, what it has to do with the functions and how it is expected to be used. And it is unnatural in the same way as for example using `new` to create callbacks.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 80 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSltMFZog6l8DRUZBn:-KZWTPUZabHu12t88IZ3:bhfykmm) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L80)):*

Timeout was set to null above so isn't this check always true and thus unnecessary?
</blockquote></details>

When I saw it in underscore's implementation I had the same question. Before this `if` statement we call original `fn` which may call throttled function again. In this case it would create new timeout.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 85 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSmnh0GakvaE9MFCoj:-KZWoXkLW2jl45CsMCmv:bdwim3a) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L85)):*

I've reviewed this file and I'm completely unclear on when throttled is actually called. Part of the reason for this I suspect is because throttle() outputs throttled who then gets called under a different name. I'm sure I'll eventually figure it out (and I eventually did) but my point is that it shouldn't have been so hard to understand the flow. The way this function is written makes it really difficult for me to understand how its used by the cache filter. Please, fix by switching to 'new' and using explicit method calls.
</blockquote></details>

Please, see my comment above.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 96 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSkuKUjRIcVhHypHpY:-KZWoarOdHCd93k5IpEW:bb6un1j) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L96)):*

We really need a comment here explaining why global is necessary. A reference to the relevant bug should do it.
</blockquote></details>

Done

---

*[thali/NextGeneration/utils/zombieFilter.js, line 138 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSojtpJRNLORiLr1bV:-KZXN17ljAqz3O9TKl2k:bbnp7hf) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L138)):*

I find the way that fakeGeneration was used here to be surprising. When I read 'fakeGeneration' that means to me 'The fake value that represents the current generation'. But that doesn't appear to be true. It seems to really be 'the fake generation value that should be used for the next event'. Which as I said, I found surprising. I would suggest that we remove ++fakeGeneration from fixedPeerHandler and instead move it here where we increment oldCachedPeer.fakeGeneration before returning it here.
</blockquote></details>

I have updated cache code. It is now written in OOP style, which makes more sense here

---

*[thali/NextGeneration/utils/zombieFilter.js, line 267 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSpGnh2M7xkT_ont1j:-KZaIQ0p5b3APxZ_f5uY:b-4xmick) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L267)):*

This is what I mean by confusing. To understand what the heck this function is one has to realize that it's actually a call to throttle/throttled. That is not what I could call obvious. But with a constructor pattern it would be.
</blockquote></details>

I refactored it a bit. Should be clearer now.

---

*[thali/NextGeneration/utils/zombieFilter.js, line 270 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664#-KZSnh3NMBmrYeNX5L7M:-KZaIW9_xZ32PThKXElD:b-238jcv) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/7d9c273a065512d95a9e64dec3ca1bdf23ac323a/thali/NextGeneration/utils/zombieFilter.js#L270)):*

Again, as with throttled, I think this approach is very surprising. Javascript has a well defined way to handle the pattern of adding a method to an object, which is using constructors and prototypes. By avoiding those mechanisms we create surprising code because the patterns defined by Javascript to handle exactly this scenario aren't being used so the developer can't make any assumptions about what is going on. Please turn all of this into a proper constructor with prototypes so we have code whose behavior will be clear to the reader.
</blockquote></details>

With OOP styled cache it is not confusing anymore (I hope)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1664)*

chapko:
Closing this PR until we have some results from the native layer

ThaliTester:
Test (Fail) 98761056 build is completed (debde49)

See https://github.com/ThaliTester/TestResults/tree/98761056debde49__1629_Fix_Android_BLE_Zombie_Problem_chapko/ for the logs

Issue #1219, opened 2016-09-29
We really need to refactor thaliMobile and other huge files in Thali


yaronyg:
thaliMobile is now 1200 lines long and contains things (like constants and other functions) that other folks need but don't need all of thaliMobile. This is true for other files as well. We really should break them up.

PR #1772, opened 2017-02-01
Clean up ThaliWifiInfrastructure

chapko:
Original implementation of ThaliWifiInfrastructure consists of the
single entity that is responsible for several unrelated and independent
tasks: listening, advertising, network changes handling, async calls
management, and more.

The purpose of this change is to split these responsibilities between
different units, make it easier to understand and test, while providing
the original API to the upper layers.

So, the result is 3 independent entities:

- WifiAdvertiser - hosts router server and advertises current peer in
  the WiFi network
- WifiListener - listens for advertisements and filters
  invalid/irrelevant messages
- ThaliWifiInfrastructure - assembles advertiser and listener, queues
  async calls, and handles network changes

Also there are some fixes in native mock and updated node-ssdp

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1772)


ThaliTester:
PR is added to the queue for testing as 1. task. (c83cb35)

ThaliTester:
Test 104148237 (c83cb35) build started.

ThaliTester:
Test 104148237c83cb35(c83cb35) has failed

See https://github.com/ThaliTester/TestResults/tree/104148237c83cb35_Clean_up_ThaliWifiInfrastructure_chapko/ for the fail logs

yaronyg:
Reviewed 14 of 14 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[test/www/jxcore/UnitTest_app.js, line 11 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5UAxd4ELhatZBtyaQ:-Kc5UAxeSu8vnW4RNh8X:b-ss3udd) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/test/www/jxcore/UnitTest_app.js#L11)):*
> ```JavaScript
> var Promise = require('bluebird');
> Promise.config({
>   cancellation: true
> ```

Why is this set to true when we don't appear to actually call .cancel anywhere? Do we have some dependent library that does?

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-Kc5RfoC3aWO6o_9qhPk:bz77lvw) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*
> ```JavaScript
>     return promise.then(returnNull).catch(returnArg);
>   };
> }());
> ```

Why isn't this just: 
function(promise) {
   return promise.then(null).catch(function(arg) { return arg; {);
}
?

Also note that muteRejection isn't the equivalent of promiseResultSuccessOrFailure. In promiseresultSuccessOrFailure if the promise returns successfully with some value we return that value. In muteRejection if the promise returns successfully we turn the returned value to null. I would have expected:

function (promise) {
   function returnArg(arg) { return arg; }
   return promise.then(returnArg).catch(returnArg);
}

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 295 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5PwDESU9XW6cqz6f1:-Kc5PwDESU9XW6cqz6f2:b24gcno) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobileNativeWrapper.js#L295)):*
> ```JavaScript
> // jscs:enable jsDoc
> module.exports.start = function (router, pskIdToSecret) {
>   targetStates.started = true;
> ```

Yup, same comment as always. Everything has to be inside of the promise queue or we risk race conditions. I'll just leave this one comment in this file. Please fix in other places in this file like stop.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 56 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc4-m4_RzoMVnT41DIn:-Kc4-m4_RzoMVnT41DIo:b-apaxnk) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L56)):*
> ```JavaScript
>  * @class WifiListener
>  */
> function WifiListener() {
> ```

This whole file is absurdly long. I think it would be a good idea if we could refactor wifiListening into its own stand alone file.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 134 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc4-yy6vb8nPAseEtdE:-Kc4-yy6vb8nPAseEtdF:bn2immp) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L134)):*
> ```JavaScript
> };
> 
> /**
> ```

NB: I would move this function before the previous function to make reading this file easier.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 181 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5CvmLkAixwGoVNpDq:-Kc5CvmLkAixwGoVNpDr:b-2petnj) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L181)):*
> ```JavaScript
>   self._isListening = false;
> 
>   return self._client.stopAsync().then(function () {
> ```

Isn't there a race condition here where someone calls stop, we immediately set _isListening to false and then call stopAsync. Meanwhile someone could immediately call start which will have us call startAsync. Are we use that calling startAsync before stopAsync has returned is safe?

Even if it is safe, will it cause any interesting issues with self._notifyStateChange potentially getting called twice with _isListening = true (because self._notifyStateChange here only gets called once stopAsync has returned and by that point startAsync might have been called?

This is why I use the promise queue approach, to avoid having to figure this stuff out. :)

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 234 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Deh2dD67rVJYD97N:-Kc5Deh2dD67rVJYD97O:bj7hnwr) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L234)):*
> ```JavaScript
> 
> /**
>  * @private
> ```

I would also put this into a separate file. 

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 338 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5ENtbNGzVLZO9nwuw:-Kc5ENtbNGzVLZO9nwux:bnvtzg3) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L338)):*
> ```JavaScript
>  * @return {Promise}
>  */
> WifiAdvertiser.prototype.stop = function () {
> ```

Same start and stop race conditions as with the listener. Again, I would use the promise queue to make the race conditions go away.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 405 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5GXNcOGvDntwv5Yj9:-Kc5GXNdcGKejya3nTa5:b-jshwgx) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L405)):*
> ```JavaScript
>   };
> 
>   function listen (server, port) {
> ```

This function seems a lot more complex than I think it needs to be. I would have done something like:

function listen (server, port) {
   return new Promise(function (resolve, reject) {
      server.listen(port, function (error) {
         return error ? reject(error) : resolve();
     })});
}

I don't think there is any need for the event handlers since the callback will return an error if one occurs and if there is no error then it means that listening worked.
         

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 423 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5HOoxXdDtBBRF1C2M:-Kc5HOoyCynbid8crBO9:b-p6fi4v) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L423)):*
> ```JavaScript
>   }
> 
>   self.routerServer = https.createServer(options, self.expressApp);
> ```

I realize that assigning to self.routerServer isn't an error but the code still causes surprise since, unless you understand exactly what makeIntoCloseAllServer does, it looks like an error. I would either inline the call to https.createServer in the next line or use a different variable to hold the interim result.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 432 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Hr8fzAmBQ5uf3xD1:-Kc5Hr8gqUkjdggBaudI:b-6j737p) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L432)):*
> ```JavaScript
>       );
>       self.routerServer = null;
>       var error = new Error('Unspecified Error with Radio infrastructure');
> ```

Shouldn't we pass the error string as part of the error here? E.g. add '+ listenerError to the string above? I realize it's already on the log but finding and correlating these things is a nightmare.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 764 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5IpjvcXtnu5jlchOG:-Kc5IpjvcXtnu5jlchOH:b-gio00a) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L764)):*
> ```JavaScript
>   }
> 
>   this._isStarted = true;
> ```

It's really dangerous to do any work before you put a command into the promise queue as this leads to exactly the kind of race conditions that the promise queue was intended to stop. I realize that there is a minor improvement in response time by doing things like checking for errors before jumping into the queue but I would suggest this is a false savings. Whatever tiny perf improvement it gives is overset by now having to make sure that stop and start don't have any possible conflict for the code run off the promisequeue. So my recommendation is to put all the code here into the promisequeue.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 790 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Jb8XO6__wdL5jzzm:-Kc5Jb8XO6__wdL5jzzn:b7dwmce) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L790)):*
> ```JavaScript
>  */
> ThaliWifiInfrastructure.prototype.stop = function () {
>   thaliMobileNativeWrapper.emitter
> ```

Same as before. I would stick everything into the promiseQueue. That way there is no possibility of any race conditions. I realize that it leaves open the possibility of getting a networkChangedNonTCP event after we have called stop but that isn't an error since technically stop hasn't been run until it returns. So we are o.k.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 835 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5JqlYimLdbvbflRt6:-Kc5JqlYimLdbvbflRt7:b-mx1adf) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L835)):*
> ```JavaScript
> function () {
>   var self = this;
>   self._isListening = true;
> ```

Same as before. By moving this outside of the promise queue you create fun race conditions where someone calls start and then immediately calls stop and by the time the body of the start code is run off the enqueue _isListening = false. That's why you have to put *all* the code into the promiseQueue.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 871 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5K4GRHJSxU55Ist2c:-Kc5K4GRHJSxU55Ist2d:bpvf4i2) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L871)):*
> ```JavaScript
> ThaliWifiInfrastructure.prototype.stopListeningForAdvertisements =
> function () {
>   this._isListening = false;
> ```

I know you are bored of my comment about promiseQueue but we have to catch every instance. I'm going to stop now and assume you will search for promiseQueue nad make sure that all the code (especially setting variables like this._isListening or other global state) is only done inside of the promiseQueue to stop race conditions.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 881 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5KK6Ftcm2X4llo4vE:-Kc5KK6Gekq3JdRQBriz:b-1654tv) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L881)):*
> ```JavaScript
> ThaliWifiInfrastructure.prototype._pauseListeningForAdvertisements =
> function () {
>   var listener = this.listener;
> ```

This is dangerous too since by the time the promiseQueue is called the active listener could have changed. Instead we need to set var self = this; and then inside of the promise queue get the listener (if set).

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 998 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5KXXVlG_zdhXVux8Q:-Kc5KXXVlG_zdhXVux8R:b-rg5i98) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L998)):*
> ```JavaScript
> 
> ThaliWifiInfrastructure.prototype._pauseAdvertisingAndListening = function () {
>   var advertiser = this.advertiser;
> ```

Same issue as above with the advertiser changing by the time the stop below is called. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772)*

ThaliTester:
PR is added to the queue for testing as 1. task. (1c74b9d)

ThaliTester:
Test 104148237 (1c74b9d) build started.

ThaliTester:
Test (Fail) 104148237 build is completed (1c74b9d)

See https://github.com/ThaliTester/TestResults/tree/1041482371c74b9d_Clean_up_ThaliWifiInfrastructure_chapko/ for the logs

ThaliTester:
PR is added to the queue for testing as 1. task. (4baf670)

ThaliTester:
Test 104148237 (4baf670) build started.

ThaliTester:
Test (Success) 104148237 build is completed (4baf670)

See https://github.com/ThaliTester/TestResults/tree/1041482374baf670_Clean_up_ThaliWifiInfrastructure_chapko/ for the logs

chapko:
I have merged #1773 PR into this branch because it is too hard to consider them separately of each other.

@yaronyg You can ignore my strikethrough replies below, I've started answering last week but things changed after the weekend.

---

Review status: 3 of 15 files reviewed at latest revision, 18 unresolved discussions.

---

*[test/www/jxcore/UnitTest_app.js, line 11 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5UAxd4ELhatZBtyaQ:-KcT1MhMBwER_oXTEpXs:boun0e5) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/test/www/jxcore/UnitTest_app.js#L11)):*

Why is this set to true when we don't appear to actually call .cancel anywhere? Do we have some dependent library that does?
</blockquote></details>

Probably some of the tests require cancellation. I just moved it here from testUtils.js because cancellation should be configured before creating promises.

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-KcT3EUDUZ_c_tCHWmLz:b-satj3m) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*

Why isn't this just: 
function(promise) {
   return promise.then(null).catch(function(arg) { return arg; {);
}
?

Also note that muteRejection isn't the equivalent of promiseResultSuccessOrFailure. In promiseresultSuccessOrFailure if the promise returns successfully with some value we return that value. In muteRejection if the promise returns successfully we turn the returned value to null. I would have expected:

function (promise) {
   function returnArg(arg) { return arg; }
   return promise.then(returnArg).catch(returnArg);
}
</blockquote></details>

It worked because all methods wrapped in `promiseResultSuccessOrFailure` always return null after success. For example, in `ThaliMobile.start` you can see

```JavaScript
.then(function (result) {
  if (result.wifiResult === null && result.nativeResult === null) {
    return result;
  }
  return Promise.reject(result);
});
```

It assumes that if result is not null, then there is an error. That's why I changed it to explicitly return null on success.

---

*[thali/NextGeneration/thaliMobileNativeWrapper.js, line 295 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5PwDESU9XW6cqz6f1:-KcTUOFd2DX5Vw29P6Ck:bsueog9) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobileNativeWrapper.js#L295)):*

Yup, same comment as always. Everything has to be inside of the promise queue or we risk race conditions. I'll just leave this one comment in this file. Please fix in other places in this file like stop.
</blockquote></details>

It was done intentionally. The idea is that there are 2 states - current state (or just `state`) and target state (`targetState`). `state` is the actual state that updated asynchronously, while `targetState` is the desired state and updated synchronously. When there are no jobs enqueued in the `promiseQueue` these states are the same. When there are some enqueued actions, `targetState` represents what the real state will be after the queue is processed.

Target state is used solely in networkChanged handler to determine if it should enqueue advertising or listening, because target state always reflects user desire, and target values can be changed only by calling start/stop methods.

BTW the same idea was previously used in the thaliWifiInfrastructure (but it was somewhat inconsistent), this is the old initial state object:
```Javascript
{
  started: false,
  stopping: false,
  listening: {
    target: false,
    current: false
  },
  advertising: {
    target: false,
    current: false
  },
  networkState: null
};
```

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 56 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc4-m4_RzoMVnT41DIn:-KcT6urHthaA45DR4O1q:b488bte) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L56)):*

This whole file is absurdly long. I think it would be a good idea if we could refactor wifiListening into its own stand alone file.
</blockquote></details>

That's what I kept in mind doing these changes. And it should be really simple. But
1. ~~Listener and advertiser are not safe to use outside of the promiseQueue because of possible race conditions~~ (fixed).
2. The similar thing should be done with thaliMobileNativeWrapper. I don't want to split this file until it is clear in what way native wrapper could be refactored. Only after that I can decide where exactly to put these modules.
3. Tests should be refactored the same way.

And actually there is #1219 issue for this, I suppose.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 134 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc4-yy6vb8nPAseEtdE:-KcTgdG9PASZnA8Y_bex:bis0m37) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L134)):*

NB: I would move this function before the previous function to make reading this file easier.
</blockquote></details>

Agree. Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 181 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5CvmLkAixwGoVNpDq:-KcTB5LkQMvFC-XSRhMg:bkyvwia) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L181)):*
> Are we use that calling startAsync before stopAsync has returned is safe?

~~I cannot guarantee it, that's why there is a promise queue in the ThaliWifiInfrastructure that solves these races. And this race condition is one of the reasons I don't export WifiListener and WifiAdvertiser and made them only accessible internally to this module.~~

> will it cause any interesting issues with self._notifyStateChange potentially getting called twice with _isListening = true

~~Yes it is possible, but with current promise queue approach it cannot happen. Once the above race condition is fixed this one will be probably fixed too.~~

> This is why I use the promise queue approach, to avoid having to figure this stuff out. :)

~~WifiListener and WifiAdvertiser are not exported, so they can't be used directly outside of the promise queue.~~

I wrapped these methods in a promise queue. This should fix race conditions listed above.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 234 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Deh2dD67rVJYD97N:-KcTgf8mDlmdhR2DH1rc:bg2cy2p) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L234)):*

I would also put this into a separate file. 
</blockquote></details>

I answered it in the comments above.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 338 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5ENtbNGzVLZO9nwuw:-KcTgspf1rd9tkH3703x:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L338)):*

Same start and stop race conditions as with the listener. Again, I would use the promise queue to make the race conditions go away.
</blockquote></details>

Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 405 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5GXNcOGvDntwv5Yj9:-KcTJXtqssXuBZHKlksZ:b822fi0) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L405)):*

This function seems a lot more complex than I think it needs to be. I would have done something like:

function listen (server, port) {
   return new Promise(function (resolve, reject) {
      server.listen(port, function (error) {
         return error ? reject(error) : resolve();
     })});
}

I don't think there is any need for the event handlers since the callback will return an error if one occurs and if there is no error then it means that listening worked.
         
</blockquote></details>

```JavaScript
server.listen(port, callback);
```
is equivalent to
```JavaScript
server.listen(port);
server.on('listening', callback);
```
In case of error listen will just throw (if there are no listeners for 'error' event). Since server comes as an argument to this function it is not clear if it has 'error' listeners or not.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 423 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5HOoxXdDtBBRF1C2M:-KcsDUKsmgMKJbftVZCy:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L423)):*

I realize that assigning to self.routerServer isn't an error but the code still causes surprise since, unless you understand exactly what makeIntoCloseAllServer does, it looks like an error. I would either inline the call to https.createServer in the next line or use a different variable to hold the interim result.
</blockquote></details>

Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 432 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Hr8fzAmBQ5uf3xD1:-KcTS3OtsARxIeHBnvH1:bketwu1) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L432)):*

Shouldn't we pass the error string as part of the error here? E.g. add '+ listenerError to the string above? I realize it's already on the log but finding and correlating these things is a nightmare.
</blockquote></details>

I agree, it makes more sense. I just reused the old code here. But this error string is defined in the spec and we have some tests that depend on it. Maybe we should rethink all errors first? Probably numeric error codes or better custom error types. Relying only on the string value feels too fragile.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 764 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5IpjvcXtnu5jlchOG:-KcTiv_6Z3JQOHYipSSU:b-3m0rm3) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L764)):*
I should rename it to the `targetState` as in thaliMobileNativeWrapper, `_isStarted` looks really confusing. But the idea is the same, it represents the desired target state. If I put it in the promise queue there would no way to know anything about the final state. In a network changed handler I need to know if I should enqueue one more `startListeningForAdvertisiments` call. But there is no way to tell if someone has already called stopListening or if there any enqueued actions or not. Promise queue is kind of a black box in which you can only put new actions.

> It's really dangerous to do any work before you put a command into the promise queue as this leads to exactly the kind of race conditions that the promise queue was intended to stop

It becomes dangerous when you cannot guarantee the order of execution. Synchronous calls don't need anything here because they are always executed in the right order. Asynchronous calls are managed by promise queue which guarantees that all actions start and end in the order they were called. To avoid race conditions you should not share any data between synchronous and asynchronous functions. `_isStarted`, `_isAdvertising` and `_isListening` never used inside of a queued actions, they are always accessed and modified synchronously.

The problem is that it adds more complexity and does not look obvious. I need to think if it can be done in a simpler way. Otherwise I will probably have to use 2 kinds of states or switch back to `enqueueAtTop` approach.

--

After some thinking I figured out that my approach also produces some race conditions (not exactly here, but in WifiListener and WifiAdvertiser),  so I wrapped everything except target state into the promise queue. And named things more correctly.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 790 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Jb8XO6__wdL5jzzm:-KcsD-Rnm1Rwm7Xj2D_Q:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L790)):*

Same as before. I would stick everything into the promiseQueue. That way there is no possibility of any race conditions. I realize that it leaves open the possibility of getting a networkChangedNonTCP event after we have called stop but that isn't an error since technically stop hasn't been run until it returns. So we are o.k.
</blockquote></details>

Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 835 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5JqlYimLdbvbflRt6:-KcU3dylU5KSP4iGzQTU:bbwsvio) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L835)):*

Same as before. By moving this outside of the promise queue you create fun race conditions where someone calls start and then immediately calls stop and by the time the body of the start code is run off the enqueue _isListening = false. That's why you have to put *all* the code into the promiseQueue.
</blockquote></details>

~~Thats why I never use _isListening or any other state flags inside of enqueued functions. So when the body of the function starts its execution I don't care what _isListening is equal to :)~~

Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 871 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5K4GRHJSxU55Ist2c:-KcsCd0jtx3itUHwadsA:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L871)):*

I know you are bored of my comment about promiseQueue but we have to catch every instance. I'm going to stop now and assume you will search for promiseQueue nad make sure that all the code (especially setting variables like this._isListening or other global state) is only done inside of the promiseQueue to stop race conditions.
</blockquote></details>

Done.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 881 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5KK6Ftcm2X4llo4vE:-KcU4dPuc18Z3cy7D7OS:b-r7u9ud) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L881)):*

This is dangerous too since by the time the promiseQueue is called the active listener could have changed. Instead we need to set var self = this; and then inside of the promise queue get the listener (if set).
</blockquote></details>

`this.listener` is set only once in constructor but I agree with this.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 998 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5KXXVlG_zdhXVux8Q:-KcU5-wKCCUL_aW2Vj7B:b-37fsjh) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L998)):*

Same issue as above with the advertiser changing by the time the stop below is called. 
</blockquote></details>

Agree. Will fix it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-:-KcsFwpV8JvXp5XKfWCb:b-i542pb)*

ThaliTester:
Test 1041482374baf670(4baf670) has failed

See https://github.com/ThaliTester/TestResults/tree/1041482374baf670_Clean_up_ThaliWifiInfrastructure_chapko/ for the fail logs

yaronyg:
Reviewed 4 of 4 files at r3, 8 of 8 files at r4.

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-KcsmgWgSEsgj1g9mu1N:bv9ewuf) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

It worked because all methods wrapped in `promiseResultSuccessOrFailure` always return null after success. For example, in `ThaliMobile.start` you can see

```JavaScript
.then(function (result) {
  if (result.wifiResult === null && result.nativeResult === null) {
    return result;
  }
  return Promise.reject(result);
});
```

It assumes that if result is not null, then there is an error. That's why I changed it to explicitly return null on success.
</blockquote></details>

But if we depend on the response being null shouldn't we check that the result is null?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 56 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc4-m4_RzoMVnT41DIn:-KcsndjPp8Tkr3886ty_:b1ywljq) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L56)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

That's what I kept in mind doing these changes. And it should be really simple. But
1. ~~Listener and advertiser are not safe to use outside of the promiseQueue because of possible race conditions~~ (fixed).
2. The similar thing should be done with thaliMobileNativeWrapper. I don't want to split this file until it is clear in what way native wrapper could be refactored. Only after that I can decide where exactly to put these modules.
3. Tests should be refactored the same way.

And actually there is #1219 issue for this, I suppose.
</blockquote></details>

We don't have to fix everything. It's fine to fix incrementally. We can fix just this file and leave notes about tests and such. If we think that every incremental improvement requires fixing the world then we'll never get anything done. I'm not super insistent on this refactoring but I wouldn't mind us starting to do refactorings bit by bit instead of all at once.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 405 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5GXNcOGvDntwv5Yj9:-KcsoQStKlWYdIJew4V_:b4iwtqc) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L405)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

```JavaScript
server.listen(port, callback);
```
is equivalent to
```JavaScript
server.listen(port);
server.on('listening', callback);
```
In case of error listen will just throw (if there are no listeners for 'error' event). Since server comes as an argument to this function it is not clear if it has 'error' listeners or not.
</blockquote></details>

I agree that the two code snippets are equivalent but one requires a callback and the other requires a bunch of machines to deal with events. So it seems like the callback should be much easier. And since the callback is actually just a hidden wrapper for events it shouldn't matter if the server already has an error listener. They will all be called.

So unless something is wrong with the callback approach I would strongly urge us to adopt it to simplify the code.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 432 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Hr8fzAmBQ5uf3xD1:-KcsozlQAjntKZtAoIKh:b63hv1z) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L432)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

I agree, it makes more sense. I just reused the old code here. But this error string is defined in the spec and we have some tests that depend on it. Maybe we should rethink all errors first? Probably numeric error codes or better custom error types. Relying only on the string value feels too fragile.
</blockquote></details>

Actually the strings were meant to be used in place of error codes because it turns out that looking up and understanding error codes is actually hard. I forgot that this is one of our few defined errors. In the past we deal with this by putting a property onto the Error object with the original string.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772)*

chapko:

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-KcvyttVwCMOsQ_iBEem:b-fvmoo3) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*
> But if we depend on the response being null shouldn't we check that the result is null?

For me it looked more like we had to always return `null` just to make `promiseResultSuccessOrFailure` work.

If I understand it correctly the main purpose of `promiseResultSuccessOrFailure` was to ignore rejections so it can be used with `Promise.all`. This way it will wait for all promises to settle and only then we check result for errors. Other usage I've seen is to just ignore errors, kind of empty `catch` block. So we really never interested in the success result, only whether we have an error or not.

Always returning argument makes it impossible unless all original promises are always resolved with `null`.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 405 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5GXNcOGvDntwv5Yj9:-Kcw66haNGiAo2x3qFIx:b-3haump) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L405)):*

I agree that the two code snippets are equivalent but one requires a callback and the other requires a bunch of machines to deal with events. So it seems like the callback should be much easier. And since the callback is actually just a hidden wrapper for events it shouldn't matter if the server already has an error listener. They will all be called.

So unless something is wrong with the callback approach I would strongly urge us to adopt it to simplify the code.
</blockquote></details>

'listening' callback doesn't report any error so we have to listen to 2 events.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 432 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Hr8fzAmBQ5uf3xD1:-Kcw6biDMz-ay3pW9izG:b-kils0p) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L432)):*

Actually the strings were meant to be used in place of error codes because it turns out that looking up and understanding error codes is actually hard. I forgot that this is one of our few defined errors. In the past we deal with this by putting a property onto the Error object with the original string.
</blockquote></details>

I think it would be easier to put error code property onto the error and use whatever message is appropriate for the context, similar to what [errno](https://github.com/rvagg/node-errno) does.

As for this specific error I don't think it is helpful to put error description into separate property because you have to know this property exists in advance.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772)*

yaronyg:

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-KcxUyF22r0-IvpiARhe:b-oi3u3) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

> But if we depend on the response being null shouldn't we check that the result is null?

For me it looked more like we had to always return `null` just to make `promiseResultSuccessOrFailure` work.

If I understand it correctly the main purpose of `promiseResultSuccessOrFailure` was to ignore rejections so it can be used with `Promise.all`. This way it will wait for all promises to settle and only then we check result for errors. Other usage I've seen is to just ignore errors, kind of empty `catch` block. So we really never interested in the success result, only whether we have an error or not.

Always returning argument makes it impossible unless all original promises are always resolved with `null`.
</blockquote></details>

Right, so if we got a success result from an underlying function (e.g. resolve()) and we didn't get null then this implies a violation of the API contract. Seems like the sort of thing we should at least log?

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 405 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5GXNcOGvDntwv5Yj9:-KcxVxHiUrkK_B0Re-Yn:b-p4agq9) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L405)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

'listening' callback doesn't report any error so we have to listen to 2 events.
</blockquote></details>

Ahhh!!!!!!!! O.k. so now I finally get it! sorry! I thought for no good reason that the callback would call us if we got an error. But I checked the docs and you are right, it will only give us the callback on the listening event, not the error event. O.k. my bad. Sorry. :(

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 432 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Hr8fzAmBQ5uf3xD1:-KcxWIP_Xx6pgqS8gEYX:bpcmc0v) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L432)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

I think it would be easier to put error code property onto the error and use whatever message is appropriate for the context, similar to what [errno](https://github.com/rvagg/node-errno) does.

As for this specific error I don't think it is helpful to put error description into separate property because you have to know this property exists in advance.
</blockquote></details>

This is a generic problem in node and by extension Javascript. The error object is pretty sad. However typically the way one is supposed to deal with this is by shoving in extra properties and then serializing the entire error object onto the log so that any extra properties automatically show up. (Yeah JSON.serialize)

For extra bonus points we could always inherit from Error and create our own Error object but there is a long tradition of just sticking in properties and printing them out which I think is fine for us. Remember, we aren't expecting anyone to make any programmatic decisions based on this property. We just need to see it in the log while we are debugging.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772)*

ThaliTester:
PR is added to the queue for testing as 2. task. (2a8f404)

ThaliTester:
Test 104148237 (2a8f404) build started.

ThaliTester:
Test (Success) 104148237 build is completed (2a8f404)

See https://github.com/ThaliTester/TestResults/tree/1041482372a8f404_Clean_up_ThaliWifiInfrastructure_chapko/ for the logs

ThaliTester:
Test 1041482372a8f404(2a8f404) has failed

See https://github.com/ThaliTester/TestResults/tree/1041482372a8f404_Clean_up_ThaliWifiInfrastructure_chapko/ for the fail logs

chapko:
Review status: 14 of 15 files reviewed at latest revision, 4 unresolved discussions.

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-Kd0VAVJVxqn1zWru0vk:b-guadb4) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*

Right, so if we got a success result from an underlying function (e.g. resolve()) and we didn't get null then this implies a violation of the API contract. Seems like the sort of thing we should at least log?
</blockquote></details>

In my understanding the reason this contract exists in the first place is because of the `promiseResultSuccessOrFailure` function. But again, according to its usage `promiseResultSuccessOrFailure` had never been interested in the success result. It just enforces promises to be always resolved with `null` only because we want to check later if there were any errors. And it just feels like a hack.

Probably the problem is that I don't see other reasons why original promises must be resolved with `null` on success.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 432 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5Hr8fzAmBQ5uf3xD1:-Kd0rETfaBLQ2C9d2rYW:b-896fix) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliWifiInfrastructure.js#L432)):*

This is a generic problem in node and by extension Javascript. The error object is pretty sad. However typically the way one is supposed to deal with this is by shoving in extra properties and then serializing the entire error object onto the log so that any extra properties automatically show up. (Yeah JSON.serialize)

For extra bonus points we could always inherit from Error and create our own Error object but there is a long tradition of just sticking in properties and printing them out which I think is fine for us. Remember, we aren't expecting anyone to make any programmatic decisions based on this property. We just need to see it in the log while we are debugging.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772)*

yaronyg:
:lgtm:

---

Reviewed 1 of 1 files at r5.

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-Kd2Q81bP_dEXsEauqg8:bbkbsw6) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*
<details><summary><i>Previously, chapko (Eugene Chapko) wrote…</i></summary><blockquote>

In my understanding the reason this contract exists in the first place is because of the `promiseResultSuccessOrFailure` function. But again, according to its usage `promiseResultSuccessOrFailure` had never been interested in the success result. It just enforces promises to be always resolved with `null` only because we want to check later if there were any errors. And it just feels like a hack.

Probably the problem is that I don't see other reasons why original promises must be resolved with `null` on success.
</blockquote></details>

Hum... I think we are talking past each other. Let me try a different tack.

muteRejection is design so that if the wrapped promise resolves then muteRejection returns a resolve with null (regardless of the value returned by the underlying promise) and if the promise rejects then muteRejection still calls resolve but this time with the error that was returned in the reject of the underlying promise.

The motivate for this design is that we want to call a number of promises and we don't want them to fail. That is, our logic has rules to continue processing even if they fail. 

So the question we have to ask ourselves is - is there a scenario where we might want the success result from the promise? Because right now we can't get it, it will turn into null.

But seeing how long we've gone on this thread I think it's fine for now to leave the function as it is (e.g. I give in!!!) and if it turns out that the resolve result of the underlying promise is important then we'll have to update muteRejection.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-:-Kd2REFBmuSk2O-894sl:bnfp4nl)*

czyzm:
:lgtm:

---

Reviewed 3 of 14 files at r1, 3 of 4 files at r3, 8 of 8 files at r4, 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-:-KdBRqDDEO7csejmaBx-:bnfp4nl)*

chapko:

---

*[thali/NextGeneration/thaliMobile.js, line 40 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772#-Kc5RfoC3aWO6o_9qhPj:-Kd5Sh1qiMhBbWg2csiw:blbw1z) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/c83cb35bb6d6da387a7e0f332cdc3fd319505718/thali/NextGeneration/thaliMobile.js#L40)):*
I think I finally see your point.

> if it turns out that the resolve result of the underlying promise is important then we'll have to update muteRejection

In this case neither of these abstractions would do their job without additional changes. We will either have false errors or will get `null` instead of returned result.

And sorry for my stubbornness :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1772)*

ThaliTester:
PR is added to the queue for testing as 1. task. (71a06e1)

ThaliTester:
Test 104148237 (71a06e1) build started.

ThaliTester:
Test (Fail) 104148237 build is completed (71a06e1)

See https://github.com/ThaliTester/TestResults/tree/10414823771a06e1_Clean_up_ThaliWifiInfrastructure_chapko/ for the logs

chapko:
Latest commit has only minor fix for test on devices. Build failed because of this:
```
Check dependencies
Code Sign error: No matching provisioning profiles found: No provisioning profiles matching an applicable signing identity were found.

error: command 'thali/install/setUpTests.sh $TEST_TYPE $SERVER_ADDRESS' failed with code 1, file 'build.sh' on line 93
```

ThaliTester:
PR is added to the queue for testing as 2. task. (c182d5d)

ThaliTester:
Test 104148237 (c182d5d) build started.

ThaliTester:
Test (Fail) 104148237 build is completed (c182d5d)

See https://github.com/ThaliTester/TestResults/tree/104148237c182d5d_Clean_up_ThaliWifiInfrastructure_chapko/ for the logs

chapko:
I've tested it locally on devices (ios and android) and all tests are passing, so I'm going to just merge it.

PR #1814, opened 2017-02-28
#1813 Fix inconsistent ThaliPeerPool enqueue API

### LINKED ISSUE #1813: Inconsistent peer pool API
`ThaliPeerPoolInterface` has `enqueue` method which either returns `null` or throws an error. But all the implementations (`ThaliPeerPoolDefault` and `ThaliPeerPoolOneAtATime`) return errors instead. Also `ThaliNotificationClient` does not wrap `enqueue` calls in the `try-catch` block, which can lead to the unhandled errors.

My suggestion is to stick to the `ThaliPeerPoolInterface` design and always throw errors. It can break custom peer pool implementations, but it is easy to update by just replacing `return` with `throw`.

chapko:
1. Makes all ThaliPeerPoolInterface implementations to always throw when invalid action is being enqueued.
2. Updates thaliNotificationClient to use `try-catch` instead of checking returned value

This is a **breaking change**. If you use custom thaliPeerPool implementation you have to throw errors from the `enqueue` method instead of returning them. 

Fixes #1813.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1814)

### DIFF
--- a/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js
+++ b/test/www/jxcore/bv_tests/testThaliManagerCoordinated.js
@@ -73,7 +73,7 @@ function debug() {
     }
     try {
       return JSON.stringify(arg, null, 2);
-    } catch(e) {
+    } catch (e) {
       return String(arg);
     }
   });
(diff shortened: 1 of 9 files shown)


ThaliTester:
PR is added to the queue for testing as 1. task. (e9a30c5)

ThaliTester:
Test 108330275 (e9a30c5) build started.

ThaliTester:
Test (Fail) 108330275 build is completed (e9a30c5)

See https://github.com/ThaliTester/TestResults/tree/108330275e9a30c5__1813_Fix_inconsistent_ThaliPeerPool_enqueue_API_chapko/ for the logs

PR #1826, opened 2017-03-13
Fix wifi network changes handling

### LINKED ISSUE #1823: Thali doesn't work after switching off WiFi
The scenario is the same as in #1822, but instead of crash, we do just do not see any replications at all. 
What is interesting is that we can see in logs that there is indeed some data exchange by bluetooth, but in the end, we don't see changes on the other device.

Here are some logs:
[S7_1](https://gist.github.com/mlesnic/c3269795792b2b4890cb8bcef1a953b4)
[S7_2](https://gist.github.com/mlesnic/dfdaddf0282bef5436333e621d511780)

chapko:
Previously `bssidName:<MAC address>` → `bssidName:null` was considered as simple BSSID change and wifiInfrastructure was restarting ssdp client and server. SSDP start always fails when bssidName is null because it means that we are not connected to any access point and couldn't advertise anything.

In reality bssid change from/to null should be handled the same way as enabling/disabling WiFi on device.

Fixes #1823.

---
This change is [Reviewable](https://reviewable.io/reviews/thaliproject/thali_cordovaplugin/1826)

### DIFF
--- a/test/www/jxcore/bv_tests/testSSDPServer.js
+++ b/test/www/jxcore/bv_tests/testSSDPServer.js
@@ -115,6 +115,38 @@ function changeBssid (value) {
   });
 }
 
+function testBssidChangeReaction (context, newBssid, callCount) {
+  var t = context.t;
+  var stubs = context.stubs;
+  // reset call counts
+  stubs.serverStart.reset();
+  stubs.serverStop.reset();
+  stubs.clientStart.reset();
+  stubs.clientStop.reset();
+  return new Promise(function (resolve) {
+    changeBssid(newBssid);
+    // TODO: #1805
+    setTimeout(function () {
+      t.equal(stubs.serverStart.callCount, callCount.start,
+        'server start called ' + callCount.start + ' times');
+      t.equal(stubs.serverStop.callCount, callCount.stop,
+        'server stop called ' + callCount.stop + ' times');
+      t.equal(stubs.clientStart.callCount, callCount.start,
+        'client start called ' + callCount.start + ' times');
+      t.equal(stubs.clientStop.callCount, callCount.stop,
+        'client stop called ' + callCount.stop + ' times');
+      if (callCount.start === 1 && callCount.stop === 1) {
+        t.ok(stubs.serverStop.calledBefore(stubs.serverStart),
+          'server stop called before start');
+        t.ok(stubs.clientStop.calledBefore(stubs.clientStart),
+          'client stop called before start');
+      }
+      resolve();
+    }, 200);
+  });
+}
+
+
 test(
   'ssdp server and client should be restarted when bssid changed',
   function () {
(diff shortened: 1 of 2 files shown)


yaronyg:
The current code seems a bit complicated. Can we just focus on the BSSID and ignore wifi being on or off?

---

Reviewed 2 of 2 files at r1.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 696 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1826#-Kf7ZSTBU2lWluTT-DQw:-Kf7ZSTBU2lWluTT-DQx:bcvagkx) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/79243628f9e032a040e434b67f89cbe143ce44cc/thali/NextGeneration/thaliWifiInfrastructure.js#L696)):*
> ```JavaScript
>   var lastStatus = this._lastNetworkStatus;
> 
>   /** true if device became connected to the WiFi access point */
> ```

We only care about wifi being on in the sense that if someone tells us to start advertising or listening and wifi is off then we have a problem. But in handleNetworkChanges I don't think we ever really care if wifi is on or off. I think the only thing we care about is: Do we have a BSSID and is it the same BSSID we had before? Can't we just focus on the BSSID and simplify all the code below?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1826#-:-Kf7_zn7a5sZylO2CVeO:bhnxywy)*

chapko:
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[thali/NextGeneration/thaliWifiInfrastructure.js, line 696 at r1](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1826#-Kf7ZSTBU2lWluTT-DQw:-Kf7k_aDAro95-COS2Z1:b-sgr1vq) ([raw file](https://github.com/thaliproject/thali_cordovaplugin/blob/79243628f9e032a040e434b67f89cbe143ce44cc/thali/NextGeneration/thaliWifiInfrastructure.js#L696)):*

We only care about wifi being on in the sense that if someone tells us to start advertising or listening and wifi is off then we have a problem. But in handleNetworkChanges I don't think we ever really care if wifi is on or off. I think the only thing we care about is: Do we have a BSSID and is it the same BSSID we had before? Can't we just focus on the BSSID and simplify all the code below?
</blockquote></details>

Yes, agree. Fixed it. Also I figured out that we actually don't need to do anything on the first networkChanged event, because it happens when thaliWifiInfrastructure is stopped or when it is starting.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1826)*

chapko:
Latest commit fixes one more scenario with wrong hostname advertising (issue #1827)

yaronyg:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1826#-:-KfDFUNgX_ensezYGG42:bnfp4nl)*

mlesnic:
Tested with TeamONE, works really fine, everytime we switch router off and on, phones are syncing immediately. 
:lgtm: 

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/thaliproject/thali_cordovaplugin/1826#-:-KfHEBsu8CDUhnMBckEN:bnw68pg)*

